PR #4, opened 2020-09-29
Add more person attributes and family endpoint

cdhorn:
David,
I think it is great you have started this! Not sure what you had in mind for overall design for the API, but took a pass at adding some more attributes to person and added a family endpoint. Did not update the swagger to reflect them yet, I can as well if this is in line with your thinking. Drop me a line at cdhorn a t embarqmail d 0 t com when you have a chance.
Thanks,
Chris

### DIFF
--- a/gramps_webapi/api/__init__.py
+++ b/gramps_webapi/api/__init__.py
@@ -5,6 +5,7 @@ from flask_restful import Api
 
 from ..const import API_PREFIX
 from .resources.person import PeopleResource, PersonResource
+from .resources.family import FamiliesResource, FamilyResource
 from .resources.token import TokenRefreshResource, TokenResource
 
 api_blueprint = Blueprint("api", __name__, url_prefix=API_PREFIX)
@@ -13,5 +14,7 @@ api = Api(api_blueprint)
 
 api.add_resource(PersonResource, "/person/<string:gramps_id>")
 api.add_resource(PeopleResource, "/person/")
+api.add_resource(FamilyResource, "/family/<string:gramps_id>")
+api.add_resource(FamiliesResource, "/family/")
 api.add_resource(TokenResource, "/login/")
 api.add_resource(TokenRefreshResource, "/refresh/")
--- a/gramps_webapi/api/resources/family.py
+++ b/gramps_webapi/api/resources/family.py
@@ -1,0 +1,52 @@
+"""Family API resource."""
+
+from .base import (
+    GrampsObjectProtectedResource,
+    GrampsObjectResourceHelper,
+    GrampsObjectsProtectedResource,
+)
+from .util import (
+    get_father_grampsid,
+    get_mother_grampsid,
+    get_children_grampsids,
+    get_citation_grampsids,
+    get_attribute_references,
+    get_event_references,
+    get_media_references,
+    get_note_grampsids,
+    get_tag_list,
+)
+
+
+class FamilyResourceHelper(GrampsObjectResourceHelper):
+    """Family resource helper."""
+
+    gramps_class_name = "Family"
+
+    def object_to_dict(self, obj):  # pylint: disable=no-self-use
+        """Return the family as a dictionary."""
+        db = self.db
+        return {
+            "gramps_id": obj.gramps_id,
+            "father": get_father_grampsid(db, obj),
+            "mother": get_mother_grampsid(db, obj),
+            "relationship": str(obj.get_relationship()),
+            "children": get_children_grampsids(db, obj),
+            "attributes": get_attribute_references(db, obj),
+            "events": get_event_references(db, obj),
+            "citations": get_citation_grampsids(db, obj),
+            "notes": get_note_grampsids(db, obj),
+            "media": get_media_references(db, obj),
+#            "media": [{"ref": r.ref, "rect": r.rect} for r in obj.get_media_list()],
+            "tags": get_tag_list(db, obj),
+            "private": obj.private,
+            "change": obj.change
+        }
+
+
+class FamilyResource(GrampsObjectProtectedResource, FamilyResourceHelper):
+    """Family resource."""
+
+
+class FamiliesResource(GrampsObjectsProtectedResource, FamilyResourceHelper):
+    """Families resource."""
--- a/gramps_webapi/api/resources/person.py
+++ b/gramps_webapi/api/resources/person.py
@@ -13,10 +13,16 @@ from .util import (
     get_citation_grampsids,
     get_deathdate,
     get_deathplace_grampsid,
-    get_event_grampsids_roles,
-    get_families_grampsids,
-    get_note_grampsids,
+    get_main_parents_grampsids,
     get_parents_grampsids,
+    get_families_grampsids,
+    get_event_references,
+    get_attribute_references,
+    get_media_references,
+    get_note_grampsids,
+    get_tag_list,
+    get_alternate_names,
+    get_person_associations
 )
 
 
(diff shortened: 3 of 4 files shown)


Nick-Hall:
@cdhorn Thanks for your contribution. It's really nice to see that this project is attracting some interest.

Some of this code seems to be re-inventing the wheel.  The core Gramps code already has [JSON Schema](http://json-schema.org/) definitions for all objects.  They can be accessed using the `get_schema` class method.  e.g. `Person.get_schema()`.

There are also `to_json` and `from_json` methods available in the `gramps.gen.lib.serialize` module.

```
from gramps.gen.lib.serialize import to_json
person = db.get_person_from_gramps_id(gramps_id)
json = to_json(person)
```

The plan is to replace the current raw object format with the json representation (maybe for v5.3?).

We could also use [JSON patch](http://jsonpatch.com/) to modify existing objects.  For example:

`[{ "op": "replace", "path": "/gender", "value": "1" }]`

This could be used with a PATCH request to change the gender of a person.

DavidMStraub:
> It's really nice to see that this project is attracting some interest.

Indeed! I will take this opportunity to move from a "push-to-master" workflow to a pull request workflow and start opening issues for the to-dos so we can divide the work.

> Some of this code seems to be re-inventing the wheel. The core Gramps code already has JSON Schema definitions for all objects. 

Maybe the best way to think about it is to think about an API schema that makes sense from a user perspective and then think about how to implement it (possibly even starting with an inelegant implementation and then streamline it as we "newcomers" improve our understanding of the Gramps codebase). I've added an OpenAPI specification [here](https://github.com/gramps-project/web-api/blob/master/gramps_webapi/data/apispec.yaml) which can be visualized in [Swagger](https://generator.swagger.io/?url=https://raw.githubusercontent.com/gramps-project/web-api/master/gramps_webapi/data/apispec.yaml).

However, in my opinion, while the API endpoints have a 1-to-1 correspondence to Gramps objects (person, family, event, ...), from a user perspective it makes sense for them to *not* be limited to the information present in the serialized raw object.

Some arguments, for the example of the person endpoint:
- the DB handle should be used as internal identifier, but not as user-facing identifier, so I would refer all related objects in the API by their GrampsIDs rather than their handles
- from a user (or app) perspective, it makes sense to directly list the event GrampsIds rather than the `EventRef` objects
- Important properties of a person, such as birthdate &place, make sense to directly return in the API, rather than having to make repeated API calls `Person` -> loop over `EventRef` -> look up `EventType` ...

What one could of course do would be to add the raw serialized object as an additional output key or as an alternative format, e.g. `https://myapi/api/person/I0001?format=raw`. Another "to do" is anyway to introduce query arguments to select a subset of specific keys to return or a specific format (so in case an app only needs a subset of the endpoint information, there is no waste of bandwidth).

DavidMStraub:
Back to reviewing the PR (I suggest continuing the general API discusion in an issue, if necessary). @cdhorn this looks great, thanks a lot! I was also working on the family endpoints, but you are ahead of me, so let's use yours. 

A couple of general comments:
- I suggest to limit PRs to a single feature (e.g. more person attributes vs. family endpoint)
- We can move attributes that are common to all Gramps objects into the base class (like tags and private); there are some of them which are common to "almost" all Gramps objects (e.g. they can all have notes except a note) and we can solve that with a derived base class
- Please always add unit tests. I think we need a simple database with example data to simplify this, I will open an issue for it. (The `example.gramps` included in Gramps is overkill for this IMO.)
- Please always add new objects/endpoints to `apispec.yaml` (feel free however to open a PR without it so we can discuss the API details before putting it into the spec)

Concerning the keys you added:
- There is the question about recursion. For media objects, the GrampsId is not enough because there is also the rectangle that is defined in the MediaRef. For events, there is the role that is defined in the EventRef. But I would not start adding dictionaries/objects for referenced objects, such as person -> media -> citation/tag/note/private as this is a rabbit hole. Can you please remove them? Only for tags I think it makes sense (so person -> tag -> name, color, priority, change) as I suspect we don't want to have a `tag` endpoint. (Or maybe we do?)

cdhorn:
@Nick-Hall I was generally aware of that having looked over some of the code the past couple days. I guess in the end maybe the question is how much additional application logic should reside behind an API versus in the clients.

@DavidMStraub I agree with your general comments in regards to PRs and scope, they make sense. Unit tests will be new for me, but good to do, and I am familiar with the OpenAPI specification.

Regarding the additional keys on the reference objects, I think you have overlooked that they are actually inherent in the data model. For example, I can associate a note or citation with the role a person played in an event, and that is stored as part of the event reference separate from any notes or citations about the event itself. So it is not recursive and the information does need to be exposed.

You mentioned eventually adding support for query arguments to return a subset of information, but the opposite may also be true as perhaps you actually do want all the details about a person and it would be more efficient to gather it in one pass than having the client make repetitive calls to assemble it all. 

Should I close this pull request and focus on refining one more specific to either the person or family endpoint? I see I broke the build tests, had not thought to look at them before hand.

DavidMStraub:
> Regarding the additional keys on the reference objects, I think you have overlooked that they are actually inherent in the data model. For example, I can associate a note or citation with the role a person played in an event, and that is stored as part of the event reference separate from any notes or citations about the event itself. So it is not recursive and the information does need to be exposed.

You are absolutely right, apologies! Makes perfect sense.

> You mentioned eventually adding support for query arguments to return a subset of information, but the opposite may also be true as perhaps you actually do want all the details about a person and it would be more efficient to gather it in one pass than having the client make repetitive calls to assemble it all.

Yes, that's true. Depending on the application, both cases might be useful. Ideally, we will eventually have options to allow to specify this in the request (maybe even something like `recursion_level=n`).

> Should I close this pull request and focus on refining one more specific to either the person or family endpoint? I see I broke the build tests, had not thought to look at them before hand.

I leave it up to you if you want to continue with this branch or do it separately. Whatever is more convenient.

Nick-Hall:
> the DB handle should be used as internal identifier, but not as user-facing identifier, so I would refer all related objects in the API by their GrampsIDs rather than their handles

The `handle` is the primary key in the object tables.  It makes sense to use handles in the API, but not display them to the end user.

The `gramps_id` is not guaranteed to be unique.  It is current policy for us to enforce them to be unique in the UI, but this was not the case in the past.  I expect that many people still have duplicate values in their data.

> from a user (or app) perspective, it makes sense to directly list the event GrampsIds rather than the EventRef objects

So how is the user going to be able to see the event reference attributes?

> Important properties of a person, such as birthdate &place, make sense to directly return in the API, rather than having to make repeated API calls Person -> loop over EventRef -> look up EventType ...

Yes, this is more like a view than an object.  It reminds me of our [Simple Access](https://github.com/gramps-project/gramps/blob/master/gramps/gen/simple/_simpleaccess.py) API.  Returning formatted values for some objects is a good idea.

My suggestion here would be mirror the Gramps json as close as possible to avoid making it too simple.

DavidMStraub:
> The gramps_id is not guaranteed to be unique. It is current policy for us to enforce them to be unique in the UI, but this was not the case in the past. I expect that many people still have duplicate values in their data.

OK, I wasn't aware of that. That's a major problem in my view. By "user" I actually also meant somebody querying the API. Having an identifier be a major part of the API design (as in `GET /api/person/{HANDLE}`) that is not even visible/accessible in the desktop app (or am I wrong) is troubling.

Maybe we can be more picky for this application and simply run a uniqueness check on startup and in the case of duplicates fail with "please fix duplicate Gramps IDs"?

> So how is the user going to be able to see the event reference attributes?

Yes you're right, I had missed that point, @cdhorn actually has solved that in a nice way.

> Yes, this is more like a view than an object.

Precisely, that's a good way to put it.

cdhorn:
@DavidMStraub I made the assumption both were always unique as well. Looked at the database code a little, it appears if there are multiple objects with the same gramps id only the first one is returned. Understanding more now it seems to me it will need to be handle, and I am also thinking @Nick-Hall advice is right and some of the keywords should be changed to align with the internal schema as well.  Your thoughts as I defer to you?

As for the convenience of having the gramps id... what about overloading the individual endpoint? So if the get by handle fails you fall back to get by gramps id? Or would that be considered poor design?

DavidMStraub:
OK, I see the point with the uniqueness of the handle, I guess it makes sense. My above concern

> Having an identifier be a major part of the API design that is not even visible/accessible in the desktop app (or am I wrong) is troubling.

can actually be addressed quite easily, I guess, since even when using the handle as the primary identifier, #17 would allow querying something like

```
/api/people/?gramps_id=I1234
/api/families/?gramps_id=F5678
...
```
So let's use the handle as a primary identifier.

Nick-Hall:
The `gramps_id` field will always be indexed so such queries should be fast.

cdhorn:
'k sounds good. Let me close this for now and I'll open a new one when ready.

PR #20, opened 2020-09-30
Remove flask-restful (fixes #19)

### LINKED ISSUE #19: Remove flask-restful
While starting to look into implementing #17, I encountered this:
https://github.com/flask-restful/flask-restful/issues/883
Namely, the main developer of flask-restful recommending to not use the package. So better to remove it now and simply use `MethodView` from `flask` - one dependency less.

DavidMStraub:
This fixes #19 by replacing `flask_restful` with `MethodView` and `reqparse` with `webargs`, as suggest by the `flask_restful` maintainer.

### DIFF
--- a/gramps_webapi/api/__init__.py
+++ b/gramps_webapi/api/__init__.py
@@ -1,17 +1,26 @@
 """REST API blueprint."""
 
+from typing import Type
+
 from flask import Blueprint
-from flask_restful import Api
 
 from ..const import API_PREFIX
+from .resources.base import Resource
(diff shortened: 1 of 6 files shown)


cdhorn:
Makes sense, I'll wait for you to merge this to add the gramps_id query arg.

DavidMStraub:
BTW @cdhorn concerning the query arg: can you please self-assign issues when you start working on them? Just so we don't do double work. Thanks!

Issue #17, opened 2020-09-29
Filter list-like endpoints with URL arguments


DavidMStraub:
For instance, `/api/people/?tag=TODO` or `/api/events/?participant=I0001`

cdhorn:
Pulled this out of title for the PR as the code for the Gramps filter framework was reverted. 

I'd like to return to this soon as that seemed to be working nicely for a first pass.

cdhorn:
@DavidMStraub @Nick-Hall So I have a pass at this doing what I had done before.  To summarize it at a high level:

The rules query parm for a major object returns a list of the available built in filter rules for that object type. An example rule entry under person would be:

{"description": "Matches people with the particular tag",
"labels": ["Tag:"],
"name": "People with the <tag>",
"rule": "HasTag"}

The logic= keyword allows selection from the supported list of ["and", "or", "xor", "one"]
The invert keyword inverts the results.
The filter= keyword then takes a dictionary as a value, with the keyword being the rule and the value being the list of arguments for it. For example:

filter={"HasTag":["DNAMatch"],"IsMale":[],"PeoplePrivate":[]}

I think this is a decent solution and preferable over introducing a whole slew of query parms for all the different things to filter on and cluttering the namespace. The person class alone has 79 rules at the moment.

Note this was based on examination of the sidebar code. I certainly don't understand the filtering code very well as I did not dive in deeply, it seems like there is more functionality I am overlooking. Is that the case Nick? Can you point me where to look if so to get a better feel for it or does this cover most of it?

Note two other changes kind of naturally fell into place when I did this.

First I did refactor the object_extend as part of this to remove duplicate code from many of the major gramps handler classes. GrampsObjectResourceHelper now handles the default case.

Second I converted all of the boolean query parms like profile, extend, and strip that used to require the =1 to be tacked on into string types so you can just use "profile", "extend", and "strip" by themselves instead. I feel that is cleaner.

Let me know if this all sounds acceptable and I'll update the apispec.yaml for the parm changes/additions and submit a pull request.

Nick-Hall:
@cdhorn It appears that you understand the basics, however there are filter rules that match other filters.  It may be better to create a rule first and then apply it in a separate query.

[...]

PR #24, opened 2020-10-03
Add more person attributes

cdhorn:
@DavidMStraub This is reswizzled to factor out gramps id and use handle and with just changes to the person endpoint.

### DIFF
--- a/gramps_webapi/api/__init__.py
+++ b/gramps_webapi/api/__init__.py
@@ -19,7 +19,7 @@ def register_endpt(resource: Type[Resource], url: str, name: str):
 
 
 # Person
-register_endpt(PersonResource, "/person/<string:gramps_id>", "person")
+register_endpt(PersonResource, "/person/<string:handle>", "person")
 register_endpt(PeopleResource, "/person/", "people")
 # Token
 register_endpt(TokenResource, "/login/", "token")
--- a/gramps_webapi/api/resources/base.py
+++ b/gramps_webapi/api/resources/base.py
@@ -6,6 +6,10 @@ import gramps.gen.lib
 from flask import abort, jsonify
 from gramps.gen.db.base import DbReadBase
 from gramps.gen.db.dbconst import CLASS_TO_KEY_MAP, KEY_TO_NAME_MAP
+from gramps.gen.errors import HandleError
+
+from webargs import fields
+from webargs.flaskparser import use_args
 
 from ..util import get_dbstate
 from . import ProtectedResource, Resource
@@ -52,10 +56,9 @@ class GrampsObjectResourceHelper:
         return self.object_class.create(raw_obj)
 
     def get_object_from_handle(self, handle: str):
-        """Get the object given a Gramps ID."""
+        """Get the object given a Gramps handle."""
         obj_class_key = CLASS_TO_KEY_MAP[self.gramps_class_name]
-        raw_obj = self.db._get_from_handle(obj_class_key, self.object_class, handle)
-        return self.object_class.create(raw_obj)
+        return self.db._get_from_handle(obj_class_key, self.object_class, handle)
 
     def get_gramps_id_from_handle(self, handle: str):
         """Get an object's Gramps ID from its handle."""
@@ -65,10 +68,11 @@ class GrampsObjectResourceHelper:
 class GrampsObjectResource(GrampsObjectResourceHelper, Resource):
     """Resource for a single object."""
 
-    def get(self, gramps_id: str):
+    def get(self, handle: str):
         """Get the object."""
-        obj = self.get_object_from_gramps_id(gramps_id)
-        if obj is None:
+        try:
+            obj = self.get_object_from_handle(handle)
+        except HandleError:
             return abort(404)
         return jsonify(self.object_to_dict_filtered(obj))
 
@@ -76,8 +80,16 @@ class GrampsObjectResource(GrampsObjectResourceHelper, Resource):
 class GrampsObjectsResource(GrampsObjectResourceHelper, Resource):
     """Resource for multiple objects."""
 
-    def get(self):
+    @use_args(
+        {"gramps_id": fields.Str()}, location="query",
+    )
+    def get(self, args):
         """Get all objects."""
+        if "gramps_id" in args:
+            obj = self.get_object_from_gramps_id(args["gramps_id"])
+            if obj is None:
+                return abort(404)
+            return jsonify([self.object_to_dict_filtered(obj)])
         return jsonify(
             [
                 self.object_to_dict_filtered(obj)
--- a/gramps_webapi/api/resources/person.py
+++ b/gramps_webapi/api/resources/person.py
@@ -9,14 +9,16 @@ from .base import (
 )
 from .util import (
     get_birthdate,
-    get_birthplace_grampsid,
-    get_citation_grampsids,
+    get_birthplace_handle,
     get_deathdate,
-    get_deathplace_grampsid,
-    get_event_grampsids_roles,
-    get_families_grampsids,
-    get_note_grampsids,
-    get_parents_grampsids,
+    get_deathplace_handle,
+    get_alternate_names,
+    get_attributes,
+    get_event_references,
+    get_media_references,
+    get_person_references,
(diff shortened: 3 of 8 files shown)


cdhorn:
Silly me, initial commit didn't actually accept handle for the individual lookup because I just kept naturally using gramps id. It works now, and if the handle lookup fails it falls back and tries to treat it like a Gramps id before throwing a 404 which is convenient for the test suite since the handle will always be changing.

DavidMStraub:
Thanks!

Concerning overloading the handle with the Gramps ID, I am a bit unhappy about it as it seems like a hack to me. For an API endpoint like `/api/person/something`, I would expect the `something` to be one kind of object only. Since we decided to use the handle for that, I would really just use that.

If someone wants to look up an object by something else, be it the Gramps ID, the persons name, ..., they can use #17 as mentioned in https://github.com/gramps-project/web-api/pull/4#issuecomment-702915896.

Hope you are OK with this.

cdhorn:
Flask jsonify sorts keys by default, do we want to preserve that behaviour?

DavidMStraub:
Interesting question, I wasn't aware of it. I would tend to say yes, especially that we aim at supporting Python 3.5 (since Gramps 5.2 will support it) and dictionaries don't remember their order until 3.6. So if `jsonify` *doesn't* sort, the order will depend on the Python version.

cdhorn:
Okay I'll shuffle stuff around to account for it.

DavidMStraub:
Why do you actually want to sort the output of `object_to_dict`? That doesn't have any effect, does it?

DavidMStraub:
Do you know why the "files changed" tab shows all the changes from my merged PR #19 as well? Can you please try rebasing on `master`?

cdhorn:
Shouldn't that just be less work for Flask jsonify later?

cdhorn:
Also sorry if I should do things differently, I'm a little new to this... please be patient with me.

DavidMStraub:
> Also sorry if I should do things differently, I'm a little new to this... please be patient with me.

No worries!

> Shouldn't that just be less work for Flask jsonify later?

The sorting is done by Python's `json` standard library, is implemented in C and is very fast (I just checked). So it's going to be faster than everything else done by this function.

Don't get me wrong, it doesn't hurt either, but I would say it's just a case of "premature optimization" :wink:

cdhorn:
@DavidMStraub can you take a look over the simplify branch in my forked repo, I think it is more like what @Nick-Hall had been suggesting and may be a better approach. One major change is having a static format, it may send more over the wire but the client doesn't need to worry about testing if a key is present all the time before referencing something. Not sure if you are in agreement with that though.

DavidMStraub:
I had a brief look. Great job, and I agree that it's much more elegant to leverage everything that is already there like `citation_list`, `note_list`, `tag_list` etc.

Concerning the "static format", I'm not so sure, I would rather keep the response limited to non-empty keys ([some discussion here](https://softwareengineering.stackexchange.com/questions/285010/null-vs-missing-key-in-rest-api-response)), but I don't think we have to decide it right now. It will be trivial to filter out empty ones later if we decide to do so.

The only thing I'm not sure about is how to deal with the "denormalized" keys. As you wrote before, using only the "raw" format will leave the "apps" to do all the work (assembling back the name from its pieces, finding the birth event and its date, formatting it ...) so we *do* want to have this "preprocessed" information. But simply adding additional keys next to the "raw data" also seems strange to me.

But I agree that your new branch is a much cleaner interpretation, so I suggest to continue with a PR of this branch and we can discuss the details there.

BTW, are you using `pre-commit` as described in the readme? The code looks like it's not formatted with `black`.

DavidMStraub:
>  But simply adding additional keys next to the "raw data" also seems strange to me.

I thought a bit more about this and one possibility would be to have
 
- `/api/person/some-handle` return the raw dictionary
- `/api/person/some-handle?format=simple` (or something similar) return something that is closer to what I had in the web app backend, so derived quantities like `{'birthdate': ..., 'birthplace': ..., 'name_given': ...}`

The reason I think this is useful is because the second, shorter format would be useful to display lists of objects (such as [this](https://mysterious-cove-02458.herokuapp.com/people) - username and password = `test`) without having to iterate through nested sub-properties, while the raw format (with optional `recursion`) could be used to show object details.

In any case, I think it's useful if, for the moment, we just implement the raw format based on your new branch and then add more stuff later.

cdhorn:
I have obligations today so will get to this later in the day.

Sorry I think I overlooked the pre-commit hooks, will setup. 

For any denormalized attributes, and I agree they are a good idea, why not group them together under a single "profile" attribute?  Then in addition to a strip option we can have a raw option that strips the profile out and a keywords option to take a comma delimited set of keys to return? Aka profile,handle,gramps_id?  I think that would work well.

DavidMStraub:
> For any denormalized attributes, and I agree they are a good idea, why not group them together under a single "profile" attribute? Then in addition to a strip option we can have a raw option that strips the profile out and a keywords option to take a comma delimited set of keys to return? Aka profile,handle,gramps_id? I think that would work well.

Yes, sound good!

Nick-Hall:
This seems to be heading in the right direction.

Do we want to include fallbacks for the birth and death events?  Perhaps these events should be structures:
```python
[{
  'type': 'Baptism',
  'date': '1900-1-1',
  'place': 'London, England'
},
{
  'type': 'Death',
  'date': '1950-2-2',
  'place': 'London, England'
}]
```
Which leads to the question:  How do we want to represent a `GrampsType`?  Should we provide a translation based on the browser locale?  e.g. `[15, 'Baptism', 'Baptême']` in French.

How should we handle the `group_as` and `sort_as` fields of a name?  Do they need to form part of the name or will all grouping and sorting be done on the server?

DavidMStraub:
> Perhaps these events should be structures:

In the "raw" format that attempts to be close to the serialized JSON, I think it would be, for a person,

```
{"_class": "Person", ...,  "event_ref_list": [
    {"_class": "EventRef", ..., "ref": "Z8BBWT1N2B53UH0LHP", ...}
], ...}
```
while @cdhorn was suggesting to have in addition a `recursion` option, that would replace the handle by the serialized event,

```
{"_class": "Person", ...,  "event_ref_list": [
    {"_class": "EventRef", ..., "ref": {
            {"_class": "Event", ..., "type": {"_class": "EventType", "string": "Birth"}}
}
        }
    ...}
], ...}
```

Your suggestion would be closer to the "simple" format discussed above, and I see motivation for both formats.

Concerning

> Which leads to the question: How do we want to represent a GrampsType? Should we provide a translation based on the browser locale? e.g. [15, 'Baptism', 'Baptême'] in French.

I would simply introduce a keyword argument `lang=de_DE` etc., since this would allow the user to change the language in an app (see e.g. [this feature request](https://github.com/DavidMStraub/gramps-webapp/issues/31)). Using the browser locale would be an app implementation detail (i.e. setting `lang` in API calls according to the locale). The API doesn't even have to be called from a browser...

BTW, how does `to_json` deal with this? I guess it simply uses the current `GRAMPSLOCALE`?

DavidMStraub:
BTW, maybe we should open separate issues for all these different topics :slightly_smiling_face:

Nick-Hall:
> Your suggestion would be closer to the "simple" format discussed above, and I see motivation for both formats.

It was a suggestion for the contents of the "profile" attribute, rather than using `birthdate`, `birthplace`, `deathdate` and `deathplace`.

Nick-Hall:
> BTW, how does to_json deal with this?

It doesn't need to it just uses the English string for pre-defined values and the string provided for custom values.  This is enough information for `from_json` to reconstruct the object.

DavidMStraub:
>  It was a suggestion for the contents of the "profile" attribute, rather than using birthdate, birthplace, deathdate and deathplace.

Oh, now I get it. Yes, good idea.

> It doesn't need to it just uses the English string for pre-defined values and the string provided for custom values. This is enough information for from_json to reconstruct the object.

OK, I was just surprised that the event role ("Primary"/"Primär") and event names ("Birth"/"Geburt") are localized (German) when I call `to_json`. For the "raw" API endpoint, I guess we also have to allow changing this.

PR #25, opened 2020-10-04
Simplify (closes #23, closes #22,  closes #21, closes #18, closes #12, closes #11, closes #10, closes #9, closes #8, closes #7, closes #6)

### LINKED ISSUE #6: Endpoints: event, events

### LINKED ISSUE #7: Endpoints: place, places

### LINKED ISSUE #8: Endpoints: citation, citations

### LINKED ISSUE #9: Endpoints: source, sources

### LINKED ISSUE #10: Endpoints: mediaobject, mediaobjects
Or `media`, but there is no `medias`...

### LINKED ISSUE #11: Endpoints: repository, repositories

### LINKED ISSUE #12: Endpoint: note
I suggest not having a `notes` endpoint as notes can have a lot of content, so it might not make sense to fetch all of them simultaneously.

### LINKED ISSUE #18: Allow limiting keys in response
If an app only requires part of the object returned at an endpoint, it should be possible to reduce bandwidth/latency by specifying the desired keys in advance, e.g. `/api/person/I0001/?keys=gramps_id,birthdate`

### LINKED ISSUE #21: Endpoint: family, families
Add family and families endpoints

### LINKED ISSUE #22: Endpoint: tag
Add a tag endpoint

### LINKED ISSUE #23: Additional attributes for person endpoint
Add additional attributes to person endpoint

cdhorn:
Refactored code using native Gramps schema and to use json encoder for filter handling


cdhorn:
Okay guys, this pass is cleaned up a little bit.  The strip/raw/keys query parms are there but the logic doesn't fully do what is desired for the keys one yet.

Raw is not really "raw" anymore. I am not sure we should provide that. If you really think that should be an option we can add support for it back in the encoder.

The swagger isn't updated to account for all endpoints and objects yet, only person. And I only tweaked the tests to make them pass.  I think it worth playing with more and being sure we have the format down the way we want before spending time getting those updated to match.

@Nick-Hall I think an expanded profile like that is a great idea!

I was thinking the encoder would be a good place to handle "recursion", it feels like it belongs there. But the context is wrong.  Have to play with it I guess.

cdhorn:
Should I be able to link a pull request with 1 or more issues? That option seems disabled for me.

DavidMStraub:
Awesome! Looking forward to review it (don't think I will manage today).

> Should I be able to link a pull request with 1 or more issues? That option seems disabled for me.

The easiest way (perhaps the only one) is to just put the issue numbers in the PR title (like so: #12 #11 ...)

cdhorn:
Hey @DavidMStraub, no rush, I have a lot going on this week too. I am interested in feedback from both you and @Nick-Hall though, especially if either of you see a better way to do or structure things.

cdhorn:
So played a bit more with this tonight. The raw query parm is gone, use skipkeys to filter specific keys out. I left "profile" as it was for now. The new "extend" parm will add an "extended" section that pretty much contains all the handles referenced elsewhere by the object. I think this is cleaner than trying to add some kind of recursion, and accomplishes what I felt was important, being able to get nearly everything about a person in one shot as the family data is extended a level as well to include all children, parents and siblings.

gramps_webapi/api/resources/base.py
```diff
@@ -52,10 +56,9 @@ def get_object_from_gramps_id(self, gramps_id: str):
         return self.object_class.create(raw_obj)
 
     def get_object_from_handle(self, handle: str):
-        """Get the object given a Gramps ID."""
+        """Get the object given a Gramps handle."""
         obj_class_key = CLASS_TO_KEY_MAP[self.gramps_class_name]
-        raw_obj = self.db._get_from_handle(obj_class_key, self.object_class, handle)
-        return self.object_class.create(raw_obj)
+        return self.db._get_from_handle(obj_class_key, self.object_class, handle)
```
Nick-Hall: Methods that start with an underscore are protected methods.  They should only be used in sub-classes.

gramps_webapi/api/resources/base.py
```diff
@@ -65,22 +68,70 @@ def get_gramps_id_from_handle(self, handle: str):
...
+            try:
+                obj = self.get_object_from_handle(handle)
+            except HandleError:
+                return abort(404)
+            return self.response([self.object_extend(obj)])
+        return self.response(
             [
-                self.object_to_dict_filtered(obj)
+                self.object_extend(obj)
                 for obj in self.db._iter_objects(self.object_class)
```
Nick-Hall: Another protected method.

gramps_webapi/api/resources/bookmark.py
```diff
@@ -0,0 +1,60 @@
...
+    @property
+    def db(self) -> DbReadBase:
+        """Get the database instance."""
+        return get_dbstate().db
+
+    def get(self, bookmark_type: str):
+        """Get list of bookmarks by type."""
+        db = self.db
+        if bookmark_type == "bookmarks":
+            return self.response(db.get_bookmarks())
```
Nick-Hall: These are actually person bookmarks.  The method name dates back to when Gramps only had one type of bookmark.

gramps_webapi/api/resources/bookmark.py
```diff
@@ -0,0 +1,60 @@
...
+            return self.response(db.get_event_bookmarks())
+        if bookmark_type == "family":
+            return self.response(db.get_family_bookmarks())
+        if bookmark_type == "media":
+            return self.response(db.get_media_bookmarks())
+        if bookmark_type == "note":
+            return self.response(db.get_note_bookmarks())
+        if bookmark_type == "place":
+            return self.response(db.get_place_bookmarks())
+        if bookmark_type == "repo":
```
Nick-Hall: Use "repository" instead of "repo"?

gramps_webapi/api/resources/type.py
```diff
@@ -0,0 +1,81 @@
...
+        """Get the type."""
+        db = self.db
+        if gramps_type == "event_attribute":
+            return self.response(db.get_event_attribute_types())
+        if gramps_type == "event":
+            return self.response(db.get_event_types())
+        if gramps_type == "person_attribute":
+            return self.response(db.get_person_attribute_types())
+        if gramps_type == "family_attribute":
+            return self.response(db.get_event_types())
```
Nick-Hall: Cut & paste problem?

Nick-Hall (commented):
You have made good progress!

It may need some refinements, but it I don't see any major issues.

gramps_webapi/api/resources/base.py
cdhorn: I'll review that more closely and see what makes sense.

gramps_webapi/api/resources/bookmark.py
cdhorn: Let's call this person then.

gramps_webapi/api/resources/bookmark.py
cdhorn: Yeah, be consistent.

gramps_webapi/api/resources/type.py
cdhorn: Yep.

gramps_webapi/api/resources/base.py
Nick-Hall: Have a look at the [`method`](https://github.com/gramps-project/gramps/blob/master/gramps/gen/db/base.py#L1436) convenience function.

cdhorn:
I'm not so happy with the extend option and rethinking it a little. Realize it doesn't really pull in what one would want. I guess this is the how much logic should exist behind the API versus in a client.

gramps_webapi/api/resources/base.py
cdhorn: Modified to use db.method now

gramps_webapi/api/resources/base.py
cdhorn: Thanks, modified to use db.method to get the method now

cdhorn:
"profile" now not generated by default, use the profile=1 query parm for it.  Only done for person, family, and event at this time. The profile now includes the birth/death profile in the format you suggested Nick with a fallback check, and for a person all the family related information. A marriage profile is in the family profile using same methadology.

cdhorn:
Renamed "parent_families" in "profile" to "other_parent_families" and factored the primary one out so as to not duplicate the data.

DavidMStraub:
Wow, this is such a huge PR that I still haven't grasped it fully, but what I've seen so far looks very good! I'll definitely go through all of it on the weekend.

cdhorn:
Thanks Nick, I looked over the locale stuff closer, let me know if refactor is any better.

cdhorn:
Hey guys!  So I took an initial pass tonight at adding support for the Gramps filtering framework and hope I did okay with it. The new keywords on the primary end points are rules, logic, invert and filter. rules=1 will list the rules available for that object type and logic and invert set the corresponding filter attributes. The filter= itself is a dict containing one or more rules in the form of the class name as key and then the arguments for it as value.  So for example this would do a search for a tag on all males with the default "and" logic: filter={"HasTag":["Tag2"],"IsMale":[]}  I know that is a bit more than just something simple like tag=Tag2 but it makes so much more functionality available.

The argument handling for the filter argument itself is a hack, if one of you is able to fix it up right please do. I tried a few things and I don't know what isn't clicking so set it aside for the time being to try to get the rest working.

gramps_webapi/api/resources/emit.py
```diff
@@ -0,0 +1,94 @@
...
+    "Note": lib.Note,
+    "Tag": lib.Tag,
+}
+
+
+class GrampsJSONEncoder(JSONEncoder):
+    """Customizes Gramps Web API output."""
+
+    def __init__(self):
+        """Initialize class."""
```
DavidMStraub, line 20: This is a strange way to initialize a class; all the arguments are set to defaults and cannot be changed on instantiation; wouldn't it be more elegend to have `__init__(self, sort_keys=True, ...)`?

gramps_webapi/api/resources/emit.py
```diff
@@ -0,0 +1,94 @@
...
+        return Response(
+            response=self.encode(payload),
+            status=200,
+            mimetype="application/json",
+        )
+
+    def api_filter(self, obj):
+        """Filter data for a Gramps object."""
+        data = {}
+        filter = False
```
DavidMStraub: I would avoid using built-ins like `filter` as variable names

gramps_webapi/api/resources/emit.py
```diff
@@ -0,0 +1,94 @@
...
+        try:
+            if isinstance(obj, PRIMARY_CLASS_MAP[self.gramps_class_name]):
+                filter = True
```
DavidMStraub: `self.gramps_class_name` is undefined.

If this is due to it being an abstract base class, I would declare a dummy class attribute (e.g. empty string) and fail in case the child class does not declare it properly.

Why does this `pass` if `self.gramps_class_name` is not in `PRIMARY_CLASS_MAP`? When can this happen?

gramps_webapi/api/resources/emit.py
```diff
@@ -0,0 +1,94 @@
+"""Gramps Json Encoder."""
+
+import inspect
+
+import gramps.gen.lib as lib
+from flask import Response
+from flask.json import JSONEncoder
+from gramps.gen.db import DbBookmarks
+
+PRIMARY_CLASS_MAP = {
```
DavidMStraub: I would move this constant dict to the `const` module.

gramps_webapi/api/resources/emit.py
```diff
@@ -0,0 +1,94 @@
...
+        data = {}
+        filter = False
+        try:
+            if isinstance(obj, PRIMARY_CLASS_MAP[self.gramps_class_name]):
+                filter = True
+        except:
+            pass
+        for key, value in obj.__dict__.items():
+            if filter:
+                if self.filter_only_keys != [] and key not in self.filter_only_keys:
```
DavidMStraub: ` if self.filter_only_keys and ...`

gramps_webapi/api/resources/emit.py
```diff
@@ -0,0 +1,94 @@
...
+class GrampsJSONEncoder(JSONEncoder):
+    """Customizes Gramps Web API output."""
+
+    def __init__(self):
+        """Initialize class."""
+        JSONEncoder.__init__(self)
+        self.sort_keys = True
+        self.ensure_ascii = False
+        self.strip_empty_keys = False
+        self.filter_only_keys = []
```
DavidMStraub, line 23: I would cast the two filter lists to sets, as this will speed up looking for keys below.

gramps_webapi/api/resources/emit.py
```diff
@@ -0,0 +1,94 @@
...
+        except:
+            pass
+        for key, value in obj.__dict__.items():
+            if filter:
+                if self.filter_only_keys != [] and key not in self.filter_only_keys:
+                    continue
+                if self.filter_skip_keys != [] and key in self.filter_skip_keys:
+                    continue
+            if key.startswith("_"):
+                key = key[2 + key.find("__") :]
```
DavidMStraub, line 76: This took me quite long to understand. What is the rational here? You want to strip initial "_" and "__" I guess? Why not just `key.lstrip("_")`? Do you expect cases where the `__` is not in the beginning? And why do we need `_`-prefixed properties at all?

gramps_webapi/api/resources/emit.py
```diff
@@ -0,0 +1,94 @@
...
+                if isinstance(value, lib.GrampsType):
+                    data[key] = str(value)
+                elif isinstance(value, lib.StyledText):
+                    data[key] = str(value)
+                elif isinstance(value, lib.PlaceName):
+                    data[key] = {
+                        "date": value.date,
+                        "lang": value.lang,
+                        "value": value.value,
+                    }
```
DavidMStraub: I would put all string constants (even date, lang, value) in `const` to ensure consistency.

gramps_webapi/api/resources/emit.py
```diff
@@ -0,0 +1,94 @@
...
+                else:
+                    if value is not None and value != [] and value != {}:
+                        data[key] = value
+            else:
+                data[key] = value
+        return data
```
DavidMStraub, line 79: I don't understand the use of `strip_empty_keys` here. If it is true, values are casted to stings and *then* empty values are skipped. If it is false, the values are left as they are. I would expect `strip_empty_keys` to *only* determine whether empty values are omitted or not. Or what am I missing?

gramps_webapi/api/resources/emit.py
DavidMStraub: Just realized that there is an abstract property on `GrampsObjectResourceHelper`. So this should be moved to `GrampsJSONEncoder`.

gramps_webapi/api/resources/base.py
```diff
@@ -3,86 +3,138 @@
...
 from . import ProtectedResource, Resource
+from .emit import GrampsJSONEncoder
 
 
-class GrampsObjectResourceHelper:
+class GrampsObjectResourceHelper(GrampsJSONEncoder):
     """Gramps object helper class."""
 
+    def __init__(self):
+        """Initialize class."""
```
DavidMStraub: Again, why are the properties (`build_profile`) not keyword arguments to the constructor?

Also, it would be great if you could explain those properties in the doc string, it's really hard otherwise to understand what's going on for who hasn't written the code themselves.

gramps_webapi/api/resources/base.py
```diff
@@ -3,86 +3,138 @@
...
 
-    @property
-    def object_class(self):
-        """Get the Gramps class of the object."""
-        obj_class_name = KEY_TO_NAME_MAP[CLASS_TO_KEY_MAP[self.gramps_class_name]]
-        obj_module = getattr(gramps.gen.lib, obj_class_name)
-        obj_class = getattr(obj_module, self.gramps_class_name)
-        return obj_class
-
     def get_object_from_gramps_id(self, gramps_id: str):
```
DavidMStraub: Not sure if this is code I wrote myself :wink: but for the type hints here and in the following methods, we should use `Handle` and `GrampsId`. In fact, when using only `str`, `mypy` has no way of knowing whether one accidentally plugs a handle into a function that expects a Gramps ID, and vice versa.

gramps_webapi/api/resources/base.py
DavidMStraub: OK, now I understand it ... you mutate these attributes in the method calls. But I find this problematic; why do we change these instance attributes just for the sake of a single method call? And then change it back in the next call? Since these attributes do nothin else but change the behaviour of `reponse`, I think they need to be keywod arguments to `response`, and not instance attributes. (The instance doesn't know anything about which attributes are user might request in a given HTTP call.)

gramps_webapi/api/resources/base.py
```diff
@@ -3,86 +3,138 @@
...
+        if "rules" in args:
+            return self.response(self.object_filter_rules())
+        if "gramps_id" in args:
+            obj = self.get_object_from_gramps_id(args["gramps_id"])
+            if obj is None:
+                return abort(404)
+            return self.response([self.object_extend(obj)])
+        if "handle" in args:
+            try:
+                obj = self.get_object_from_handle(handle)
```
DavidMStraub: `handle` is undefined.

DavidMStraub (commented):
Hi @cdhorn, again it looks really great! I started reviewing the code, apologies if I'm being pedantic, but being pedantic at the beginning will pay off later I believe :wink: 

Will add more comments later.

gramps_webapi/api/resources/family.py
```diff
@@ -0,0 +1,53 @@
...
+                   get_family_profile_for_object, get_media_for_references,
+                   get_person_by_handle)
+
+
+class FamilyResourceHelper(GrampsObjectResourceHelper):
+    """Family resource helper."""
+
+    gramps_class_name = "Family"
+
+    def object_extend(self, obj):  # pylint: disable=no-self-use
```
DavidMStraub: There is self use in this method (so the comment should go away). This was probably done erroneously by myself.

gramps_webapi/api/resources/person.py
```diff
@@ -1,49 +1,58 @@
...
-    get_deathplace_grampsid,
-    get_event_grampsids_roles,
-    get_families_grampsids,
-    get_note_grampsids,
-    get_parents_grampsids,
-)
+import gramps.gen.filters.rules.person as rule_classes
+
+from .base import (GrampsObjectProtectedResource, GrampsObjectResourceHelper,
+                   GrampsObjectsProtectedResource)
```
DavidMStraub: This is not formatted with `black`; are you using the pre-commit hook?

DavidMStraub: In case you are using `isort`, it is always best to first sort and then format, since `isort` has its own ideas about formatting that are different from `black`.

gramps_webapi/api/resources/translation.py
```diff
@@ -0,0 +1,38 @@
...
+        location="query",
+    )
+    def get(self, args):
+        """Get translation."""
+        if "strings" not in args:
+            return self.response(GRAMPS_LOCALE.get_language_dict())
+
+        try:
+            strings = json.loads(args["strings"])
+        except:
```
DavidMStraub: Please avoid bare `except`s.

gramps_webapi/api/resources/translation.py
```diff
@@ -0,0 +1,38 @@
...
+            abort(400)
+        if "lang" in args and args["lang"] is not None:
```
DavidMStraub: `if args.get("lang"):`

gramps_webapi/api/resources/util.py
```diff
@@ -1,112 +1,246 @@
...
+        "media": get_media_for_references(db, obj),
+        "notes": [db.get_note_from_handle(handle) for handle in obj.note_list],
+        "tags": [db.get_tag_from_handle(handle) for handle in obj.tag_list],
+    }
+
+
+def get_sex_profile(person: Person):
+    """Get character substitution for enumerated sex."""
+    if person.gender == person.MALE:
+        return "M"
```
DavidMStraub: These should be string constants in `const`

DavidMStraub:
This is great, but it would really help to have more atomic PRs in the future. 29 changed files with so much new stuff  - its really very hard to see what is going on. A single endpoint would have been enough to discuss the basic API structure.

Anyway, I shouldn't complain since it's great that we will have *everything* once this is merged, but it takes time to understand and review.

gramps_webapi/api/resources/util.py
```diff
@@ -1,112 +1,246 @@
...
+        "place": pd.display_event(db, event),
+    }
+
+
+def get_event_profile_for_handle(db: DbReadBase, handle: Handle) -> Dict:
+    """Get event profile given a handle."""
+    try:
+        obj = db.get_event_from_handle(handle)
+    except HandleError:
+        return None
```
DavidMStraub, line 90: That's a nice example of the usefulness of type hints and `mypy`.

By declaring `Dict`, you are telling the static type checker that this function always returns a dict (unless it raises). But then you return `None`, which is inconsistent.

So the type hint should be changed to `Optional[None]`.

This might seem unnecessary and annoying when you start doing it, but believe, it becomes very very useful down the line when you refactor stuff ... always knowing whether a function returns a specific type or can return `None` and having a static type checker keeping an eye on it.

I should actually add `mypy` to the CI pipeline.

DavidMStraub:
So I was able to generate a few outputs using the Gramps example database and I really like the API. Also the `profile` and `extend` solution is really good and, combined with the `keys` arguement, allows quite flexible use of the API.

But now the new filter feature (which is a great idea) again adds a ton of code and you mention yourself that part of it is still a hack.

Can't we please just get the original PR with all the new structure etc. over the finish line and then add more features one after another? If you revert just this single commit, that would allow as to focus on getting everything merged and then continue from there.

I also suggest not adding too much features without having unit tests for the existing ones in place.

Again, great work! :tada:

gramps_webapi/api/resources/emit.py
cdhorn: Agree, fixed and on GrampsObjectResourceHelper. The attributes there apply only to primary resources so those 2 I thought should be set there.

gramps_webapi/api/resources/emit.py
cdhorn: Yep, change to apply_filter.

gramps_webapi/api/resources/emit.py
cdhorn: Just realized setting gramps_class_name property of None on GrampsObjectResourceHelper solves what I needed to figure out, which is whether it was being called from a primary Gramps object as the key filters are only intended to apply to those.

gramps_webapi/api/resources/emit.py
cdhorn: It is no longer needed.

gramps_webapi/api/resources/emit.py
cdhorn: Duh, fixed.

gramps_webapi/api/resources/emit.py
cdhorn: I can do that I guess.

cdhorn: I'm not really sure what that accomplishes. The Gramps schema isn't likely to change.

gramps_webapi/api/resources/emit.py
cdhorn: If the value itself is a Gramps object the comparison to an empty list or dict will fail and throw an exception as the later have no value. This is sort of an attempted optimization but without careful review of the Gramps schema so some cases may be missed. It could be:

            if self.strip_empty_keys:
                for gramps_class in self.gramps_classes:
                    if isinstance(value, gramps_class):
                        data[key] = value
                        break
		if key not in data:
                    if value is not None and value != [] and value != {}:
                        data[key] = value
            else:
               data[key] = value

Thoughts? Is there a better way to handle this?

gramps_webapi/api/resources/base.py
cdhorn: Maybe this should be refactored, I will look at it.

gramps_webapi/api/resources/base.py
cdhorn: Yeah I added them in some places but didn't review carefully, will look at.

cdhorn:
Reverted filter, focusing on your comments above. Yeah I know I kind of ran with it. Once this is in I can look at filter again or if you agree with format focus on getting apispec up to snuff then unit tests. I think those should be separate pr.

gramps_webapi/api/resources/base.py
cdhorn: Oops

gramps_webapi/api/resources/family.py
cdhorn: Cleaned up

gramps_webapi/api/resources/translation.py
cdhorn: Fixed

gramps_webapi/api/resources/person.py
cdhorn: Still not setup right, need to look closer at it.

gramps_webapi/api/resources/translation.py
cdhorn: Fixed.

gramps_webapi/api/resources/util.py
cdhorn: Agree

gramps_webapi/api/resources/person.py
cdhorn: (gramps_webapi) [webapi@abyss web-api]$ more .pre-commit-config.yaml
repos:
  - repo: https://github.com/psf/black
    rev: stable
    hooks:
      - id: black
  - repo: https://github.com/pre-commit/mirrors-isort
    rev: v4.3.21
    hooks:
      - id: isort
  - repo: https://github.com/PyCQA/pylint
    rev: pylint-2.5.3
    hooks:
      - id: pylint
        stages: [commit]
        args:
          - --score=n
  - repo: https://github.com/PyCQA/pydocstyle
    rev: 5.0.2
    hooks:
      - id: pydocstyle
        stages: [commit]

I will reverse order.

gramps_webapi/api/resources/base.py
cdhorn: Yeah, much cleaner now.

cdhorn:
Probably needs more cleanup if you can take another pass through David.  And **thank you** and Nick for being pedantic, I wouldn't be learning anything otherwise!

gramps_webapi/api/resources/emit.py
DavidMStraub: Yeah maybe I'm being a bit extreme here. But for instance there might be several places in the code where there is a key indicating a locale string or language, and to avoid having "lang" in one place, "locale" in another, and maybe "language" somewhere else, using a constant for everything makes it easier to remain consistent. It also more easily catches typos (since a typo in a variable raises an error).

gramps_webapi/api/resources/emit.py
DavidMStraub: I still don't understand why the Gramps class instances are converted to strings (`data[key] = str(value)`) but *only* when `strip_empty_keys` is true. Why is that?

What I would find logical would be simply
```python
if not self.strip_empty_keys or not is_null(value):
    data[key] = value
```

and then
```python
def is_null(object) -> bool:
    ...
```
should somehow handle how to determine whether Gramps objects are null

gramps_webapi/api/resources/person.py
DavidMStraub: Oh indeed, looks like I screwed that up, sorry :hand_over_mouth:

gramps_webapi/api/resources/emit.py
cdhorn: Yeah this wasn't needed anymore in latest pass. Did make the character substitution for sex aka gender constants as suggested.

gramps_webapi/api/resources/person.py
cdhorn: Actually, after running pre-commit install to install the hook, which we need to update README for maybe, I can't commit because they all seem to conflict with each other.  So I disabled that and ran manually again like I was. Maybe I'm missing something.

gramps_webapi/api/resources/emit.py
cdhorn: Oh! Actually isn't this more proper?

```
if not self.strip_empty_keys or value:
    data[key] = value
```

gramps_webapi/api/resources/person.py
cdhorn: isort fails because it reformatted what black did, and black fails because it reformatted what isort did. 

pylint is unhappy about lots of stuff.  Do we want to fix as much of that as possible?  I can look at that more later tonight after church.

cdhorn: Although some of what pylint seems to not like is what black and isort did. I imagine there must be a way to skip some checks to make them all play nice.

Nick-Hall: We can modify the `.pylintrc` file.  What doesn't it like?

cdhorn: Numerous wrong hanging indentations before blocks are the result of black reformatting.
db as both an attribute and variable doesn't conform to snake_case.
Some classes have too few public methods since they currently have none.
Too many return statements.
Too many branches.
Some duplicate code across files.
Use of [] and {} is dangerous as default arguments.
Method object_extend is abstract in GrampsObjectResourceHelper but not overridden, a few unused imports.

cdhorn: Added a .pylintrc to disable C0330 and C0326 as per recommendations here: https://black.readthedocs.io/en/stable/compatible_configs.html

Also updated version of isort being used and set it to use the black compatible profile. 

Added mypy to the pre-commit config and cleaned some stuff up it was not happy about. 

Added a .flake8 but did not add it to the pre-commit config. I think maybe we should.  I did run it manually though and fixed an issue it found that pylint did not. It is also complaining about a naked except: in gramps_webapi/auth.py but left that alone for now.

Was able to get the pylint exit code down from 30 to 10, but not there yet.

cdhorn: Oh my oh my. I was curious and just started running pylint against the Gramps codebase...

gramps_webapi/api/resources/emit.py
DavidMStraub: I think this won't work as e.g. `0` evaluates as `False`. So you will discard all keys where the value is 0.

DavidMStraub: In addition, I wouldn't know what `bool(some_gramps_object)` even does...

gramps_webapi/api/resources/person.py
DavidMStraub: Yeah that's why I think enforcing formatting and being super strict is helpful when starting a project from scratch :slightly_smiling_face:

gramps_webapi/api/resources/emit.py
cdhorn: Oh that's right, I forgot 0.  Need to tack that last test on the end. 

As I currently understand it bool of an object returns True if it exists unless the class implements a __bool__() or __len()__ method that happens to return otherwise. bool of [] or {} or None always returns False. So that seems to be a reasonable test to me.

cdhorn: Ah but False might be the value of the key. Crumb buns.

cdhorn: So this should do, am I still overlooking a case...
`            if not self.strip_empty_keys or value or int(value) == 0 or isinstance(value, bool):
		data[key] = value
`

gramps_webapi/api/resources/person.py
cdhorn: I am really liking this pre-commit setup, especially black.

cdhorn: Some of the pylint complaints I'm a little unsure how to go about resolving.  Do you want them all resolved before a merge?  And what else would you like to see cleaned up before one?

cdhorn: Also, going to start on apispec.yaml updates today.

cdhorn:
I need to sleep more, no idea what I was thinking changing that code although replacement is cleaner.

A couple questions... 

Should the ?handle= and ?gramps_id= query parms on /person return the single object or the single object in a list? I was returning as a object in a list to be consistent with the endpoint itself but the parm is sort of intended to match /{object}/{handle}

Also since /translate returns possible list of translations would it be more consistent to use /translate/{lang} with ?strings=[] to get a translation?

DavidMStraub:
> I need to sleep more, no idea what I was thinking changing that code although replacement is cleaner.

Yes it was quite ambitious doing everything at once, don't burn out! :fire: But we're almost there I think :rocket: 

> Should the ?handle= and ?gramps_id= query parms on /person return the single object or the single object in a list? I was returning as a object in a list to be consistent with the endpoint itself but the parm is sort of intended to match /{object}/{handle}. Actually not sure the swagger lets you define more than one return schema so maybe I answered my own question.

Yes I would also say a list since the same pattern can be used for cases where more than one object is returned, e.g. `?tag=todo`. But why do we even have a `?handle=`? that seems redundant.

> Also since /translate returns possible list of translations would it be more consistent to use /translate/{lang} with ?strings=[] to get a translation?

Yes, looks good.

gramps_webapi/api/resources/emit.py
DavidMStraub: But `int` will fail very often, e.g. on lists or dicts.

I think it's cleaner if we just check explicitly what we want to remove: iterables with length zero and `None`; that's it. Right?

So I would do:

```python
if not self.strip_empty_keys or not is_null(value):
    data[key] = value

def is_null(value):
    if value is None:
        return True
    try:
        return len(value) == 0
    except TypeError:  # object has no length
        pass
    return False
```

gramps_webapi/api/resources/person.py
Nick-Hall: > Added a .pylintrc to disable C0330 and C0326

It is better to use readable settings rather than codes.  I have created PR #27 which adds the settings to a default configuration file.

Nick-Hall: > Added a .flake8 but did not add it to the pre-commit config. I think maybe we should.

The main Gramps project uses pylint rather than flake8.

Nick-Hall: > Do you want them all resolved before a merge?

No.  Use pylint for guidance.  We aim for a pylint score of 9.0 or higher in new code.

However, if there are good reasons to write code with warnings then do so.  There is no need to disable a warning locally in the code.  I would rather be able to see it.

A good example is avoiding the use of variables `x`, `y` and `z`.  These are normally not good choices for variable names, but if you are writing graphics code then they may be appropriate.

gramps_webapi/api/resources/emit.py
cdhorn: Yes you are right, done. I did not know you could nest functions, that is neat!  So much to learn...

cdhorn:
Have been working on the apispec and I think I recall you used metadata in your original David so it seemed logical to do that here and merge the types into it since they're kind of related to the application configuration. If you think they belong separate I can break out again.

cdhorn:
Blech, pushed the apispec I was editing. Didn't mean to add that to a commit yet.

cdhorn:
Large chunk of overhaul of apispec.yaml is done. I'm wondering if I should pull examples from something like that.

DavidMStraub:
> I'm wondering if there is a good test or sample Gramps database that has a tree that is using all the features to full extent as never thought to look. Probably should pull all the examples from something like. Any thoughts?

Yes, good idea. There is [example.gramps](https://raw.githubusercontent.com/gramps-project/gramps/master/example/gramps/example.gramps).

DavidMStraub:
@cdhorn is something still missing to merge this? It doesn't have to be 100% perfect, but once merged it will be simpler to move forward implementing other features. 100 comments and 12 closed issues is a record for a PR anyway :rocket:

cdhorn:
I don't think so if you think it is okay now. There may be a bug in the keys handling, I ran into some anomalies working through the swagger update.  But can fix later.  I just commited what I have for the swagger for now as all the endpoints and models are in it. I am going to spend more time with the example.gramps and overhaul the examples in it, that can be a separate PR.

cdhorn:
Finished adding the examples tonight so that should be mostly squared away for the moment.

DavidMStraub:
Great! @Nick-Hall agree to merge?

Nick-Hall:
Yes.  Feel free to merge.

Issue #5, opened 2020-09-29
Add database to use for unit tests


DavidMStraub:

cdhorn:
Probably makes sense to use example.gramps for this too.

DavidMStraub:
Closed by https://github.com/gramps-project/web-api/pull/29.

PR #28, opened 2020-10-13
Add gramps.example (work in progress) #5

DavidMStraub:
This is a work in progress. The first commit finds the `gramps.example` from Gramps and loads it, returning a database instance to be used for unit test.

What is still missing is being able to use this also with unit tests that instantiate the app, in which case a "proper" file-based database is needed.


DavidMStraub:
Used the wrong branch, sorry. Need to redo.

PR #29, opened 2020-10-13
 Add gramps.example (work in progress) #5

DavidMStraub:
This is a work in progress. The first commit finds the `gramps.example` from Gramps and loads it, returning a database instance to be used for unit test. (It takes the gzipped version from the `doc` folder and unzips it to a temporary file, since the ungzipped one from the `tests` folder that is used in the `gramps` unit tests is not available for an installed `gramps`.)

What is still missing is being able to use this also with unit tests that instantiate the app, in which case a "proper" file-based database is needed.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,1 +1,38 @@
 """Unit test package for gramps_webapi."""
+
+import gzip
+import os
+import tempfile
+import unittest
+
+from gramps.gen.db.base import DbReadBase
+from gramps.gen.db.utils import import_as_dict
+from gramps.gen.user import User
+from gramps.gen.utils.resourcepath import ResourcePath
+
+
+class ExampleDb:
+    """Gramps example database handler."""
+
+    def __init__(self) -> None:
+        """Initialize self."""
+        _resources = ResourcePath()
+        doc_dir = _resources.doc_dir
+        self.path_gz = os.path.join(doc_dir, "example", "gramps", "example.gramps.gz")
+        if not os.path.isfile(self.path_gz):
+            raise ValueError("example.gramps not found at {}".format(self.path_gz))
+        self.path = None
+
+    def load(self) -> DbReadBase:
+        """Extract the example DB to a temp file and return a DB instance."""
+        with gzip.open(self.path_gz, "rb") as f_gzip:
+            file_content = f_gzip.read()
+            with tempfile.NamedTemporaryFile(delete=False, suffix=".gramps") as f:
+                f.write(file_content)
+                self.path = f.name
+        return import_as_dict(self.path, User())
+
+    def close(self) -> None:
+        """Delete the temporary file."""
+        if self.path is not None:
+            os.remove(self.path)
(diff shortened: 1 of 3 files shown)


DavidMStraub:
I'm confused. My virtualenv install of the wheel (and also the one at Travis) has an `example.gramps` at `ResourcePath().doc_dir + '/examples/gramps/'`, while my Ubuntu `.deb` installed package has an `example.gramps.gz` in the same place. @Nick-Hall do you know why?

It would be great being able to avoid shipping another copy of `example.gramps` with this package.

Nick-Hall:
It's just a Debian convention that example files are zipped, so they zip it on installation.

You don't have to unzip it though.  The importer will detect if it is zipped and unzip if necessary.

DavidMStraub:
> You don't have to unzip it though. The importer will detect if it is zipped and unzip if necessary.

But if I just to `import_as_dict("path/to/example.gramps.gz", User())`, I get `None`. I guess what you mean is not using the `import_as_dict` function but importing the example tree into a new database?

Nick-Hall:
The [`open_file`](https://github.com/gramps-project/gramps/blob/master/gramps/plugins/importer/importxml.py#L416) method of the importer detects zipped files, but the [`import_from_filename`](https://github.com/gramps-project/gramps/blob/master/gramps/gen/db/utils.py#L154) utility function looks at the [extension](https://github.com/gramps-project/gramps/blob/master/gramps/plugins/importer/import.gpr.py#L121).

It isn't very user friendly to expect a Debian user to have to rename the example file.  I need to think of the best solution.

DavidMStraub:
> It isn't very user friendly to expect a Debian user to have to rename the example file. I need to think of the best solution.

The way I've implemented it so far is to automatically extract it to a temporary directory which is then deleted again, so there is nothing to do for the user. Now that I know that it can sometimes be zipped and sometimes not, I can simply check that and only do the extraction if necessary.

Nick-Hall:
Yes.  An alternative would be to create a copy or symlink with the ".gramps" extension.

I was thinking more for users wanting to open the example file directly.

DavidMStraub:
I now handle both a zipped and unzipped example tree file and added two handlers to be used in unit test:

- an in-memory DB
- an SQLite DB in a temporary directory to which the Gramps database path is set temporarily.

DavidMStraub:
I should not that importing the example tree takes roughly 8 seconds on my computer, so while we can use the new classes in `setUpClass` and `tearDownClass` of test cases, we shouldn't do that too often. Maybe we can have one big test case (or maybe two) where we use the example DB.

cdhorn:
Looks good to me. I've never really worked with unittest before, will be good to learn more about it.  Looking at it quickly my initial thought is it probably makes sense to use setUpModule and tearDownModule and then have a TestCase class for each of the endpoints in the module.

DavidMStraub:
Awesome, didn't know about these. (I'm using `unittest` a lot but I guess I've never read the documentation properly :wink:)

DavidMStraub:
I tried to make use of `setupClass` but it didn't really work well since it doesn't allow storing any information to use in `tearDownClass` (except perhaps as global variables), so I think it will be better using `setUpClass` after all. Since this code doesn't affect any of the feature branches, I'll go ahead and merge it.

PR #30, opened 2020-10-18
Add filtering support, rename endpoints, refactor object extend (closes #17)

### LINKED ISSUE #17: Filter list-like endpoints with URL arguments

cdhorn:
@DavidMStraub @Nick-Hall I think this may be ready to merge if it passes muster. It does the following:

1. Renames all the endpoints to their plural form as they represent collections and not singular objects. Note I did not rename the source files in the resources directory, those are still singlar. ie person.py.
2. Adds support for the filtering framework by adding the /filters/{{namespace}} endpoint with the ability to get/create/update/delete custom filters and adding the filter= and rules= query parameters to the people, families, events, places, citations, sources, repositories, media and notes endpoints. filter= take the name of a custom filter. rules= takes a JSON structured argument to build a filter dynamically. 
3. Refactors object_extend to eliminate replicated code.
4. Refactors some of the boolean query parms into strings so the presence of the keyword is sufficient. ie profile, extend, strip.
5. Updates apispec.yaml to reflect the above changes and fix a couple issues.

This again tweaks the test cases to pass but does not add new ones. Once this is in and the example.gramps is merged I plan on focusing on test cases for everything done to date.

### DIFF
--- a/gramps_webapi/api/resources/base.py
+++ b/gramps_webapi/api/resources/base.py
@@ -97,12 +97,6 @@ class GrampsObjectsResource(GrampsObjectResourceHelper, Resource):
             if obj is None:
                 abort(404)
             return self.response(200, [self.object_extend(obj, args)], args)
-        if "handle" in args:
-            try:
-                obj = self.get_object_from_handle(args["handle"])
-            except HandleError:
-                abort(404)
-            return self.response(200, [self.object_extend(obj, args)], args)
         query_method = self.db_handle.method(
             "get_%s_from_handle", self.gramps_class_name
         )
--- a/gramps_webapi/api/resources/base.py
+++ b/gramps_webapi/api/resources/base.py
@@ -7,12 +7,14 @@ from flask import Response, abort
 from gramps.gen.db.base import DbReadBase
 from gramps.gen.errors import HandleError
 from gramps.gen.lib.primaryobj import BasicPrimaryObject as GrampsObject
-from webargs import fields
+from webargs import fields, validate
 from webargs.flaskparser import use_args
 
 from ..util import get_dbstate
 from . import ProtectedResource, Resource
 from .emit import GrampsJSONEncoder
+from .filters import apply_filter
+from .util import get_extended_attributes
 
 
 class GrampsObjectResourceHelper(GrampsJSONEncoder):
@@ -23,9 +25,11 @@ class GrampsObjectResourceHelper(GrampsJSONEncoder):
     def gramps_class_name(self):
         """To be set on child classes."""
 
-    @abstractmethod
     def object_extend(self, obj: GrampsObject, args: Dict) -> GrampsObject:
         """Extend the base object attributes as needed."""
+        if "extend" in args:
+            obj.extended = get_extended_attributes(self.db_handle, obj)
+        return obj
 
     @property
     def db_handle(self) -> DbReadBase:
@@ -52,11 +56,11 @@ class GrampsObjectResource(GrampsObjectResourceHelper, Resource):
 
     @use_args(
         {
-            "strip": fields.Boolean(missing=False),
-            "keys": fields.DelimitedList(fields.Str()),
+            "strip": fields.Str(validate=validate.Length(equal=0)),
+            "keys": fields.DelimitedList(fields.Str(), missing=[]),
             "skipkeys": fields.DelimitedList(fields.Str(), missing=[]),
-            "profile": fields.Boolean(missing=False),
-            "extend": fields.Boolean(missing=False),
+            "profile": fields.Str(validate=validate.Length(equal=0)),
+            "extend": fields.Str(validate=validate.Length(equal=0)),
         },
         location="query",
     )
@@ -65,8 +69,8 @@ class GrampsObjectResource(GrampsObjectResourceHelper, Resource):
         try:
             obj = self.get_object_from_handle(handle)
         except HandleError:
-            return abort(404)
-        return self.response(self.object_extend(obj, args), args)
+            abort(404)
+        return self.response(200, self.object_extend(obj, args), args)
 
 
 class GrampsObjectsResource(GrampsObjectResourceHelper, Resource):
@@ -74,13 +78,15 @@ class GrampsObjectsResource(GrampsObjectResourceHelper, Resource):
 
     @use_args(
         {
-            "gramps_id": fields.Str(),
-            "handle": fields.Str(),
-            "strip": fields.Boolean(missing=False),
-            "keys": fields.DelimitedList(fields.Str()),
+            "gramps_id": fields.Str(validate=validate.Length(min=1)),
+            "handle": fields.Str(validate=validate.Length(min=1)),
+            "strip": fields.Str(validate=validate.Length(equal=0)),
+            "keys": fields.DelimitedList(fields.Str(), missing=[]),
             "skipkeys": fields.DelimitedList(fields.Str(), missing=[]),
-            "profile": fields.Boolean(missing=False),
-            "extend": fields.Boolean(missing=False),
+            "profile": fields.Str(validate=validate.Length(equal=0)),
+            "extend": fields.Str(validate=validate.Length(equal=0)),
+            "filter": fields.Str(validate=validate.Length(min=1)),
+            "rules": fields.Str(validate=validate.Length(min=1)),
         },
         location="query",
     )
@@ -89,19 +95,30 @@ class GrampsObjectsResource(GrampsObjectResourceHelper, Resource):
         if "gramps_id" in args:
             obj = self.get_object_from_gramps_id(args["gramps_id"])
             if obj is None:
-                return abort(404)
-            return self.response([self.object_extend(obj, args)], args)
+                abort(404)
+            return self.response(200, [self.object_extend(obj, args)], args)
         if "handle" in args:
             try:
                 obj = self.get_object_from_handle(args["handle"])
             except HandleError:
-                return abort(404)
-            return self.response([self.object_extend(obj, args)], args)
-        iter_method = self.db_handle.method("iter_%s_handles", self.gramps_class_name)
+                abort(404)
+            return self.response(200, [self.object_extend(obj, args)], args)
         query_method = self.db_handle.method(
             "get_%s_from_handle", self.gramps_class_name
         )
+        if "filter" in args or "rules" in args:
+            handle_list = apply_filter(self.db_handle, args, self.gramps_class_name)
+            return self.response(
+                200,
+                [
+                    self.object_extend(query_method(handle), args)
+                    for handle in handle_list
+                ],
+                args,
+            )
+        iter_method = self.db_handle.method("iter_%s_handles", self.gramps_class_name)
         return self.response(
+            200,
             [
                 self.object_extend(query_method(handle), args)
                 for handle in iter_method()
--- a/gramps_webapi/api/resources/emit.py
+++ b/gramps_webapi/api/resources/emit.py
@@ -26,11 +26,14 @@ class GrampsJSONEncoder(JSONEncoder):
             getattr(lib, key) for key, value in inspect.getmembers(lib, inspect.isclass)
         ]
 
-    def response(self, payload: Any, args: Dict = {}) -> Response:
-        # pylint:disable=dangerous-default-value
+    def response(
+        self, status: int = 200, payload: Any = {}, args: Dict = {}
+    ) -> Response:
         """Prepare response."""
         if "strip" in args:
-            self.strip_empty_keys = args["strip"]
+            self.strip_empty_keys = True
+        else:
+            self.strip_empty_keys = False
         if "keys" in args:
             self.filter_only_keys = args["keys"]
         if "skipkeys" in args:
@@ -38,7 +41,7 @@ class GrampsJSONEncoder(JSONEncoder):
 
         return Response(
             response=self.encode(payload),
-            status=200,
+            status=status,
             mimetype="application/json",
         )
 
--- a/gramps_webapi/api/resources/emit.py
+++ b/gramps_webapi/api/resources/emit.py
@@ -1,7 +1,7 @@
 """Gramps Json Encoder."""
 
 import inspect
-from typing import Any, Dict
+from typing import Any, Dict, Optional
 
 import gramps.gen.lib as lib
 from flask import Response
@@ -27,9 +27,10 @@ class GrampsJSONEncoder(JSONEncoder):
         ]
 
     def response(
-        self, status: int = 200, payload: Any = {}, args: Dict = {}
+        self, status: int = 200, payload: Any = {}, args: Optional[Dict] = None
     ) -> Response:
         """Prepare response."""
+        args = args or {}
         if "strip" in args:
             self.strip_empty_keys = True
         else:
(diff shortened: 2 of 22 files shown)


DavidMStraub:
Thanks!

Again, please try to separate features/changes in the future. E.g. having the renaming of *all* endpoints plus the massive filtering feature in a single PR makes it hard to review.

cdhorn:
Yes, I know, going forward I will try to keep them smaller and more specific like you did with the example.gramps work.

gramps_webapi/api/resources/base.py
```diff
@@ -65,22 +69,24 @@ def get(self, args: Dict, handle: str) -> Response:
         try:
```
DavidMStraub, line 69: This method is getting really long; it would be more readable by putting the various `if` cases into separate methods. But I leave it up to you.

gramps_webapi/api/resources/emit.py
```diff
@@ -26,19 +26,22 @@ def __init__(self):
             getattr(lib, key) for key, value in inspect.getmembers(lib, inspect.isclass)
         ]
 
-    def response(self, payload: Any, args: Dict = {}) -> Response:
-        # pylint:disable=dangerous-default-value
+    def response(
+        self, status: int = 200, payload: Any = {}, args: Dict = {}
+    ) -> Response:
```
DavidMStraub, line 31: *Never* use an empty dict as a default function argument. It is mutable and any modification that you make will persist. (See e.g. [here](https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python)).

Use:

```python
def f(args: Optional[Dict] = None):
    args = args or {}
```

gramps_webapi/api/resources/event.py
```diff
@@ -24,9 +24,9 @@ class EventResourceHelper(GrampsObjectResourceHelper):
     def object_extend(self, obj: Event, args: Dict) -> Event:
         """Extend event attributes as needed."""
         db_handle = self.db_handle
-        if args["profile"]:
+        if "profile" in args:
```
DavidMStraub, line 27: Why is this no longer a boolean? Will `profile=0` be interpreted as `True`?

DavidMStraub:
I looked at everything *except* the filtering feature and it looks good! Only a couple of minor comments.

Concerning the filters:
- Can you give one or two example URLs to make it easier to understand how it works?
- Do you plan to add unit tests?

gramps_webapi/api/resources/base.py
cdhorn: This will need to be refactored for sorting so can address it then.  I think I agree with you, handle goes away. With the HasIdOf rule the gramps_id is not needed either.

gramps_webapi/api/resources/emit.py
cdhorn: I will fix this.

gramps_webapi/api/resources/event.py
cdhorn: If profile is not present in the parms then it is not generated, if ?profile is in the parms it is generated. No need for =0 or =1. I think it is cleaner. Same for the others that were booleans.

cdhorn:
Well, that's embarassing, indented something too far. Please use latest, I also rebased and fixed the empty dict default arg. 

I included what I hoped was a clear and concise description for the rules= query parm in the apispec.yaml in this PR, please review it and let me know if you think it is sufficient or could be worded better.

Before starting examine the filters people endpoint and look over the available rules:
`api/filters/people`

This constructs a filter for a specific gramps_id, hence the reason I think we no longer need it although it is a convenient "shorthand":
`api/people/?rules={"r":[{"n":"HasIdOf","v":["P1"]}]}`

This constructs a filter on all males with the tag named "DNA Match" who are marked private:
`api/people/?rules=[{"n":"HasTag","v":["DNA Match"]},{"n":"IsMale"},{"n":"PeoplePrivate"}]}`

This constructs a filter to return people matching tag "Direct Ancestor" or "Military":
`api/people/?rules={"f":"or","r":[{"n":"HasTag","v":["Direct Ancestor"]},{"n":"HasTag","v":["Military"]}]}`

This constructs a filter to return all female ancestors of person with gramps_id P1:
`api/people/?rules={"r":[{"n":"IsAncestorOf","v":["P1"]},{"n":"IsFemale"}]}`

Let us suppose we used POST and created a named custom filter "FemaleAncestors" with the above filter. (Note for POST it expects full keywords in the JSON and not the shorthand characters used in the query parm to help shorten it)

This then would let us run the custom filter directly as Nick suggested:
`api/people/?filter=FemaleAncestors`

This then uses that custom filter in another filter we are constructing to fetch all descendents of the female ancestors of the person:
`api/people/?rules={"r":[{"n":"IsDescendantOfFilterMatch","v":["FemaleAncestors"]}]}`

The thought occured to me we might provide a set of "shorthand" query parms for common filters, like the tags= example, but with such rich functionality available with rules= are they really needed?


Regarding unit tests, as I previously mentioned my plan is to back track and focus on unit tests for **_all_** of the existing endpoints **_and_** query parms, including everything in this PR, as I said I would do so and see that as my responsibility.  Rather than one big PR I will try to submit those in smaller chunks, I think an endpoint at a time for the major objects and then one final one for all the loose ends makes the most sense.

There will probably be one small PR before then once this is merged to modify the extend query parm to accept a optional list of specific keys to process. For example, extend with no arguments currently defaults to everything, but extend=events,media on a person would only return those two. I did not think of doing that when I refactored it and it is so obvious it should be done that way in hindsight to make it more flexible.

The reason I would submit that one before diving into test cases is so as I work through the unit tests for each end point I can include the necessary cases for that as well. I considered adding it to this PR as I did refactor that code as part of this but decided it could just be a quick follow on. 

Let me know if my current plans sound agreeable to you. Actually I will open the test case issues tonight.

DavidMStraub:
> I included what I hoped was a clear and concise description for the rules= query parm in the apispec.yaml in this PR, please review it and let me know if you think it is sufficient or could be worded better.

Sorry for having missed that!

> Before starting examine the filters people endpoint and look over the available rules:
> `api/filters/people`
> 
> This constructs a filter for a specific gramps_id, hence the reason I think we no longer need it although it is a convenient "shorthand":
> `api/people/?rules={"r":[{"n":"HasIdOf","v":["P1"]}]}`
> 
> This constructs a filter on all males with the tag named "DNA Match" who are marked private:
> `api/people/?rules=[{"n":"HasTag","v":["DNA Match"]},{"n":"IsMale"},{"n":"PeoplePrivate"}]}`
> 
> This constructs a filter to return people matching tag "Direct Ancestor" or "Military":
> `api/people/?rules={"f":"or","r":[{"n":"HasTag","v":["Direct Ancestor"]},{"n":"HasTag","v":["Military"]}]}`
> 
> This constructs a filter to return all female ancestors of person with gramps_id P1:
> `api/people/?rules={"r":[{"n":"IsAncestorOf","v":["P1"]},{"n":"IsFemale"}]}`

That's really cool, I like it!

What I don't quite understand is the rationale for the shorthands ... given the response will typically be orders of magnitude larger than the request, bandwidth cannot be an issue and I don't think it is more readable either. I also don't think there is any relevant limit on the length of the query string. And the code would be much simpler without translating between short and long forms. Why did you introduce it?

> Let us suppose we used POST and created a named custom filter "FemaleAncestors" with the above filter. (Note for POST it expects full keywords in the JSON and not the shorthand characters used in the query parm to help shorten it)
> 
> This then would let us run the custom filter directly as Nick suggested:
> `api/people/?filter=FemaleAncestors`
> 
> This then uses that custom filter in another filter we are constructing to fetch all descendents of the female ancestors of the person:
> `api/people/?rules={"r":[{"n":"IsDescendantOfFilterMatch","v":["FemaleAncestors"]}]}`

Yes, this is great. I think those are two separate and both very well motivated use cases:

1. An app user creating a custom filter, like in desktop Gramps, and using it in reports/searches/other filters
2. The app itself using the filter to request a subset of items, e.g. for filtering a long list or for displaying search results.

In the second case, the `GET` endpoint is perfect since it doesn't create anything persistent, while the `POST` endpoint implements the first case.

> The thought occured to me we might provide a set of "shorthand" query parms for common filters, like the tags= example, but with such rich functionality available with rules= are they really needed?

Yes, good point. Let's leave it out for the moment in order not to complicate things. Probably it's not necessary anyway.

> Regarding unit tests, as I previously mentioned my plan is to back track and focus on unit tests for **_all_** of the existing endpoints **_and_** query parms, including everything in this PR, as I said I would do so and see that as my responsibility. Rather than one big PR I will try to submit those in smaller chunks, I think an endpoint at a time for the major objects and then one final one for all the loose ends makes the most sense.

:+1:

gramps_webapi/api/resources/filters.py
```diff
@@ -0,0 +1,298 @@
+"""Gramps filter interface."""
+
+import json
+from typing import Dict, List, Set
+
+import gramps.gen.filters as filters
+from flask import Response, abort
+from gramps.gen.db.base import DbReadBase
+from gramps.gen.filters import GenericFilter
+from marshmallow import Schema, ValidationError, fields
```
DavidMStraub, line 10: Have you forgotten adding `marshmallow` to `setup.py`? I don't know if there are any tests importing this module yet...

gramps_webapi/api/resources/filters.py
```diff
@@ -0,0 +1,298 @@
...
+
+class FilterResource(ProtectedResource, GrampsJSONEncoder):
+    """Filter resource."""
+
+    @use_args(
+        {"filter": fields.Str(), "rule": fields.Str()},
+        location="query",
+    )
+    def get(self, args: Dict, namespace: str) -> Response:
+        """Get available custom filters and rules."""
```
DavidMStraub, line 164: This is a loooong method. Possible to refactor? We can also do it later, but also for unit tests it's difficult having so long methods where we cannot test individual pieces.

gramps_webapi/api/resources/filters.py
```diff
@@ -0,0 +1,298 @@
...
+
+    @use_args(FilterSchema(), location="json")
+    def post(self, args: Dict, namespace: str) -> Response:
+        """Create a custom filter."""
+        try:
+            namespace = GRAMPS_NAMESPACES[namespace]
+        except KeyError:
+            abort(404)
+
+        new_filter = build_filter(args, namespace, keys="long")
```
DavidMStraub: A function argument that only has two valid values (in this case "long" and "short") cries out for being a boolean rather than a string where all except two values are undefined. `long_keys=True`.

gramps_webapi/api/resources/filters.py
```diff
@@ -0,0 +1,298 @@
...
+        return abort(404)
+
+    @use_args(
+        {
+            "name": fields.Str(required=True),
+            "force": fields.Boolean(required=False, missing=False),
+        },
+        location="json",
+    )
+    def delete(self, args: Dict, namespace: str) -> Response:
```
DavidMStraub, line 222: It's great that you're using type hints everywhere, it makes the code much more easy to understand.

For dicts, it's also helpful to indicate the type of the keys (and possibly values); this also catches possible bugs early in my experience. In this case, problably `Dict[str, str]`.

gramps_webapi/api/resources/filters.py
```diff
@@ -0,0 +1,298 @@
...
+        filters.reload_custom_filters()
+        for filter_rule in filters.CustomFilters.get_filters(namespace):
+            if new_filter.get_name() == filter_rule.get_name():
+                abort(422)
+        filters.CustomFilters.add(namespace, new_filter)
+        filters.CustomFilters.save()
+        return self.response(201, {"message": "Added filter: " + new_filter.get_name()})
+
+    @use_args(FilterSchema(), location="json")
+    def put(self, args: Dict, namespace: str) -> Response:
```
DavidMStraub, line 196: Another long method.

DavidMStraub:
Done reviewing :slightly_smiling_face:

cdhorn:
> What I don't quite understand is the rationale for the shorthands ... Why did you introduce it?

In all honesty I had the query parms worked out first and thought keeping them short made sense, and then decided to use the full keywords when I got around to the other code. I agree it makes sense to get rid of them, it is less code and probably clearer. I will refactor in a little while and push a commit doing so later today.

With rules= functionality do you object to me removing gramps_id= and handle= from GrampsObjectsResource now as well when I do so?

gramps_webapi/api/resources/filters.py
cdhorn: It is required for webargs, do you think it needs to be added separately?

gramps_webapi/api/resources/filters.py
cdhorn: I'm not sure criteria for what is considered long. I guess the two functions it performs could be broken out and called separately.

gramps_webapi/api/resources/filters.py
cdhorn: Yes, good point. That will go away with a refactor.

gramps_webapi/api/resources/filters.py
cdhorn: Yes, I am trying to get in the practice of doing so, they are helpful. Good idea on the Dict hints, did not occur to me, I will try to do that as well.

DavidMStraub:
> In all honesty I had the query parms worked out first and thought keeping them short made sense, and then decided to use the full keywords when I got around to the other code. I agree it makes sense to get rid of them, it is less code and probably clearer. I will refactor in a little while and push a commit doing so later today.

:+1:

> With rules= functionality do you object to me removing gramps_id= and handle= from GrampsObjectsResource now as well when I do so?

handle definitely, for this we have a dedicated endpoint.

For `gramps_id`, I am more worried about performance. I find it quite likely that an app might choose to call individual objects by their Gramps IDs rather than handles (e.g. this is how my existing app does it). And while the filter rule with `HasIdOf` and an implementation using `get_person_from_gramps_id` (taking person as an example) return the same result, their implementations are veryy different: the filter implementation will fetch **all** people in the database, construct `Person` instances for each of them, iterate through the created instances and return a single one. The `..._from_gramps_id` method will instead use a database query with `WHERE` clause, which I expect to be *much* faster (even though I haven't tried it and even though the default SQLite DB doesn't have any indices, so the full table must be read anyway). But admittedly from an API point of view it doesn't make much sense to duplicate this functionality.

DavidMStraub:
And perhaps we shouldn't do premature optimization at the expense of transparency :slightly_smiling_face:

cdhorn:
Looked closer at that, yes you are right there is a definite benefit to keeping gramps_id as is. I will pull handle out though, as you previously mentioned it is unnecessary.

Nick-Hall:
> The ..._from_gramps_id method will instead use a database query with WHERE clause, which I expect to be much faster (even though I haven't tried it and even though the default SQLite DB doesn't have any indices, so the full table must be read anyway).

The `gramps_id` columns are indexed and these methods will be fast.

cdhorn:
Also refactored get method in FilterResource class.

DavidMStraub:
So, all set already?

cdhorn:
If everything passes muster with you, sure! :-)

Issue #33, opened 2020-10-24
Modify extend query parm to accept optional keyword list


cdhorn:
Currently the extend option returns all the objects referenced by an object and not a subset. This will modify it to accept a comma delimited keyword list of specific ones to return so if a user only wanted events & media for a person and not all the rest they can request it.

cdhorn:
Closing, PR #50 merged

PR #50, opened 2020-10-25
Refactor extend to accept keyword list, fix emit bug (Closes Issue #33)

cdhorn:
This refactors the extend query parm itself so if provided with no args all records are returned as it currently behaves but if provided a comma delimited list of keywords only that subset is extracted and returned with the base record. Also fixes a missed case in emit with keys= handling.  The apispec.yaml has also been updated as appropriate.

### DIFF
--- a/gramps_webapi/api/resources/citation.py
+++ b/gramps_webapi/api/resources/citation.py
@@ -21,8 +21,15 @@ class CitationResourceHelper(GrampsObjectResourceHelper):
         """Extend citation attributes as needed."""
         if "extend" in args:
             db_handle = self.db_handle
-            obj.extended = get_extended_attributes(db_handle, obj)
-            obj.extended["source"] = get_source_by_handle(db_handle, obj.source_handle)
+            obj.extended = get_extended_attributes(db_handle, obj, args)
+            if (
+                "all" in args["extend"]
+                or "" in args["extend"]
+                or "source" in args["extend"]
+            ):
+                obj.extended["source"] = get_source_by_handle(
+                    db_handle, obj.source_handle, args
+                )
         return obj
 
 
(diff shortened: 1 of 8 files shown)


gramps_webapi/api/resources/citation.py
```diff
@@ -21,8 +21,15 @@ def object_extend(self, obj: Citation, args: Dict) -> Citation:
         """Extend citation attributes as needed."""
         if "extend" in args:
             db_handle = self.db_handle
-            obj.extended = get_extended_attributes(db_handle, obj)
-            obj.extended["source"] = get_source_by_handle(db_handle, obj.source_handle)
+            obj.extended = get_extended_attributes(db_handle, obj, args)
+            if (
+                "all" in args["extend"]
+                or "" in args["extend"]
```
DavidMStraub: What is the rationale for the `"" in args["extend"]`? Is this in alias for `all`? It would trigger in cases like `extend=place,,father` (double comma).

cdhorn: Yes that handles the ?extend case and treats it the same as extend=all

I had thought webargs DelimitedList should let me set the default if missing and that didn't seem to work and I wanted to try to preserve the other behaviour. A check of marshmallow docs doesn't show a DelimitedList so it is a webargs construct.

Good point on the double comma, I had not thought of that. I should change to `args["extend"] == [""]` or do you suggest a better way of handling?

DavidMStraub: Did you check that `?extend` actually parses as `[""]`? And can this be relied on? It seems like a hack to me. `extend=all` is clean and unambiguos. And we don't have to worry about backwards compatibility :hand_over_mouth:

cdhorn: I did and it does, but if you think it is cleaner I'll drop support for the plain jane ?extend case.

DavidMStraub: I leave it up to you.

cdhorn: I trust your instinct on dropping it, doing so now.

cdhorn:
There, should be good to go if you approve.

PR #51, opened 2020-10-30
WIP: media files (fixes #13).

### LINKED ISSUE #13: Media file and thumbnail endpoints

DavidMStraub:
This is work in progress of implementing #13.

While I have been struggling with unit tests, I think I'll wait for @cdhorn's next PR and use the same approach :wink: 

Implemented so far: the endpoint `/api/media/somehandle/file` that simply downloads the file.

Implementation notes:

- The class `LocalFileHandler` inheriting from `FileHandler` might seem like overkill but it will allow generalizing this to cloud storage as done [here for S3](https://github.com/DavidMStraub/gramps-webapp/blob/master/gramps_webapp/media.py#L40-L116)
- When using local files, by using `flask.send_from_directory`, it is guaranteed that only files that are actually contained in the media base folder will be accessible. This prevents users from adding a media object with an absolute media path pointing to, say, `/etc/passwd`.
- The question is whether `MEDIA_BASE_DIR` *must* be specified or whether we should default to the directory in the Gramps settings. When using the API locally with an existing Gramps installation, that actually makes sense. When used on a server, it makes less sense.

### DIFF
--- a/gramps_webapi/api/__init__.py
+++ b/gramps_webapi/api/__init__.py
@@ -3,6 +3,8 @@
 from typing import Type
 
 from flask import Blueprint, current_app
+from webargs import fields, validate
+from webargs.flaskparser import use_args
 
 from ..const import API_PREFIX
 from .auth import jwt_required_ifauth
@@ -24,6 +26,7 @@ from .resources.source import SourceResource, SourcesResource
 from .resources.tag import TagResource, TagsResource
 from .resources.token import TokenRefreshResource, TokenResource
 from .resources.translate import TranslationResource, TranslationsResource
+
 
 api_blueprint = Blueprint("api", __name__, url_prefix=API_PREFIX)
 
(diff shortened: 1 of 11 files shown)


DavidMStraub:
Forgot to mention, had to move the auth decorators out of `api.resources` since they are now used in `api.__init__`.

DavidMStraub:
This is now complete from my side except for unit tests. 

I implemented the following endpoints as discussed in #13:

> - `/media/<string:handle>/file` for the file
>- `/media/<string:handle>/thumbnail/<width:int>/<height:int>` for a thumbnail fitting into width/height
>- `/media/<string:handle>/cropped/<x1:int>/<y1:int>/<x2:int>/<y2:int>` for a cropped image
>- `/media/<string:handle>/cropped/<x1:int>/<y1:int>/<x2:int>/<y2:int>/thumbnail/<width:int>/<height:int>` for the thumbnail of a cropped image

For PDF files, `file` returns the PDF itself, while the other 3 endpoints refer to an image of the first page.

On all the thumbnail/cropped endpoints, I also added a boolean query argument `square`, which returns a the image cropped to a centered square (I found that useful for my web app, e.g. in profile or gallery views).

gramps_webapi/api/__init__.py
```diff
@@ -74,9 +79,55 @@ def register_endpt(resource: Type[Resource], url: str, name: str):
...
 )
 # Metadata
 register_endpt(MetadataResource, "/metadata/<string:datatype>", "metadata")
+
+# Media files
+@api_blueprint.route("/media/<string:handle>/file")
+@jwt_required_ifauth
+def download_file(handle):
+    """Download a file."""
+    base_dir = current_app.config["MEDIA_BASE_DIR"]
```
cdhorn, line 92: I know Gramps has get_mediapath/set_mediapath methods to store it in the database. How does that related to having the MEDIA_BASE_DIR config option?

cdhorn: Oh, future S3 support I guess. Maybe I just answered my own question.

cdhorn:
This looks pretty good to me!  I like that you handled PDF files too.  As for MEDIA_BASE_DIR, I do not think it should be required and falling back to the one stored in Gramps is fine.

DavidMStraub:
Sorry, the diff is now completely messed up since I rebased onto your testing branch. This will be cleaned up again after your PR is merged.

DavidMStraub:
Fixed, ready from my side

PR #53, opened 2020-11-01
Implement HTML note output (fixes #26)

### LINKED ISSUE #26: Notes HTML output
Add HTML output for notes. This can be based on https://github.com/DavidMStraub/gramps-webapp/blob/master/gramps_webapp/gramps.py#L531-L580, but support for internal links needs to be added.

Still to be decided how that fits into the API.

DavidMStraub:
As discussed in #26, I added HTML note output to the `/notes/` endpoint. The code is mostly taken from `gramps_webapp`, which in turn is partly taken from the narrative web plugin.

API details:
- Input: `/api/notes/?formats=html`, `/api/notes/myhandle/?formats=html` where `formats` can be a comma-delimited list in the future.
- Output: a key `formatted` is added to the output, which in turn contains a key `html` with the escaped HTML string (and possibly other formats in the future)

Implementation details:
- I added the `formats` key to the base object, even though it has no effect on any other endpoint. I did this mostly because I couldn't think of a better way to do it (which for sure doesn't mean there is none)
- The HTML output is sanitized with a white list of tags, attributes, and styles, to prevent security issues in web apps through injected Javascript etc. This could potentially become configurable through query arguments in the future, although I don't really see the need at the moment

Open issues:
- unit tests missing

### DIFF
--- a/gramps_webapi/api/html.py
+++ b/gramps_webapi/api/html.py
@@ -1,0 +1,119 @@
+"""HTML backend for styled text."""
+
+import bleach
+
+from gramps.gen.lib import Note, NoteType, StyledText
+from gramps.plugins.lib.libhtml import Html
+from gramps.plugins.lib.libhtmlbackend import HtmlBackend, process_spaces
+
+
+ALLOWED_TAGS = [
+    "a",
+    "abbr",
+    "acronym",
+    "b",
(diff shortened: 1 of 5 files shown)


DavidMStraub:
Done.

Issue #46, opened 2020-10-24
Unit tests for translations and translations/{code} endpoints


cdhorn:

cdhorn:
Closing, see PR #54

Issue #44, opened 2020-10-24
Unit tests for bookmarks and bookmarks/{namespace} endpoints


cdhorn:

cdhorn:
Closing, see PR #54

PR #54, opened 2020-11-01
End point unit tests (Closes #34 through #44 and #46)

### LINKED ISSUE #34: Finish unit tests for people and people/{handle} endpoints
Unit tests for filter, rules, strip, keys, skipkeys, profile, and extend options.

cdhorn:
@DavidMStraub if this meets your approval I'd like to get it merged so you can start making use of it as well.

There is still work to do, but this puts a sort of framework in place and is hopefully the largest chunk of it. 

I will be focusing on the /api/filters tests next.  The filter= parm is not tested in the base object endpoints, I think that will all be handled exclusively in /api/filters as need to create the filter first before can call it.  

Note there are numerous other changes in here to address bugs and other things as a result of having worked through these.

Also note no JWT tests in here either, but certainly needed to make sure all the endpoints are protected. I'm unsure of best approach to do that.  You had them separate from the unauthenticated API tests, do you think the same should be done here?  Modify things so this just includes that by default and have the one set of endpoint tests? Have a separate test_endpoints_auth subpackage but maybe only include the most basic of tests and not everything this is trying to do?

Also should test_api.py be removed since this sort of supercedes it?


cdhorn:
Hmm. I didn't think I changed anything to cause that.

E           ValueError: Database path for family tree 'example_gramps' not found

DavidMStraub:
It might be that this works only locally, but not on Travis. I had similar problems when trying to make this work. It's tricky because of the way Gramps handles paths. I'll look into it as well.

cdhorn:
> It might be that this works only locally, but not on Travis. I had similar problems when trying to make this work. It's tricky because of the way Gramps handles paths. I'll look into it as well.

Yeah I have no issues locally. I'm not that familiar with that part of the Gramps code, hope you are able to figure it out.

DavidMStraub:
What I suspect is that you have an `example_gramps` DB in you local Gramps installation, which is used by the unit test, but the `ExampleDbInMemory` instance that you are creating in the `setUpModule` is not used at all.

`WebDbManager` uses `CLIDbManager.get_family_tree_path`, which looks for an actual directory inside tha current Gramps `database.path`, and I don't think it's possible to work with `ExampleDbInMemory` at all. (This class was rather meant for testing functions that act on database instances, not the whole app). 

I tried to make unit tests work using `ExampleDbSQLite` and also had some issues, e.g. initially it complaoined that the sqlite plugin was not loaded.

cdhorn:
Ah, yup, renamed my .gramps to move it out of the way and get the same error locally now.

cdhorn:
Oh my, we don't just need the database to do this right, we kind of need the whole Gramps HOME_DIR structure.  For example, the custom filters are stored in custom_filters.xml in ~/.gramps/gramps51/custom_filters.xml

DavidMStraub:
For the Gramps home dir structure, we can use the `GRAMPSHOME` environment variable.

I just spent quite some time trying to get this to work (see my `test_example` branch) but I didn't manage :disappointed: 

@Nick-Hall, do you have an idea?

Basically, we want to have `GRAMPSHOME` in a temporary directory and import `gramps.example` into a new database called "example", which can be used in unit tests and cleaned up at the end.

cdhorn:
Yep, I was messing with that for a few hours last night after my last comment. The problem is by time we get to setting GRAMPSHOME some of the imports already loaded gramps.gen.const so the various DIR variables are set all over the place. I started playing with importlib.reload() before I had to go to bed, will mess with it more in a while.

cdhorn:
Well, I was able to get everything sorted out and working properly with a temporary GRAMPSHOME, and learned a lot doing so. I'm not sure if there is a better way of handling things though. Let me know what you think when you have time to look through it.

cdhorn:
FYI I checked and on Ubuntu while example.gramps is compressed the .jpg files are not, and after import the mediapath is set to `{GRAMPS_RESOURCES}/doc/gramps/example/gramps` so I think they can be used with unit tests as is. If you knew that already okay. I was looking as it occurred to me if needed I should copy the image files to the temporary GRAMPSHOME and decompress them as part of the fixture setup.

DavidMStraub:
That's absolutely brilliant, great job! :rocket: 

I also tried running the tests from you branch locally (Ubuntu) and almost all of them (182) worked, but I also get 12 failures. It looks like they all have the same cause: For many keys, I get translated strings rather than English ones (my system locale is German, `de_DE.UTF-8`), e.g. `Tod` instead of `Death`, `Geburt` instead of `Birth`, etc. 

I guess the easiest way to solve this would be to fix the Gramps locale to English in the unit tests; I will experiment how to achieve that.

DavidMStraub:
OK so I can make *all* tests work by adding

```python
os.environ["LANGUAGE"] = "en"
```

in `tests/__init__.py` *before* importing gramps (this is important as importing Gramps instantiates all kinds of singletons which are then hard to get rid of).

It's not exactly elegant, but since we only do it for unit tests, I would say it's fine!

Nick-Hall:
> Basically, we want to have GRAMPSHOME in a temporary directory and import gramps.example into a new database called "example", which can be used in unit tests and cleaned up at the end.

It looks like you have already found a solution.

You can set the `GRAMPSHOME` environment variable.  The default database directory will be in a `grampsdb` folder under it.

Then you can set the `database.path` configuration parameter if you want it somewhere else.

Nick-Hall:
> I also tried running the tests from you branch locally (Ubuntu) and almost all of them (182) worked, but I also get 12 failures. It looks like they all have the same cause: For many keys, I get translated strings rather than English ones (my system locale is German, de_DE.UTF-8), e.g. Tod instead of Death, Geburt instead of Birth, etc.

This would seem to be a bug in the JSON representation of a `GrampsType`.  Perhaps we should use the numeric code in addition to the translated string.  Alternatively, we could use the untranslated string for the pre-defined types.

cdhorn:
Can you try placing it in tests/test_endpoints/__init__.py in this code block before the reload?

```
    TEST_GRAMPSHOME = tempfile.mkdtemp()
    os.environ["GRAMPSHOME"] = TEST_GRAMPSHOME
    importlib.reload(gramps.gen.const)
    from gramps.gen.const import USER_DIRLIST
```
I think it would be better to set there maybe to make it clearer as to the purpose.

cdhorn:
If that works I can add a commit for it, or if you have looked things over and want to merge you can merge and add it after.

DavidMStraub:
> Can you try placing it in tests/test_endpoints/__init__.py in this code block before the reload?

Unfortunately that didn't work. I suspect it's because the `GrampsLocale` singleton is instantiated already in `tests/__init__.py`.

> This would seem to be a bug in the JSON representation of a GrampsType. Perhaps we should use the numeric code in addition to the translated string. Alternatively, we could use the untranslated string for the pre-defined types

If I understand `emit.py` correctly, @cdhorn you are not using the Gramps JSON serialization, but a generalization/modification of it imlemented in `GrampsJSONEncoder.api_filter`. Is that correct?

DavidMStraub:
>  Perhaps we should use the numeric code in addition to the translated string. Alternatively, we could use the untranslated string for the pre-defined types.

That's a very good point; in fact we might not want the types to be localized, right? 

I found out we can easily do this, without the above locale hack, by replacing `str(obj)` with `obj.xml_str()` in `emit.py` and `str(event.type)` by `event.type.xml_str()` in `util.py`.

With these changes, all unit tests except one pass for me *without* changing the locale: namely, the "after" in the date is localized in `test_people_endpoint_profile`. However, I think it *does* make sense for dates in "profiles" to be localized, as these would likely directly be used by apps (e.g. in list views) and cannot easily be translated (using e.g. the `translate` endpoint) like the event type names can.

cdhorn:
> If I understand emit.py correctly, @cdhorn you are not using the Gramps JSON serialization, but a generalization/modification of it imlemented in GrampsJSONEncoder.api_filter. Is that correct?

Yes, a modification that was inspired by `__default` in `gramps.lib.serialize`. Looking at that again I see `return str(obj)` on a Gramps type should probably have been `return getattr(obj, 'string')` further down in `GrampsJSONEncoder.default` which I reckon is the same as `obj.xml_str()`?  I'm not bi-lingual unfortunately and how to properly handle locale is something I'm not very familiar with, my apologies if I make mistakes with it.

Note looking at it again we do not want to handle dates the way `serialize.py` does as the swagger schema should be of one type.  You'll notice I had to normalize a few None values to "" in the `api_filter` method in this pull request otherwise the schema validation fails.

DavidMStraub:
> ooking at that again I see `return str(obj)` on a Gramps type should probably have been `return getattr(obj, 'string')` further down in `GrampsJSONEncoder.default` which I reckon is the same as `obj.xml_str()`? I



> Looking at that again I see `return str(obj)` on a Gramps type should probably have been `return getattr(obj, 'string')` further down in `GrampsJSONEncoder.default` which I reckon is the same as `obj.xml_str()`?

No, `str(obj)` and `getattr(obj, 'string')` are equivalent since `GrampsTypeMeta` sets

```python
string = property(__str__, __set_str, None, "Returns or sets string value")
```
The only difference between `__str__` (which is called by `str`) and `xml_str` is that the former calls `_I2SMAP`, which gives the translated event type names, while the latter calls `_I2EMAP`, which gives the untranslated ones.

cdhorn:
Just commited that change with xml_str(). 

Maybe we should add a locale parm to override that and return data using current locale.

Did you see anything else in the code that is problematic?

DavidMStraub:
> Just commited that change with xml_str().

Perfect.

> Maybe we should add a locale parm to override that and return data using current locale.

Yes, I will open an issue so we don't forget about it.

> Did you see anything else in the code that is problematic?

No, to be honest I didn't look at all the tests in detail, but there's lots of them and they pass which is what counts :wink: 

Let's merge this so we can start adding tests in other branches as well.

cdhorn:
Sounds good!

Issue #52, opened 2020-11-01
Backlinks


DavidMStraub:
This is to continue the discussion of backlinks started in #13.

Backlinks (i.e. finding objects referencing a given object) are useful for finding objects that reference a media object, but also much more generally.

So the question is how we implement those in the API.

As far as I can see, DBAPI implements `get_backlink_handle` by a SQL lookup of the `references` table, where both the referring and the referred handles are indexed in the default SQLite implementation, which is great, so this should be very fast.

I see the following options:

1. Always add all backlinks to all object endpoints (in spite of the above, this might lead to a slowdown - we would need to check)
2. Add a boolean query switch `backlinks` to enable backlinks on all object endpoints.
3. Add a `backlink` endpoint do fetch backlinks of a specific object.

In the case of 2., the backlinks of a given media object would be obtained by
`/api/media/myhandle/?backlink=1`
and looking for the "backlink" key.
In the case of 3. by
`/api/media/backlinks/myhandle`

cdhorn:
I think case 2, it can be enabled on /api/media and /api/media/{handle}. I agree a separate "backlinks" keyword in the base response is good. 

I think case 3 is not clean as that position in the path is for {handle} which can have any value.  You could do /api/media/{handle}/backlinks though. But I think 2 is more flexible if you enable it on both endpoints as the code will be there anyway.

DavidMStraub:
In progress at #55.

cdhorn:
PR #55 merged, closing issue

PR #55, opened 2020-11-04
Add backlinks (#52)

DavidMStraub:
I started implementing this. Notes:

- Output is 
```
...,
backlinks: {
  Family: [
    'abcd',
    'defgh',
    ...
  ],
  ...
},
...
```
- Do we need to add an iteration where handles are replaced by full objects or are handles enough?
- In order not to have to modify all the resource child classes, I replaced `object_extend` by a new method `full_object` which does `object_extend` (overloaded by child classes) plus backlinks. This is very ugly, but can be refactored later.

### DIFF
--- a/gramps_webapi/api/resources/base.py
+++ b/gramps_webapi/api/resources/base.py
@@ -27,9 +27,9 @@ class GrampsObjectResourceHelper(GrampsJSONEncoder):
 
     def full_object(self, obj: GrampsObject, args: Dict) -> GrampsObject:
         """Get the full object with extended attributes and backlinks."""
-        obj = self.object_extend(obj, args)
         if args.get("backlinks"):
             obj.backlinks = get_backlinks(self.db_handle, obj.handle)
+        obj = self.object_extend(obj, args)
         return obj
 
     def object_extend(self, obj: GrampsObject, args: Dict) -> GrampsObject:
--- a/gramps_webapi/api/resources/base.py
+++ b/gramps_webapi/api/resources/base.py
@@ -14,7 +14,7 @@ from ..util import get_dbstate
 from . import ProtectedResource, Resource
 from .emit import GrampsJSONEncoder
 from .filters import apply_filter
-from .util import get_extended_attributes
+from .util import get_backlinks, get_extended_attributes
 
 
 class GrampsObjectResourceHelper(GrampsJSONEncoder):
@@ -24,6 +24,13 @@ class GrampsObjectResourceHelper(GrampsJSONEncoder):
     @abstractmethod
     def gramps_class_name(self):
         """To be set on child classes."""
+
+    def full_object(self, obj: GrampsObject, args: Dict) -> GrampsObject:
+        """Get the full object with extended attributes and backlinks."""
+        obj = self.object_extend(obj, args)
+        if args.get("backlinks"):
+            obj.backlinks = get_backlinks(self.db_handle, obj.handle)
+        return obj
 
     def object_extend(self, obj: GrampsObject, args: Dict) -> GrampsObject:
         """Extend the base object attributes as needed."""
@@ -64,6 +71,7 @@ class GrampsObjectResource(GrampsObjectResourceHelper, Resource):
             "profile": fields.Str(validate=validate.Length(equal=0)),
             "extend": fields.DelimitedList(fields.Str(validate=validate.Length(min=1))),
             "formats": fields.DelimitedList(fields.Str()),
+            "backlinks": fields.Boolean(missing=False),
         },
         location="query",
     )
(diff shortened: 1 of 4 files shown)


DavidMStraub:
> Do we need to add an iteration where handles are replaced by full objects or are handles enough?

@cdhorn, I guess the way to do that would be to add it under `extended`,

```
...
'extended': {
  ...,
  'backlinks' {
    'Family': [
      {'gramps_id': ...},
      ...,
    ],
...
```
Do you think it should be done?

cdhorn:
Yes I think it is a good idea to give the user the option if they want it. 

Should the keyword be lowercase to be consistent with the rest of the keywords in the API?

DavidMStraub:
True, this is a raw output from the database, but it makes sense to convert it to lowercase.

DavidMStraub:
Now this is really strange. I only changed the case of the keyword, and now the previously successful test fails because there are different backlinks. How is that possible?

Will have to continue debugging tomorrow. :sleeping:

DavidMStraub:
Fixed the unit tests and added `backlinks` extended attribute.

gramps_webapi/api/resources/base.py
```diff
@@ -63,6 +70,7 @@ class GrampsObjectResource(GrampsObjectResourceHelper, Resource):
             ),
             "profile": fields.Str(validate=validate.Length(equal=0)),
             "extend": fields.DelimitedList(fields.Str(validate=validate.Length(min=1))),
+            "backlinks": fields.Boolean(missing=False),
```
cdhorn, line 74: Can we treat this in same manner as I did the other booleans like profile just above where absence of the string implies false and presence in the argument list implies true?

cdhorn (commented):
No apischema.yaml update to keep it in sync?  I can submit something separate for that after you merge it if you like.

DavidMStraub:
> No apischema.yaml update to keep it in sync? I can submit something separate for that after you merge it if you like.

Good point, I will do it.

gramps_webapi/api/resources/base.py
DavidMStraub: It's true that it doesn't make sense to be inconsistent, so we should use the same everywhere.

The reason I used the built-in boolean field type was that I was feeling uneasy about treating an empty string as "true". I even found that there is a [discussion](https://github.com/marshmallow-code/marshmallow/issues/713) to treat an empty string as "none". I mean, I don't claim that the way we do it now will break since we just check if the value is present in the query string. But it somehow doesn't feel entirely clean.

For instance, by forcing the "user" (or rather, apps) to use `something=1` and `something=0`, (or `true`/`false`), it would also be possible to change the default from false to true at some point. It somehow seems cleaner.

cdhorn: That is a good point about the default that I had not thought of. That makes good sense to me, I agree. I will submit something to switch the ones I had changed back to booleans.

DavidMStraub:
@cdhorn, I was getting second thoughts about this. Doing the profiling, I saw that fetching the entire people list without or with backlinks only increases the time from 561 to 631 ms on my machine, i.e. by 10%. Only when using `extend=backlinks` is the time significantly increased (which is understandable). So maybe it was not such a good idea introducing a query parameter for this. My original thinking was that this was a slow operation, but since we realized it is just fetching from a fully indexed DB table, this is not the case.

So now I think it would be better to just treat `backlinks` like any other key and show it by default. It can still be switched off by setting `keys` (at the moment we waste time by discarding keys after having calculated them, but this is just a performance question that can be improved later, not an API issue).

Do you agree?

cdhorn:
I'm honestly not sure what is best because I am unsure of the size of your tree, how many backlinks different objects have on average, and how often backlinks are used or needed in an application.

I just ran a fresh import of my Ancestry.com tree.  There are 19452 people, 7196 families, 64366 events, 160323 citations, 1463 sources, 10 measly repositories, 16654 media objects, and 330349 notes.  There is at least one note for every citation.  If I import all the screenshots for citations the number of media objects increases accordingly. I suspect the performance profile for this may be a little different from your tree.  

Regarding keys/skipkeys/strip, because of the OOP design and data model the object has to be instantiated to access the attributes and then keys/skipkeys/strip filter the data when the object is serialized for output. Can you think of a more efficient way of doing that with the current Gramps design?

DavidMStraub:
> Regarding keys/skipkeys/strip, because of the OOP design and data model the object has to be instantiated to access the attributes and then keys/skipkeys/strip filter the data when the object is serialized for output. Can you think of a more efficient way of doing that with the current Gramps design?

You are right, there is nothing to optimize for the keys that come from serializing the object.

It would be interesting to see the performance on your big database, but maybe the original implementation with the `backlinks` key was better after all, as it gives most flexibility.

cdhorn:
I've not much experience with code profiling. I see you use snakeviz for helping analyze the data. What library do you recommend for the profiling itself?  cProfile?

Nick-Hall:
We have a small [profiling](https://gramps-project.org/wiki/index.php/Debugging_Gramps#Use_profiling) section in the wiki.

DavidMStraub:
I use Jupyter notebooks with %timeit and %snakeviz (like the one linked above) and am happy with it so far.

DavidMStraub:
Added the missing API docs. Ready to merge from my side.

cdhorn:
Sounds good!

PR #57, opened 2020-11-05
Refactor GRAMPSHOME to fix singletons

cdhorn:
@DavidMStraub I started on filter endpoint tests and immediately ran into an issue as it found the `custom_filters.xml` in my home `.gramps` directory. 

So I ended up having to refactor things again and set `GRAMPSHOME` in the top level `__init__.py` before any Gramps imports just as you had found when experimenting with `os.environ["LANGUAGE"] = "en"` before.  I think it is good actually, code is cleaner now. 

I would advise merging this right away.  I included the little I have for `test_filter.py` simply because it will fail if it finds a `custom_filters.xml` with filters defined, as it expects there to be none when it starts running, so it is good to have to confirm the test fixture setup should be as expected.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -6,8 +6,6 @@ import shutil
 import tempfile
 import unittest
 
-global TEST_GRAMPSHOME
-
 TEST_GRAMPSHOME = tempfile.mkdtemp()
 os.environ["GRAMPSHOME"] = TEST_GRAMPSHOME
 
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,17 +1,21 @@
 """Unit test package for gramps_webapi."""
 
 import gzip
-import importlib
 import os
 import shutil
 import tempfile
 import unittest
-from typing import Optional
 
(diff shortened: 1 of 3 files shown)


DavidMStraub:
Yes, makes sense! But can you please remove the `global` keyword on all top level declarations, i.e. outside functions/classes and see if it still works? I'm pretty sure it has no effect.

cdhorn:
Oh yeah for GRAMPSHOME not needed, done. It is needed in test_endpoints/__init__.py because of scoping, both of those are set inside setupUpModule. Or at least I'm not aware of another way to handle that, is there one?

DavidMStraub:
It should be enough to have `global TEST_CLIENT, TEST_OBJECT_COUNTS` in `setUpModule`, the `global TEST_CLIENT, TEST_OBJECT_COUNTS` at the top level doesn't have any effect I believe. Or does something break when you remove it?

cdhorn:
Oh I didn't realize that was unnecessary, yes works fine.

Issue #15, opened 2020-09-29
Group-based user permissions


DavidMStraub:
Possibility to assign users to groups and to assign permissions to groups. First feature to be implemented: ability to read private records.

PR #59, opened 2020-11-05
Add XML endpoint

DavidMStraub:
This adds a simple `/api/xml` endpoint that returns the entire tree as a Gramps XML file download. With the option `?compress=1`, it returns the download in gzip format.

This can be used by apps to export the tree.

This can be improved in the future with the use of filters and by restricting private records, e.g. in combination with users groups (#15), but this PR is a start.

### DIFF
--- a/gramps_webapi/api/__init__.py
+++ b/gramps_webapi/api/__init__.py
@@ -22,6 +22,7 @@ from .resources.source import SourceResource, SourcesResource
 from .resources.tag import TagResource, TagsResource
 from .resources.token import TokenRefreshResource, TokenResource
 from .resources.translate import TranslationResource, TranslationsResource
+from .resources.xml import export_xml
 
 api_blueprint = Blueprint("api", __name__, url_prefix=API_PREFIX)
 
@@ -74,9 +75,10 @@ register_endpt(TranslationResource, "/translations/<string:isocode>", "translati
 register_endpt(TranslationsResource, "/translations/", "translations")
 # Relation
 register_endpt(
-    RelationResource,
-    "/relations/<string:handle1>/<string:handle2>",
-    "relations",
+    RelationResource, "/relations/<string:handle1>/<string:handle2>", "relations",
 )
 # Metadata
 register_endpt(MetadataResource, "/metadata/<string:datatype>", "metadata")
+
+# XML export
+api_blueprint.route("/xml")(export_xml)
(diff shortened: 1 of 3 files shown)


DavidMStraub:
The tests fail since `gramps.plugins.export.exportxml` seems to require `gi`, which is surprising. I believe this is due to this unused import:

https://github.com/gramps-project/gramps/blob/c567b9e399514cfbf3bc6e5dd3996a04b7f2ed97/gramps/plugins/export/exportxml.py#L65

```python
from gramps.gui.plug.export import WriterOptionBox, WriterOptionBoxWithCompression
```
I believe this can be removed. Thoughts, @Nick-Hall?

cdhorn:
Why don't we make the endpoint /api/export/xml or /api/export/gramps? May want to provide a /api/export/gedcom in the future, and who knows how things will evolve over time.

DavidMStraub:
Good idea, done!

Nick-Hall:
> The tests fail since `gramps.plugins.export.exportxml` seems to require `gi`, which is surprising.

Our [Programming_guidelines](https://gramps-project.org/wiki/index.php/Programming_guidelines#Imports) only state that the `gen` module should be self-contained.  However, it would obviously be useful if some plugin types such as importers, exporters and reports did  not reference gui code.

I'm not sure if you can simply remove the import statement you refer to.  I seem to remember someone previously having problems with this.

Nick-Hall:
> May want to provide a /api/export/gedcom in the future

Why don't we create an endpoint `/api/exporters` to list all the exporters?

Then you could have `/api/exporters/{plugin id}/options` to list the available options, and `/api/exporters/{plugin id}/options/{option id}` to give details of an option.

Something similar would also work for reports and tools.  The [Command line](https://gramps-project.org/wiki/index.php/Gramps_5.1_Wiki_Manual_-_Command_Line) section of the manual may be worth reading when you design this.

DavidMStraub:
> I'm not sure if you can simply remove the import statement you refer to. I seem to remember someone previously having problems with this.

Oh I think I found something, in a different plugin (`exportgedcom.py`):

```python
# keep the following line even though not obviously used (works on import)
from gramps.gui.plug.export import WriterOptionBox
```

So it seems to be doing some magic ...

> Why don't we create an endpoint `/api/exporters` to list all the exporters?
> 
> Then you could have `/api/exporters/{plugin id}/options` to list the available options, and `/api/exporters/{plugin id}/options/{option id}` to give details of an option.

Good ideas! Will look into it.

cdhorn:
Yes, I like that suggestion!  So seems we want /api/reports and /api/tools at some point too.

Are there other peices of potentially missing functionality like that we've overlooked so far?

DavidMStraub:
Closing this, see #77.

PR #60, opened 2020-11-06
Filter endpoint tests, unittest asserts, pylint cleanups (Closes #45)

### LINKED ISSUE #45: Unit tests for filters/{namespace} endpoint
Get, post, put, delete

cdhorn:
@DavidMStraub this actually touches things in five areas, which I know you won't be thrilled about, but I just struggled with doing things one way for `test_filters.py` and not fixing the other test cases I put together to be conformant.

1. Changes `/api/filters` endpoint so delete now follows the more standard convention of having the item in the path as well as tweaks the two query parms on a get and `apispec.yaml` update.

2. The remainder of the `/api/filters` tests in `test_filters.py`. Those cover create / update / apply / delete of a custom filter for each endpoint. 

3. It replaces _all_ of the plain `asserts` in the test_endpoints modules with the more specific unittest ones like `assertEqual`, `assertTrue`, and `assertIn` which I realized I really should have done from the start.

4. It renames `rv` to `result` in all the test_endpoints modules along with other changes to try to appease pylint as best I could. 

5. It includes two changes to `.pylintrc` that I had to make as it refused to run at all on my machine. The first removes the duplicate [MESSAGES CONTROL] section at the bottom and the second changes fail-under=10.0 to fail-under=10 as pylint wanted it to be an integer. I am unsure on this one, it clearly must be working on your system. I'm using 2.6.0.

Sorry again for the size. I know you are working on test cases too and may be some collision there. So if you're stuff is ready to merge maybe do that first and I can rebase and revise this.  Let me know your thoughts.

### DIFF
--- a/gramps_webapi/api/__init__.py
+++ b/gramps_webapi/api/__init__.py
@@ -14,7 +14,7 @@ from .resources.bookmark import BookmarkResource, BookmarksResource
 from .resources.citation import CitationResource, CitationsResource
 from .resources.event import EventResource, EventsResource
 from .resources.family import FamiliesResource, FamilyResource
-from .resources.filters import FilterResource
+from .resources.filters import FilterResource, FiltersResource
 from .resources.media import MediaObjectResource, MediaObjectsResource
 from .resources.metadata import MetadataResource
 from .resources.name_groups import NameGroupsResource
@@ -81,7 +81,7 @@ register_endpt(BookmarkResource, "/bookmarks/<string:namespace>", "bookmark")
 register_endpt(BookmarksResource, "/bookmarks/", "bookmarks")
 # Filter
 register_endpt(FilterResource, "/filters/<string:namespace>/<string:name>", "filter")
-register_endpt(FilterResource, "/filters/<string:namespace>", "filters")
+register_endpt(FiltersResource, "/filters/<string:namespace>", "filters")
 # Translate
 register_endpt(TranslationResource, "/translations/<string:isocode>", "translation")
 register_endpt(TranslationsResource, "/translations/", "translations")
--- a/gramps_webapi/api/__init__.py
+++ b/gramps_webapi/api/__init__.py
@@ -80,7 +80,8 @@ register_endpt(NameGroupsResource, "/name-groups/", "name-groups")
 register_endpt(BookmarkResource, "/bookmarks/<string:namespace>", "bookmark")
 register_endpt(BookmarksResource, "/bookmarks/", "bookmarks")
 # Filter
-register_endpt(FilterResource, "/filters/<string:namespace>", "filter")
+register_endpt(FilterResource, "/filters/<string:namespace>/<string:name>", "filter")
+register_endpt(FilterResource, "/filters/<string:namespace>", "filters")
 # Translate
 register_endpt(TranslationResource, "/translations/<string:isocode>", "translation")
 register_endpt(TranslationsResource, "/translations/", "translations")
--- a/gramps_webapi/api/resources/filters.py
+++ b/gramps_webapi/api/resources/filters.py
@@ -7,8 +7,8 @@ import gramps.gen.filters as filters
 from flask import Response, abort
 from gramps.gen.db.base import DbReadBase
 from gramps.gen.filters import GenericFilter
-from marshmallow import Schema, ValidationError, fields
-from webargs import validate
+from marshmallow import Schema
+from webargs import ValidationError, fields, validate
 from webargs.flaskparser import use_args
 
 from ...const import GRAMPS_NAMESPACES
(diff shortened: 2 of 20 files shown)


DavidMStraub:
> It replaces all of the plain asserts in the test_endpoints modules with the more specific unittest ones like assertEqual, assertTrue, and assertIn which I realized I really should have done from the start.

Well, `assert` [is the `pytest` way of doing things](https://docs.pytest.org/en/stable/assert.html), so as long as we use `pytest` I think it's fine both ways.

I know Gramps uses only `unittest`, but it actually still uses `python setup.py test`, which is [officially deprecated](https://github.com/pypa/setuptools/pull/1878).

Running plain `unittest` is much less useful/powerful for debugging compared to `pytest` in my view. And the alternatives to `pytest` like `nose` and `nose2` look pretty much abandoned.

But it's true that `pytest` has a quite opinionated way of writing tests, so maybe the best approach is to keep them `unittest` compatible and use `pytest` only as a test runner.

gramps_webapi/api/resources/filters.py
```diff
@@ -74,6 +78,8 @@ def get_custom_filters(args: Dict[str, str], namespace: str) -> List[Dict]:
                     ],
                 }
             )
+    if "filters" in args and len(args["filters"]) != len(filter_list):
+        abort(404)
```
DavidMStraub, line 82: Why 404 and not 400, isn't this an invalid request?

gramps_webapi/api/__init__.py
```diff
@@ -68,7 +68,8 @@ def register_endpt(resource: Type[Resource], url: str, name: str):
 register_endpt(BookmarkResource, "/bookmarks/<string:namespace>", "bookmark")
 register_endpt(BookmarksResource, "/bookmarks/", "bookmarks")
 # Filter
-register_endpt(FilterResource, "/filters/<string:namespace>", "filter")
+register_endpt(FilterResource, "/filters/<string:namespace>/<string:name>", "filter")
+register_endpt(FilterResource, "/filters/<string:namespace>", "filters")
```
DavidMStraub: So now you have multiple endpoints for the same resource class and the different verbs will have different number of arguments. Does that work? What if I send a `GET` request to `/filters/namespace`? Won't that cause a 500 server error? Don't you have do define `name` as a keyword argument (with default value) on all the verbs' methods? And return 405 if it's set on the wrong method?

gramps_webapi/api/resources/filters.py
cdhorn: My thinking is that it did not find all the filters in the list because one was missing, and we would provide 404 for a missing object.  I did the same for rules just above and do the same below in build filter.

DavidMStraub: Ok.

gramps_webapi/api/__init__.py
cdhorn: Oh yeah, you're right, I'm not handling that right. I will fix that up, thanks!

cdhorn:
> Well, assert is the pytest way of doing things, so as long as we use pytest I think it's fine both ways.

Yeah, this is my first real experience writing extensive test cases so I don't really know the pros and cons of each framework and approach. I'm hoping what I did for the test cases overall is okay, it made sense to me but you are far more experienced than I.

gramps_webapi/api/__init__.py
cdhorn: Rebased as well.  I think that is a bit better now. Boy I miss the obvious sometimes.

cdhorn:
If this looks okay now could you consider merging it so I can rebase the others?  Or merge one of the others so I can rebase what is left? Or do you want me to combine a couple of them?

jralls:
I am unalterably and intensely opposed to having anything even vaguely resembling a web API in GnuCash, and even more opposed to one written by the Guile developers. That is a security hole so huge that no responsible development team without the requisite time and skills could foist upon its users.

If you want to support such a thing then write it as a plugin and host it yourself.

cdhorn:
> I am unalterably and intensely opposed to having anything even vaguely resembling a web API in GnuCash, and even more opposed to one written by the Guile developers. That is a security hole so huge that no responsible development team without the requisite time and skills could foist upon its users.

Hi John, I am a little confused by this statement, what does GnuCash have to do with Gramps?  Did you mean to say you are opposed to a REST API for Gramps?

jralls:
I'm very sorry, I confused this PR with [a GnuCash one](https://github.com/Gnucash/gnucash/pull/769) also by someone named Christopher who has a habit of proclaiming his PRs ready to merge when they aren't. Please disregard the noise.

DavidMStraub:
Let's merge this before more opposition arises :wink:

Issue #56, opened 2020-11-04
Query parameter for locale?


DavidMStraub:
This would allow localizing date formats in the output. See the discussion in https://github.com/gramps-project/web-api/pull/54.

cdhorn:
Closed by #98

PR #63, opened 2020-11-08
Refactor types into separate endpoint. (Closes #62)

### LINKED ISSUE #62: Break types out into /api/types/default and /api/types/custom
@DavidMStraub initially we had a separate /api/types and then moved it to /api/metadata but I think I'd like to move it back to /api/types.  It just seems more "right", I don't know how to explain it. In doing so I also want to break it up so we have /api/types/default and /api/types/custom so under /api/types/default we can also expose the numeric values associated with them.

I'm looking over the relationship calculator and endpoint and I didn't really understand it at first and left most of the functionality out.  So I want to refactor that and develop the tests for it as I do so. There are some types in there for siblings and partners that I think maybe should be exposed. So if you are okay with this proposed change I will work on /api/types to get that sorted and then move onto the relationship endpoint work. Let me know your thoughts.

cdhorn:
@DavidMStraub this should be pretty self contained, it takes a basic pass at implementing /api/types/default and /api/types/custom endpoints. Some basic tests and an apispec.yaml update included.


DavidMStraub:
The code looks good, I am just a bit confused by the API: for all the other endpoints, the "root" endpoint returns all the information and the "leafs" only return a subset, right? So when calling `/api/types/default`, why am I not getting everything, like 

```
[
  {
    'id': 0,
    'type': 'attribute_types',
    'types:' [
      {
        'id': 0,
        'type': 'Unknown'
      },
      ...
    ], ...
  }, ...
]
```

Another thing I don't understand is the difference between `?locale=true` and `/map`. They both return the translated default types. The only 2 difference I can tell:

- For `?locale=true` 0 is "Unknown", while for `map`, 0 is the translation of "Custom"
- For `/map`, there is in addition -1, which is the translation of "Unknown".

To me it looks like `/map` can be removed, unless I'm missing something.

cdhorn:
> So when calling /api/types/default, why am I not getting everything

For the others there are calls that return everything, in this case there is no single call I noticed to get all the different attribute types in one shot.  Bookmarks works the same way at the moment.  

That is obviously inconsistent design wise though.  I can refactor it in a little while to do that.

In doing so do you think then `/api/types` itself should then return a dict with all default and custom types?
 
> the difference between ?locale=true and /map

Did I use locale here the way you envisioned?

As for maps, the maps are dictionaries and not lists.  The key to the map is the number used internally by Gramps, it seemed to make sense to expose it and not assume the number was the index to the item in the list. 

Note there are two things I didn't add in here as well that only appear to exist for `event_types`. One is the abbreviations, the other the menu. The menu is the object that groups the pre-defined event types by category like Religious, Vocation, Family, Legal, etc that you see when adding an event in the GUI. Those might be useful if someone wanted to leverage them in a web frontend. The menu provides the list of category members using their numeric representation, you would want the map to get the values. I figured could maybe add those somehow later on.  Do you think I should try to do that now too? Maybe /menu and /abbreviation endpoints but then they only ever work on default event_types?

DavidMStraub:
> In doing so do you think then /api/types itself should then return a dict with all default and custom types?

That sounds like a good idea.

> Did I use locale here the way you envisioned?

In the future (#56) I'd rather expect we have something like `?locale=de_DE`, but for the moment having it a boolean that uses the default locale is fine.

> As for maps, the maps are dictionaries and not lists. The key to the map is the number used internally by Gramps, it seemed to make sense to expose it and not assume the number was the index to the item in the list. There is a separate method to get the map for the Gramps type.

Funny, in the Firefox JSON preview, a list and a dictionary with numeric keys look *exactly* the same.

This is also why I hadn't noticed that the endpoints apart from `/map` are actually lists (my use uf `id` above was because I had erroneaously thought these were dicts).

I guess I don't fully understand how we would use these endpoints.

cdhorn:
> In the future (#56) I'd rather expect we have something like ?locale=de_DE

I thought that might be what you had in mind for that but was not sure.

> Funny, in the Firefox JSON preview, a list and a dictionary with numeric keys look exactly the same.

I know! That confused me at first too until I looked at the raw data.

> I guess I don't fully understand how we would use these endpoints.

The map and maybe menu and abbreviation ones?  I guess it depends on how a front end app is designed when it comes to create / update operations or if it might want to use abbreviations somewhere. If you needed abbreviations in your front end how would you go about generating them? Might they differ by locale?  Then the abbreviations provided by Gramps are useful.  If you needed the user to select an event type to create an event would you provide the one huge list or a set of categories and let them drill down like the Gramps GUI does? If the later then the menu provides that construct already and is useful. 

I guess it comes down to how much of Gramps functionality should the API try to expose. The more we can expose the more choices a consumer of the API has in how they go about things and the larger the existing code base they can leverage.

It is an amazing day here, more like late spring than fall, so I'll refactor this later tonight.

cdhorn:
Okay this changes things for consistency so /api/types returns all types and then as path extends continues to drill down to a specific one.

cdhorn:
Realized we don't need those and I did get things mostly right the /api/relations the first time now that working on tests for it so factored them out. This should be done now unless you see anything else that you think needs to be adjusted.

PR #64, opened 2020-11-09
Refactor /api/bookmarks for consistency with other endpoints (Closes #66)

### LINKED ISSUE #66: /api/bookmarks should return all bookmarks like other collections
As per discussion about types this should be changed for consistency.

cdhorn:
This one is self contained and refactors /api/bookmarks to return all of them and then drill down to specific set to be consistent with the other endpoints based on discussion about PR #63

### DIFF
--- a/gramps_webapi/api/resources/bookmark.py
+++ b/gramps_webapi/api/resources/bookmark.py
@@ -1,11 +1,51 @@
 """Bookmark API resource."""
 
+from typing import List, Optional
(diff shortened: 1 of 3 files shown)


DavidMStraub:
Looks good! But after merging the other PRs it looks like there is a small conflict in a unit test.

cdhorn:
Rebased now, this should be ready.

PR #65, opened 2020-11-09
Refactor /api/relations and add unit tests (Closes #47)

### LINKED ISSUE #47: Unit tests for relations/{handle1}/{handle2} endpoint
Not sure this needed as filters may provide the functionality, if so may remove it.

cdhorn:
This removes ?all parm and adds /all in path as response formats differed which was not clean.  Includes unit tests and apispec.yaml update.

Now wondering if I should have just added this to PR #60

### DIFF
--- a/gramps_webapi/api/__init__.py
+++ b/gramps_webapi/api/__init__.py
@@ -21,7 +21,7 @@ from .resources.name_groups import NameGroupsResource
 from .resources.note import NoteResource, NotesResource
 from .resources.person import PeopleResource, PersonResource
 from .resources.place import PlaceResource, PlacesResource
-from .resources.relation import RelationResource
+from .resources.relation import RelationResource, RelationsResource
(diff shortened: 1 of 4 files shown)


DavidMStraub:
Looks good!

Concerning `locale`, I guess this is can later be promoted to use a `locale` parameter on the base class's `get`, right?

Could you add a unit test for `locale`, too? This would ease refactoring as part of #56.

cdhorn:
> Concerning locale, I guess this is can later be promoted to use a locale parameter on the base class's get, right?

Probably, but this might need to get refactored again at some point, I am unsure.  I did not realize before but if I recall there appear to be plugin relationship calculators for different languages. I'm not sure how this and they are related.

> Could you add a unit test for locale, too?

I will look at it.

cdhorn:
Rebased. Unsure if locale test I added should do anything more specific than it is.

DavidMStraub:
Perfect.

PR #72, opened 2020-11-13
Add /api/reports endpoint (Closes #67)

### LINKED ISSUE #67: Add /api/reports endpoint
Expose the reports plugin API.

cdhorn:
@DavidMStraub @Nick-Hall this is a work in progress that takes a first shot at adding the `/api/reports` endpoint. 

`/api/reports` returns the list of available reports and all their attributes and options, `/api/reports/{report}` does so for a specific report, and then `/api/reports/{report}/file` generates one.  

The web ones are filtered out and not returned at this time. I suppose one could write code to tar or zip up the directory and pass that down to the client but not sure it is really needed with a Web front end like David is building.

The report options are passed in using `options={}` query parm.  All options are passed in as strings and converted later.  So an example would be like this:

`
http://127.0.0.1:5000/api/reports/det_descendant_report/file?options={"gen":"5","incphotos":"True","off":"rtf"}`

I added some code to try to validate the options passed in against the available report options to verify the request is valid. I also set it up so it defaults to `pdf` if the `off` option is not provided. The file is generated temporarily on disk, read into the buffer to send to the client, and then deleted.

If you guys have a chance and can look it over and let me know if you see any problems, things that could be done a better way, etc as well as additional functionality I may be missing that you think is needed I'd appreciate it. I'll work on apischema updates and unit tests for it next. Thanks!

### DIFF
--- a/gramps_webapi/api/__init__.py
+++ b/gramps_webapi/api/__init__.py
@@ -48,7 +48,7 @@ from .resources.notes import NoteResource, NotesResource
 from .resources.people import PeopleResource, PersonResource
 from .resources.places import PlaceResource, PlacesResource
 from .resources.relations import RelationResource, RelationsResource
-from .resources.reports import ReportResource, ReportRunnerResource, ReportsResource
+from .resources.reports import ReportFileResource, ReportResource, ReportsResource
 from .resources.repositories import RepositoriesResource, RepositoryResource
 from .resources.sources import SourceResource, SourcesResource
 from .resources.tags import TagResource, TagsResource
@@ -144,7 +144,7 @@ register_endpt(
     "relations",
 )
 # Reports
-register_endpt(ReportRunnerResource, "/reports/<string:report_id>/file", "run-report")
+register_endpt(ReportFileResource, "/reports/<string:report_id>/file", "report-file")
 register_endpt(ReportResource, "/reports/<string:report_id>", "report")
 register_endpt(ReportsResource, "/reports/", "reports")
 # Exporters
--- a/gramps_webapi/api/__init__.py
+++ b/gramps_webapi/api/__init__.py
@@ -144,8 +144,8 @@ register_endpt(
     "relations",
 )
 # Reports
-register_endpt(ReportRunnerResource, "/reports/<string:id>/file", "run-report")
-register_endpt(ReportResource, "/reports/<string:id>", "report")
+register_endpt(ReportRunnerResource, "/reports/<string:report_id>/file", "run-report")
+register_endpt(ReportResource, "/reports/<string:report_id>", "report")
 register_endpt(ReportsResource, "/reports/", "reports")
 # Exporters
 register_endpt(
--- a/gramps_webapi/api/__init__.py
+++ b/gramps_webapi/api/__init__.py
@@ -48,6 +48,7 @@ from .resources.notes import NoteResource, NotesResource
 from .resources.people import PeopleResource, PersonResource
 from .resources.places import PlaceResource, PlacesResource
 from .resources.relations import RelationResource, RelationsResource
+from .resources.reports import ReportResource, ReportRunnerResource, ReportsResource
 from .resources.repositories import RepositoriesResource, RepositoryResource
 from .resources.sources import SourceResource, SourcesResource
 from .resources.tags import TagResource, TagsResource
@@ -142,6 +143,10 @@ register_endpt(
     "/relations/<string:handle1>/<string:handle2>/all",
     "relations",
 )
+# Reports
+register_endpt(ReportsResource, "/reports/", "reports")
+register_endpt(ReportResource, "/reports/<string:id>", "report")
+register_endpt(ReportRunnerResource, "/reports/<string:id>/file", "run-report")
 # Exporters
 register_endpt(
     ExporterFileResource, "/exporters/<string:extension>/file", "exporter-file"
--- a/gramps_webapi/api/__init__.py
+++ b/gramps_webapi/api/__init__.py
@@ -144,9 +144,9 @@ register_endpt(
     "relations",
 )
 # Reports
-register_endpt(ReportsResource, "/reports/", "reports")
-register_endpt(ReportResource, "/reports/<string:id>", "report")
 register_endpt(ReportRunnerResource, "/reports/<string:id>/file", "run-report")
+register_endpt(ReportResource, "/reports/<string:id>", "report")
+register_endpt(ReportsResource, "/reports/", "reports")
 # Exporters
 register_endpt(
     ExporterFileResource, "/exporters/<string:extension>/file", "exporter-file"
--- a/gramps_webapi/api/resources/reports.py
+++ b/gramps_webapi/api/resources/reports.py
@@ -1,0 +1,214 @@
+"""Reports Plugin API resource."""
+
+import io
+import json
+import os
+import uuid
+from pathlib import Path
+from typing import BinaryIO, Dict
+
+from flask import Response, abort, send_file
+from gramps.cli.plug import CommandLineReport, cl_report
+from gramps.gen.db.base import DbReadBase
+from gramps.gen.filters import reload_custom_filters
+from gramps.gen.plug import CATEGORY_BOOK, CATEGORY_WEB, BasePluginManager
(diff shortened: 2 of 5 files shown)


gramps_webapi/api/__init__.py
```diff
@@ -86,10 +86,16 @@ def register_endpt(resource: Type[Resource], url: str, name: str):
...
 # Relation
 register_endpt(
-    RelationResource, "/relations/<string:handle1>/<string:handle2>", "relations",
+    RelationResource,
+    "/relations/<string:handle1>/<string:handle2>",
+    "relations",
 )
 # Metadata
-register_endpt(MetadataResource, "/metadata/<string:datatype>", "metadata")
+register_endpt(MetadataResource, "/metadata/", "metadata")
```
DavidMStraub: Why was this changed here, given it is unrelated to reports?

gramps_webapi/api/resources/reports.py
```diff
@@ -0,0 +1,214 @@
...
+        "icondir": report_data.icondir,
+        "category": report_data.category,
+        "module": report_data.mod_name,
+        "reportclass": report_data.reportclass,
+        "require_active": report_data.require_active,
+        "report_modes": report_data.report_modes,
+        "option_class": report_data.optionclass,
+        "options_dict": report.options_dict,
+        "options_help": report.options_help,
+    }
```
DavidMStraub, line 76: Stupid question, but this dict looks like it is always `"key": report_data.key`, couldn't that be abbreviated somehow? Perhaps using `report_data.__dict__`?

gramps_webapi/const.py
```diff
@@ -56,3 +65,46 @@
...
+    "rtf": MIME_RTF,
+    "dot": MIME_DOT,
+    "gspdf": MIME_GSPDF,
+    "gvpdf": MIME_GVPDF,
+    "svg": MIME_SVG,
+    "svgz": MIME_SVGZ,
+    "jpg": MIME_JPEG,
+    "gif": MIME_GIF,
+    "png": MIME_PNG,
+}
```
DavidMStraub: This looks to me like it is a subset of `mimetypes.types_map` (which is part of the standard library), is there a good reason for us to not use that?

DavidMStraub:
I think it's very good! The API makes perfect sense to me.

One thing I haven't fully understood yet is whether we need PyGObject as a dependency to run any of the reports. That's not a big problem, but we then have to decide how to deal with it - perhaps leave it as an optional dependency and disable the report and export (#59) endpoints when it's missing. I guess we will find out once you add unit tests and run them on Travis.

gramps_webapi/api/resources/reports.py
cdhorn: Not stupid, I'll have to look at it again.  Possible and I just didn't think of it.

gramps_webapi/api/__init__.py
cdhorn: I probably mixed something up, it'll get fixed by time this is ready for merging.  I have some code laying around that refactored the metadata endpoint.

gramps_webapi/const.py
cdhorn: Oh I was not aware of that!  Yes, I will certainly use that then.

cdhorn:
Might be a stupid question on my part but wouldn't that already be a dependency for the Gramps package itself?

DavidMStraub:
It's a depedency for the desktop app, but not necessarily for the Python package. In the current version of `setup.py`, where I added dependencies for use with `pip`, it is treated as an extra dependency:
https://github.com/gramps-project/gramps/blob/master/setup.py#L418

The issue with PyGObject is that one cannot simply `pip install` it (it has binary dependencies), so by promoting it to an install requirement, one loses direct `pip` installability for the dependent package, which would be a pity.

Honestly I don't really understand why some of the reports and exporters depend on GTK even if they can be invoked from the command line (without GUI), but I guess Gramps was never really intended to be run without a GUI.

Nick-Hall:
Gramps was designed so that all reports can be run from the command line.  The idea is that the `cli` and `gen` modules should not have any `gui` dependencies.  However, the may be a few places in the core code where they still remain.  This project provides a good opportunity to find and fix them.

gramps_webapi/api/resources/reports.py
cdhorn: Unfortunately we only want report specific attributes in the response here and that approach would include attributes associated with other plugin types.

cdhorn:
Well I guess that seems to answer the question about Gnome dependencies.

cdhorn:
@Nick-Hall @DavidMStraub  Looking at this all morning, and if I understand things properly, the issue seems to me to be that the CLI report dependencies get dragged in depending on the docgen plugins in use.  In the case of Postscript and PDF reports the cairodoc.py module that renders Postscript and PDF reports. So some output formats may work without PyGObject and other things will not. I suppose other docgen backends may also have dependencies that might make them problematic in a similar manner as well. 

> perhaps leave it as an optional dependency and disable the report and export (#59) endpoints when it's missing.

If my analysis is correct, I suspect the export plugins will work fine.

But for reports I'm unsure what to do.

I suppose I could try to figure out which output formats may or may not work and then selectively disable them, filtering them out of the get results and throwing an error if they can not be run.  So then the endpoint would always be there and the exposed functionality would vary based on environment.

I am not sure how unit tests should be handled then.  I suppose build them around output formats that should always work?

Thoughts on what best to do here?

cdhorn:
Doing some testing on a bare Ubuntu VM configured the same as the Travis setup, RTF reports work just fine.  So I am going to probe and adjust the exposed formats based on what is available in the environment.

DavidMStraub:
Sounds good! If most reports work without PyGObject, I suggest to not install it on Travis and only add unit tests for the ones that do, for the time being.

cdhorn:
The rtfdoc and odfdoc plugins use gramps.gen.utils.image which uses GdxPixbuf.

cdhorn:
Ah I put fix for the test this failed on in the metadata PR.

Nick-Hall:
> In the case of Postscript and PDF reports the cairodoc.py module that renders Postscript and PDF reports.

Yes.  I'd forgotten that we use the Cairo library to render pdf.

DavidMStraub:
@Nick-Hall, concerning

> Gramps was designed so that all reports can be run from the command line. The idea is that the `cli` and `gen` modules should not have any `gui` dependencies. However, the may be a few places in the core code where they still remain. This project provides a good opportunity to find and fix them.

in #82 @cdhorn now added PyGObject as a dependency to be able to export, but for many exporters the only use is the import of `gramps.gui.plug.export`, which has no effect (as far as I can tell) when exporting from the command line, but cannot be removed since it executes some code which is needed for GUI export. (See e.g. this commit https://github.com/gramps-project/gramps/commit/4ae5c6477bf2f14ba8fa3910a3d67889d7fdfaed).

An easy way to fix this - i.e. make important exporters like Gedcom and Gramps XML work without GTK - would be to replace
```python
from gramps.gui.plug.export import WriterOptionBox, WriterOptionBoxWithCompression
```
by
```python
try:
    from gramps.gui.plug.export import WriterOptionBox, WriterOptionBoxWithCompression
except ImportError:
    pass
```
in the respective files.

What do you think about this?

Nick-Hall:
It would be nice use the same solution as the reports code.  So we could have a command line options class and also a gui version.  However, I'm not sure if such a change would be easy.

cdhorn:
@DavidMStraub I did not want to add Graphviz binaries to dependencies for the tests as I do not know what else that will pull in or may need so this checks and disables the reports that rely on it if it is not available. I think this ready now, not sure if you want to look it over again.

cdhorn:
> It would be nice use the same solution as the reports code. So we could have a command line options class and also a gui version. However, I'm not sure if such a change would be easy.

I will try to take a look at that once a few more things are wrapped up with this API and see if I can figure it out.

I seems to me we may want to change some other things in Gramps core to facilitate things this API could be used for.

For example the custom filters are global across databases.  I think ideally a user might want global ones, but may also want some tied to a specific database if they reference objects in that database.  And then if web based applications were to support multiple users you might want to break them up by user.

I kind of wondered why they are not just stored in the database to start with, but I'm sure there are good reasons.

Issue #73, opened 2020-11-14
Get default locale


DavidMStraub:
The `/translations/somelocale` endpoint is really useful to translate e.g. a web app frontend, however I realized that it is not possible at present to get the default language/locale of the Gramps instance, only the list of available languages. Since currently the default locale is also used in the translation of dates (see #56), I think it would make sense to add the possibility to get the default language and to use this for translating strings.

However, I am not sure what the best location would be for this information. Perhaps changing the `/api/translations` endpoint to

```
{'available': [...], 'default': {'isocode': 'en', ...}}
```

cdhorn:
A couple thoughts...

I have an updated metadata I was playing with and actually put it in there.  But it could be put under here. 

A stupid question probably but is that something that can be changed after Gramps is instantiated?  Maybe /api/locale endpoint to get and set it?

Also I had not realized/known when I last touched this that in the apispec additionalProperties is used to model dictionaries. The returned list of translations was initially a dictionary and I changed it but think it is more natural to change it back. I can submit a PR to do that if you agree, that was a mistake on my part.

DavidMStraub:
> I have an updated metadata I was playing with and actually put it in there. But it could be put under here.

Metadata would be perfectly fine.

> A stupid question probably but is that something that can be changed after Gramps is instantiated? Maybe /api/locale endpoint to get and set it?

I wouldn't recommend it as this will affect all users. I think it's better to keep it as part of the server configuration and to use the `/translation/...` and `?locale=...` options for user-dependent language config.

> Also I had not realized/known when I last touched this that in the apispec additionalProperties is used to model dictionaries. The returned list of translations was initially a dictionary and I changed it but think it is more natural to change it back. I can submit a PR to do that if you agree, that was a mistake on my part.

Yes, sounds good.

cdhorn:
Thinking about it, locale really doesn't have to do with Gramps or the database, so I am thinking a /api/locale may make sense from that point of view. Thoughts?  If okay I'll submit a PR for that and fixup translations in the same.

Nick-Hall:
You can't change the default locale, but it is possible to run reports in different languages.  We use deferred translation to implement this.

cdhorn:
Question too about this.  As I understand is "lang" is the ISO code.  I referred to it as isocode in translations endpoint. Is it more proper to refer to it that way or as lang?

cdhorn:
/api/metadata now contains locale/lang as per PR #76

PR #76, opened 2020-11-14
Refactor /api/metadata and add tests (Closes #48 and #73)

### LINKED ISSUE #48: Unit tests for metadata/summary, metadata/researcher, metadata/surnames

cdhorn:
This refactors /api/metadata into a single get against the base endpoint and it returns everything. It is quick enough it doesn't seem to make sense to break things up into a bunch of smaller peices although that is probably the more RESTful thing to do.

The Gramps config settings and ENV settings are exposed in here now, as is locale.

Because I read in the schema to get the schema version for this I considered exposing that, but not sure if that would be here or /api/schema.  Do you think it should be somewhere or no?

Let me know if you would like anything added / removed / changed.

### DIFF
--- a/gramps_webapi/api/__init__.py
+++ b/gramps_webapi/api/__init__.py
@@ -113,7 +113,7 @@ register_endpt(
     "relations",
 )
 # Metadata
-register_endpt(MetadataResource, "/metadata/<string:datatype>", "metadata")
+register_endpt(MetadataResource, "/metadata/", "metadata")
 
 # Media files
 @api_blueprint.route("/media/<string:handle>/file")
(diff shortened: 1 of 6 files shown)


DavidMStraub:
Yes, makes sense! Also from an app point of view, I think one would like to fetch all the info at once, e.g. to display on a status page.

Concerning the content, I tend to think `mediapath` should be removed, given a server might handle media differently (in a different path, in the cloud, ...).

In general, I think the values containing absolute paths should not be accessible to every user because of possible security implications (the file structure of the server should be of no concern to the user). At some point we're going to have user-based permissions (#15) but in this case I wonder whether it wouldn't be easier to just omit them - they are really only relevant for the server admin, who doesn't need the API to find the path.

cdhorn:
Yeah I see the security view point. I will revisit this.

cdhorn:
@DavidMStraub rebased, should be good to go now.

DavidMStraub (approved):

PR #78, opened 2020-11-15
Use localized string for event type in profile

DavidMStraub:
A one-line PR: while using the un-localized event type names in `/events`, using `event.type.xml_str()` is the correct thing to do (as we discussed [here]), I realized it makes less sense for `profile`: here the date string is anyway localized and the whole point of the profile is to allow apps to display lists without having to do a lot of transformation. So I think it would be better to have the event type localized in the event profile dict. This is achieved by simply using `str(event.type)` instead.

### DIFF
--- a/gramps_webapi/api/resources/util.py
+++ b/gramps_webapi/api/resources/util.py
@@ -81,7 +81,7 @@ def get_sex_profile(person: Person) -> str:
 def get_event_profile_for_object(db_handle: DbReadBase, event: Event) -> Dict:
     """Get event profile given an Event."""
     return {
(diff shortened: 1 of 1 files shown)


cdhorn:
Could use locale as a boolean as I am elsewhere to toggle the behaviour...

DavidMStraub:
But if `locale` is going to turn into a string (#56), I would expect that without a string, it uses the default locale (for things that are supposed to be localized), no?

cdhorn:
Yeah I suppose you are right.

DavidMStraub:
Anyway we can revisit this as part of #56.

PR #80, opened 2020-11-16
WIP: Implement full-text search engine (fixes #74)

### LINKED ISSUE #74: Full text search
From the app point of view, a really great feature would be a Google-style full text search where I can just search for an arbitrary word and I get a list of all person, place, note, ..., objects that contain it. The API to implement that would be quite obvious,

```
/api/search/?query=something
```

and it would return a dictionary with lists of objects,
```
{
  'people':  [...],
  'notes': [...],
}
```

In order not to have to read the entire database on every search query, such a feature would require an index. Fortunately, there are excellent Python-only indexers, e.g. [whoosh](https://whoosh.readthedocs.io/en/latest/index.html). Building the index might take a couple of seconds, but updating it would be very fast as we can leverage the `change` timestamp field in the database.

If this works and turns out to be useful, it might even be something that finds its way into the Gramps desktop app ...

DavidMStraub:
As discussed in #74, I added a search indexer using `whoosh` combined with Gramps's `get_text_data_list` and `get_text_data_child_list` methods. It iterates over all objects in the DB and indexes a text string that is the concatenation of all text data and the text data of children (in the case of person objects, those are names and attributes, for instance). It works really well and is very fast! You can look at the unit tests to see how it works.

To do list:

- [x] Add endpoint 
- [x] Add paging to endpoint
- [x] Implement `extend` (which will replace the handle list by an object list)
- ~Implement incremental reindexing (checking the last modified date of the index and the DB and then fetching only objects that have changed)~
- [x] Trigger the full indexing (at application startup if it doesn't exist?)
- ~Trigger the incremental reindexing (before every query?)~
- [x] Add config option for index dir

### DIFF
--- a/gramps_webapi/api/__init__.py
+++ b/gramps_webapi/api/__init__.py
@@ -50,6 +50,7 @@ from .resources.places import PlaceResource, PlacesResource
 from .resources.relations import RelationResource, RelationsResource
 from .resources.reports import ReportFileResource, ReportResource, ReportsResource
 from .resources.repositories import RepositoriesResource, RepositoryResource
+from .resources.search import SearchResource
 from .resources.sources import SourceResource, SourcesResource
 from .resources.tags import TagResource, TagsResource
 from .resources.token import TokenRefreshResource, TokenResource
@@ -137,14 +138,10 @@ register_endpt(TranslationResource, "/translations/<string:language>", "translat
 register_endpt(TranslationsResource, "/translations/", "translations")
 # Relations
 register_endpt(
-    RelationResource,
-    "/relations/<string:handle1>/<string:handle2>",
-    "relation",
+    RelationResource, "/relations/<string:handle1>/<string:handle2>", "relation",
 )
 register_endpt(
-    RelationsResource,
-    "/relations/<string:handle1>/<string:handle2>/all",
-    "relations",
+    RelationsResource, "/relations/<string:handle1>/<string:handle2>/all", "relations",
 )
 # Reports
 register_endpt(ReportFileResource, "/reports/<string:report_id>/file", "report-file")
@@ -160,6 +157,8 @@ register_endpt(ExportersResource, "/exporters/", "exporters")
 register_endpt(MetadataResource, "/metadata/", "metadata")
 # User
 register_endpt(UserChangePasswordResource, "/user/password/change", "change_password")
+# Search
+register_endpt(SearchResource, "/search/", "search")
 
 # Media files
 @api_blueprint.route("/media/<string:handle>/file")
--- a/gramps_webapi/api/resources/search.py
+++ b/gramps_webapi/api/resources/search.py
@@ -10,10 +10,16 @@ from . import ProtectedResource
 class SearchResource(ProtectedResource):
     """Fulltext search resource."""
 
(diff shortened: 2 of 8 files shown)


Nick-Hall:
Interesting.  We should certainly consider using this in Gramps core.

DavidMStraub:
- Added automatic full reindex at app startup. This is pretty fast so I think it is acceptable.
```
[2020-11-27 21:00:26,549] INFO in app: Building search index ...
[2020-11-27 21:00:30,968] INFO in app: Done building search index.
 * Serving Flask app "gramps_webapi.app" (lazy loading)
 ```
(This is on my laptop with example.gramps, which is not that small a DB.)
- De-scoped the incremental indexing. I think this is less important as long as we're read-only, and the write operations can be used as a trigger, once implemented.
- Added a trivial endpoint so far. I was a bit torn what URL and which verb to use, but `GET` `/api/search?q=query` seemed the most natural to me in the end.

**Comment by cdhorn** [comment]
That makes sense to me, but just curious why q= and not query=?

**Comment by DavidMStraub** [comment]
> That makes sense to me, but just curious why q= and not query=?

Good point, changed to `query`.

**Comment by DavidMStraub** [comment]
Done! The API is now
```
/api/search/?query=something&page=2&pagesize=20
```
and the results look like this:

```
{
  "hits": [
    {
      "handle": "b39fe2e143d1e599450",
      "object": {
        "gramps_id": "N0009",
        "private": false,
        ...
      },
      "object_type": "note",
      "rank": 0,
      "score": 5.522300827719273
    },
    ...
  ],
  "info": {
    "offset": 0,
    "pagecount": 1,
    "pagenum": 1,
    "pagesize": 2,
    "runtime": 0.0010769180007628165,
    "total": 2
  }
}
```

cdhorn:
Just looked it over, nothing major jumps out. This really adds a nice feature and was a great idea.

I used offset & limit for paging support in the other PR I opened. As we should be consistent I will change that to use page and pagesize later on tonight.

You forgot to add the endpoint to the apispec.yaml but other than that I say get it merged!

DavidMStraub:
- [ ] Add API docs

cdhorn:
Realized you overlooked the preamble too...

DavidMStraub:
- [ ] Add license preamble

DavidMStraub:
Added the preamble and API spec; ready to merge from my side!

(For the API spec, I wasn't sure what to put for `object` since it can be any kind of object (person, event, ...), so I left it empty for the moment.)

cdhorn:
Good morning!  Yeah the spec can always get refined later on.  Have you looked at flasgger before?

https://github.com/flasgger/flasgger

Not sure if we would want to add a dependency like that though or if licensing is compatible.

Merge away!

DavidMStraub:
> Have you looked at flasgger before?

Yes I had briefly looked at it; I have nothing against it, but my first impression was that it was overkill; adding the specs to the doc strings instead of manually might be cleaner, but I don't think shipping the whole Swagger UI along with the API is necessary. But if it reduces documentation work, we can consider it.

cdhorn:
Like everything it has pros and cons, but I do agree it would add a lot of unnecessary overhead. From a security standpoint I suppose you wouldn't want that.

Issue #75, opened 2020-11-14
License preamble


cdhorn:
@DavidMStraub @Nick-Hall Should we be including the license preamble and exclusion of warranty in the header for every source file and likely our copyright as well?  I notice Gramps follows that practice. Or is it deemed sufficient to have the LICENSE file in the package?

Nick-Hall:
Yes.  We should follow the recommendation in the appendix of the [license](https://github.com/gramps-project/gramps/blob/master/COPYING#L282).  Each file should contain the preamble and copyright lines.  See the main Gramps repository for examples.

cdhorn:
@DavidMStraub a PR to fix this will touch everything. Unsure if you want to put it together or if I should try to.  

May I propose as part of it that the files in resources for the major objects that are singular be renamed to match the plural endpoints? person.py -> people.py and so forth?

DavidMStraub:
I'm fine with both changes (and happy if I don't have to do it :wink:).

DavidMStraub:
Solved by #81.

PR #81, opened 2020-11-18
Refactor profile query parm, add preamble to all modules, add name-formats (Closes #71 #75)

### LINKED ISSUE #71: Profile without families
As mentioned in #70, the `profile` option for people leads to long computation time when `include_families` is switched on in the relevant method, so I think we have to provide a way to disable it, but still keep the profile for names & events.

Possibilities I can think off:

- Treat `profile` like `extend`: `profile=name,families,events`, `profile=all`
- New endpoints: `/api/people/profile`, `/api/people/handle/profile` showing only the profile and with query config options

Opinions?

cdhorn:
@DavidMStraub I know this seems big because the preamble is in every module but figured might as well get that over with at the same time. Changes are as follows:

1. Add the preamble. I tried to add copyrights to reflect who worked on what, please tweak as you see fit afterwards if you want to.
2. Renames the singular files in resources to their plural form to match their endpoints. 
3. Converts strip to a boolean as discussed for consistency.
4. Refactors profile to work like extend supporting the keywords all, self, families, and events and adjusts unit tests for it.
5. Adds a trivial /api/name-formats endpoint for sake of completeness and quick schema test for it.
6. Updates apispec.yaml for the above changes.  I also added the formats= parm for the /api/notes endpoints as I noticed it was missing.


gramps_webapi/api/resources/emit.py
```diff
@@ -36,7 +56,7 @@ def response(
         payload = payload or {}
         args = args or {}
         if "strip" in args:
-            self.strip_empty_keys = True
+            self.strip_empty_keys = args["strip"]
         else:
             self.strip_empty_keys = False
```
DavidMStraub, line 61: Why is this `if`/`else` still needed rather than just `self.strip_empty_keys = args["strip"]`? Perhaps needs adding a default value in the webargs schema.

DavidMStraub:
Looks good! If you address the comment and the merge conflict, good to merge from my side!

gramps_webapi/api/resources/emit.py
cdhorn: So it continues to work for any endpoint as opposed to only those derived from GrampsObjectResource and GrampsObjectsResource.

PR #82, opened 2020-11-18
Add /api/exporters endpoint (Closes #77)

### LINKED ISSUE #77: Add /api/exporters
Related to discussion in PR #59

cdhorn:
@DavidMStraub @Nick-Hall this is what I have for the /api/exporters endpoint so far.

The ExportOptions is a replacement for WriterOptionBox with no Gnome dependencies. It is refactored to a point where it does not really resemble the original, which provided the roadmap, so I am unsure if I should carry the copyrights in that module over?

If you guys could please look it over and let me know if you see any major issues with how I went about things.  Nick if you could please look at what I did for the /api/reports as well, I would really appreciate your time and feedback as you know the code far better than I and may see issues I had not thought of or realized.  Thanks!


DavidMStraub:
That looks great!!

The failing test is just a Python 3.5 ordering issue, perhaps due to an unordered dict somewhere, you can just turn the lists into sets in the test.

cdhorn:
Should have tested that on a stripped down Ubuntu VM first. 🙁

DavidMStraub:
We can install PyGObject on Travis, then I expect much of it will work:

```
sudo apt install python3-gi-cairo gir1.2-gtk-3.0
pip install PyGObject
```

DavidMStraub:
Sorry, I think my installations instructions were wrong, wait a second...

DavidMStraub:
```
sudo apt install libgirepository1.0-dev gcc libcairo2-dev pkg-config python3-dev gir1.2-gtk-3.0 
pip install pycairo PyGObject
```

cdhorn:
Problem looks like for Python 3.6 and up the pycairo package requires a version of cairo not available on xenial.
Is it okay to pin pycairo at 1.19.1?

DavidMStraub:
I rather suspect it's because I incorrectly suggested to install cairo with apt rather than pip above. Can you try whether it works with the updated command?

cdhorn:
Sure, will do

cdhorn:
Yep >3.5 all failed. 

pycairo 1.20.1 and up require cairo 1.15.10+ which is not available on xenial.

cdhorn:
Need to pin pycairo at 1.19.1 or switch build platform to bionic I guess.

DavidMStraub:
OK thanks for checking! Then I agree pinning it  to 1.19.1 (on Travis) is the best solution.

cdhorn:
Oh I guess that doesn't work the way I thought it does. Thoughts?

cdhorn:
Well that fixed the cairo issue at least. Now seems to be some Gramps packaging issue.

cdhorn:
@DavidMStraub I hope the Travis change is acceptable...

.travis.yml
```diff
@@ -1,12 +1,23 @@
...
+    python: 3.6
+    dist: bionic
+  - name: "Python 3.7 on Bionic"
+    python: 3.7
+    dist: bionic
+  - name: "Python 3.8 on Bionic"
+    python: 3.8
+    dist: bionic      
+before_install:
+  - sudo apt-get -y install appstream pkg-config libcairo2-dev gir1.2-gtk-3.0 libgirepository1.0-dev
```
DavidMStraub, line 17: I see you added the appstream dependency after a failing test where Gramps couldn't find the resource path. Can you elaborate, what does this do?

gramps_webapi/api/resources/emit.py
```diff
@@ -53,7 +53,8 @@ def response(
         args: Optional[Dict] = None,
     ) -> Response:
         """Prepare response."""
-        payload = payload or {}
+        if payload != []:
+            payload = payload or {}
```
DavidMStraub: This is a bit confusing. Can `payload` be anything other than a dict, list, or None? If not, I suggest to type it as `Optional[Union[Dict, List]]` (`Optional[Any]` is equivalent to `Any` anyway). Then
```python
if payload is None:
    payload = {}
```
would be much more explicit.

gramps_webapi/api/resources/exporters.py
```diff
@@ -0,0 +1,372 @@
...
+            return file_name, "." + extension
+
+
+class ExportersResource(ProtectedResource, GrampsJSONEncoder):
+    """Exporters resource."""
+
+    @property
+    def db_handle(self) -> DbReadBase:
+        """Get the database instance."""
+        return get_dbstate().db
```
DavidMStraub: Given this property is used by 3 classes, could it be turned into a function?

DavidMStraub (commented):
Very good!

I think we should also add `pycairo` and `PyGObject` to `extras_require` in `setup.py` then.

Do you have an overview which plugins will work with and without them installed?

What happens if they are not installed, which error is returned?

.travis.yml
cdhorn: It seems to be needed for the translation files to be built properly.  Why I do not know, I can credit Google with the finding as some users on Launchpad ran into the same issue.

gramps_webapi/api/resources/emit.py
cdhorn: Payload can be an empty list.  The statement `payload = payload or {}` was causing the empty list to be returned as an empty dictionary which is incorrect.

gramps_webapi/api/resources/exporters.py
cdhorn: Yeah I just sort of carried that all over the place from your use of it for some unknown reason, but yeah it is unnecessary as `get_dbstate().db` could be used directly.  I can factor that out.

cdhorn:
> I think we should also add pycairo and PyGObject to extras_require in setup.py then.

Will do. 

> Do you have an overview which plugins will work with and without them installed?

I did not look at all the different types of plugins that closely yet.  None of the exporters work at all without them though.  For reports only the text reports work and only a subset of output formats. 
 
> What happens if they are not installed, which error is returned?

An `ImportError` is raised on `import gi`, so for reports that is used as a test to determine which to enable/disable. See gramps_webapi.const in that PR.

.travis.yml
DavidMStraub: OK, thanks.

gramps_webapi/api/resources/emit.py
DavidMStraub: Right, I just found the code
```python
if payload != []:
    payload = payload or {}
```
confusing and thought 
```python
if payload is None:
    payload = {}
```
would be more understandable: the `payload` argument is only modifed when it is `None`, in which case it is set to an empty dict.

DavidMStraub:
I tried to understand why those plugins (like XML etc.) fail and I think in many cases it's just imports from `gramps.gui` that are not used at all when not invoked from the GUI. The cleanest solution would be to add a `try ... except` in Gramps itself in case `gi` is not available. But for now I think your approach is the best we can do.

DavidMStraub:
> An ImportError is raised on import gi, so for reports that is used as a test to determine which to enable/disable. See gramps_webapi.const in that PR.

I didn't find it, where exactly?

cdhorn:
Find the gramps_webapi/const.py in the diff for the other PR.  This snippet is used to control which are made available:

```
try:
    import gi

    REPORT_DEFAULTS = {
        CATEGORY_TEXT: "pdf",
        CATEGORY_DRAW: "pdf",
        CATEGORY_GRAPHVIZ: "gspdf",
    }
    REPORT_FILTERS = []
except ImportError:
    REPORT_DEFAULTS = {
        CATEGORY_TEXT: "rtf",
    }
    REPORT_FILTERS = ["pdf", "ps", "gspdf", "gvpdf"]
```

DavidMStraub:
Oh, sorry, missed that it's in the other PR. BTW, what's the status of the reports PR? Should that be merged before this one?

cdhorn:
I think both are ready.  There may be things to refine or change once you or others actually start using them, especially if someone sees a better way of doing it. 

Nick knows the code so well I was hoping if he had some time he could look them over as he could quickly point out any glaring issues, but I am sure he is busy with many other things.

cdhorn:
I'm rebasing and updating this one, so maybe merge it first when ready and then I'll prepare the reports one.  The span code is almost done I'll likely open a PR for that soon too.

gramps_webapi/api/resources/emit.py
cdhorn: Oh yeah... yes it is. Thank you, changed.

cdhorn:
Seems good to go...

Nick-Hall:
This looks good to me.  I suggest that you merge it.

You should include the copyrights of the original authors if their work influenced your code.

It has occurred to me that we need to change our programming guidelines so that code from the `gui` module is not used in certain types of plugin.  Ideally exporter options should be handled in a similar way to report options in the core Gramps code, but I'm not sure how easy sure a change would take to write.  Enclosing gui imports in a `try-except` block isn't the best solution, but may be acceptable.

cdhorn:
> You should include the copyrights of the original authors if their work influenced your code.

Done, thanks for taking a look over it as well Nick.

PR #83, opened 2020-11-18
Use JSON (not form) for token/user endpoints

DavidMStraub:
This changes the token and password endpoints to expect a JSON payload in POST rather than a form. I had used a form accidentally originally, but this doesn't make sense as we use JSON for the rest of the API. Adapted the unit tests accordingly.


cdhorn:
Makes sense and looks good to me!

Issue #85, opened 2020-11-20
Default language


DavidMStraub:
Although the new metadata endpoint returns the default locale fixing #73, this is unfortunately not enough for an application to fetch the strings in the default language. As an example, my locale is `de_DE`, and this is what I get for `locale.lang` under `/metdata`, but the corresponding Gramps language (and the language accepted by the `translations` endpoint) is `de`. In principle the application could also fetch the list of translations from `/translations` and do some internal logic, but that would require 3 HTTP calls just to get the strings, and would duplicate code that is in Gramps anyway. I also wouldn't like to simply take the first to characters of the locale, since Gramps *does* distinguish between `pt_BR` and `pt_PT`, and some other examples.

So in my opinion we should add a second key to the `locale` dict in the metadata endpoint given the actual Gramps language the default locale implies.

I think it would be essentially obtained with this code from `grampslocale.py`:

```python
        try:
            lang = _LOCALE_NAMES[lang_code][2]
        except KeyError:
            try:
                lang = _LOCALE_NAMES[lang_code[:2]][2]
            except KeyError:
                LOG.debug("Gramps has no translation for %s", lang_code)
                lang = None
        except IndexError as err:
            LOG.debug("Bad Index for tuple %s\n" % _LOCALE_NAMES[lang_code][0])
            lang = None
```

where `lang_code` is what is currently returned as `language`.

DavidMStraub:
On a related note, why are the underscores in locales replaced with hyphens in the current implementation of `/api/translations/`?

cdhorn:
Yeah I will undo that.  Are the locales not iso codes and the iso codes actually use - in them?  Or am I mixing up two or three different things? I think when I did that I set it to accept either, but yeah it should be consistent.

cdhorn:
So as best I could tell this indicates language codes use a dash: 'https://en.wikipedia.org/wiki/IETF_language_tag'

But this indicates POSIX platforms don't follow that standard: 'https://en.wikipedia.org/wiki/Locale_(computer_software)#POSIX_platforms'

In the Python locale library the getdefaultlocale method in here: 'https://docs.python.org/3/library/locale.html'

Actually references the original RFC which uses a dash here: 'https://tools.ietf.org/html/rfc1766.html'

Are they interchangeable to some extent or no?

DavidMStraub:
[...]

Issue #84, opened 2020-11-20
Problems with metadata


DavidMStraub:
I am having problems with the new metadata endpoint related to localization.

- The keys `Family Tree` and `Database` that are used in `family_tree_summary` are localized, but the English values are hard-coded. We need to use `gettext` to translate them or the dictionary will look up will fail, causing an `UnboundLocalError` below.
- The keys that are just converted to snake case (`number_of_places`) are also localized, so they end up as `anzahl_der_orte` etc. in my case. This is not good as we need those to be constant.

DavidMStraub:
@cdhorn I took the liberty of assigning the issue to you as you :slightly_smiling_face:

cdhorn:
Sure, I will take a look at it. Sorry if I didn't realize that.

DavidMStraub:
Fixed by #86 :tada:

PR #86, opened 2020-11-21
Fix broken locale/language handling (For #84 and #85)

cdhorn:
@DavidMStraub I think I have the right frame of reference now and this should work much better for you...


DavidMStraub:
Perfect, thanks for the quick fix! I tested the branch and it works on my machine as well. Good idea also to add the `incomplete_translation` key.

PR #87, opened 2020-11-22
Update filters, metadata, apispec.yaml

cdhorn:
@DavidMStraub this tweaks /api/metadata again, adds a base /api/filters that returns all the namespaces to be consistent with the other endpoints that return full collections, and contains some updates for the apispec.yaml including the various image endpoints.

### DIFF
--- a/gramps_webapi/api/resources/filters.py
+++ b/gramps_webapi/api/resources/filters.py
@@ -184,6 +184,23 @@ class CustomFilterSchema(FilterSchema):
     )
 
 
+class FiltersResources(ProtectedResource, GrampsJSONEncoder):
+    """Filters resources."""
+
+    @use_args(
+        {},
+        location="query",
+    )
+    def get(self, args: Dict[str, str]) -> Response:
+        """Get available custom filters and rules."""
+        results = {}
+        for namespace in GRAMPS_NAMESPACES:
+            rule_list = get_filter_rules(args, GRAMPS_NAMESPACES[namespace])
+            filter_list = get_custom_filters(args, GRAMPS_NAMESPACES[namespace])
+            results[namespace] = {"filters": filter_list, "rules": rule_list}
+        return self.response(200, results)
+
+
 class FiltersResource(ProtectedResource, GrampsJSONEncoder):
     """Filters resource."""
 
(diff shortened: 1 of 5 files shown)


gramps_webapi/api/resources/filters.py
```diff
@@ -184,6 +184,23 @@ class CustomFilterSchema(FilterSchema):
     )
 
 
+class FiltersResources(ProtectedResource, GrampsJSONEncoder):
+    """Filters resources."""
+
+    @use_args(
+        {},
+        location="query",
+    )
```
DavidMStraub, line 193: Just curious, what is the purpose of this empty schema?

gramps_webapi/data/apispec.yaml
```diff
@@ -131,6 +131,37 @@ paths:
         422:
           description: "Unprocessable Entity: Invalid token."
```
DavidMStraub, line 133: Thanks for adding this! Had forgotten it.

DavidMStraub:
Looks good! Thanks for completing the API spec.

DavidMStraub (approved):

gramps_webapi/api/resources/filters.py
cdhorn: Yeah I was unsure if I should add that, have done so in a few places but not everywhere it could be.  It is a way of having webargs validate the user didn't provide any query parms and if they did it treats it as an error.

gramps_webapi/data/apispec.yaml
cdhorn: Actually you know I should try to review all the error codes... we're probably missing a number of them on the endpoints.  Hmmm....

cdhorn:
You're welcome!

gramps_webapi/data/apispec.yaml
cdhorn: Yeah don't merge this yet.  I will look those over later tonight when I have time.

cdhorn: Hey shouldn't the /login and /refresh endpoints return 201 and not 200 as they are posts and respond with a new token?

cdhorn: And then update password would be a put not a post and return 200 as it updates the password?  Actually refresh is more like an update too since it's changing the token?

DavidMStraub: Hmm... since the login & refresh do not modify anything on the server side, I would say 200 is fine (by using JWT, there is no session handling on the server at all and no concept of users being logged in or not).

But you are right about using put and returning 200 for the password update!

Issue #88, opened 2020-11-24
age & span


cdhorn:
Provide profile= option to include age and an /api/events/{handle}/span/{handle} endpoint to calculate the span between two dates.

PR #91, opened 2020-11-26
Add age/span support; sort, offset & limit support; and locale support for profile and relationship calculator (Closes #32, #56, #88)

### LINKED ISSUE #32: Add paging
For collection endpoints that return a list of Gramps objects, we should add the possibility of "paging" to allow clients to lazily load objects in chunks.

I suggest to do this by adding `limit` and `offset` query parameters to `GrampsObjectsResource`.

To make this really useful, we should however also add sorting and it's not entirely clear to me how to do this generically. For instance, what would the query be to return the first 20 people in a list ordered by birth date?

cdhorn:
@DavidMStraub this one does the following:

1. Add age and span as keywords to the profile= query parm.  Age is self explanatory, the keyword span is for familial events and instead of being measured off birth is measured off of the marriage or marriage equivalent.  So if a couple had their 50th wedding anniversary the span would report "50 years", or if they divorced it would have the length of time the marriage lasted.  I did not add proper locale= support here yet, need to look at that. That can always come in another PR.

2. Add a /api/events/{handle1}/span/{handle2} end point that can be used to calculate the span between any two events. This supports locale={language} for the returned string.

3. I updated the relationship calculator again so it now uses the localized plugin if one is available and it too now supports locale={language} for the returned string when possible.

### DIFF
--- a/gramps_webapi/api/__init__.py
+++ b/gramps_webapi/api/__init__.py
@@ -32,7 +32,7 @@ from .file import LocalFileHandler
 from .resources.base import Resource
 from .resources.bookmarks import BookmarkResource, BookmarksResource
 from .resources.citations import CitationResource, CitationsResource
-from .resources.events import EventResource, EventsResource
+from .resources.events import EventResource, EventSpanResource, EventsResource
 from .resources.exporters import (
     ExporterFileResource,
     ExporterResource,
@@ -81,6 +81,9 @@ register_endpt(PeopleResource, "/people/", "people")
 register_endpt(FamilyResource, "/families/<string:handle>", "family")
 register_endpt(FamiliesResource, "/families/", "families")
 # Events
+register_endpt(
+    EventSpanResource, "/events/<string:handle1>/span/<string:handle2>", "event-span"
+)
 register_endpt(EventResource, "/events/<string:handle>", "event")
 register_endpt(EventsResource, "/events/", "events")
 # Places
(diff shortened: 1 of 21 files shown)


cdhorn:
@DavidMStraub so to add to the list above now:

4. I refactored GrampsObjectsResource so it now uses the get_%s_handles methods. Most have a sort available and I enabled by default, it is fast enough not sure we need to bother with an option to disable it. The ones that do not support sorting at moment are events, repositories, and notes. This really doesn't address sorting as a whole, but it gets you some sorted data to play with in the front end. 

5. Added offset & limit support to go with the above as they just slice the list.

I think this will be useful for the front end work, if it looks okay to you might want to merge it before the other two.

cdhorn:
For sorting do you see a need to be able to use sort keys based on different locales or is the active locale sufficient?

cdhorn:
@DavidMStraub I know this is growing.... added the initial pass at sorting support for collections.  sort= a comma delimited list of sort keys which vary by Gramps object class.  A + for ascending in front of a key is optional but permitted, a - indicates descending.  Sorts are applied from first in list to last. 

This is only at collection level and not intended to attempt to sort something like say the events returned in the individual person response. I'm not sure if there is a need for that.

cdhorn:
@DavidMStraub I am going to try hard not to add anything else to this.  I know we need unit tests for the sorting and localization support this adds but my thinking is to include those in a separate PR to address issue #92 I opened last night since I'll end up touching all the tests for it.

If possible I'd like to see the reports merged so I can rebase this before this is merged if that is okay with you.

Is there anything else you would like me to work on that is needed or would be useful for your front end before I work on issues #91 and #92 ?

DavidMStraub:
Hi, it's great there is so much progress, but it's really hard to review & to get an overview when there is such a huge number of features and changed files in a single PR. Is still all so intertwined that you cannot add it separately?

cdhorn:
I know I start running with something and get carried away sometimes. Let me figure out how to do that for you.

cdhorn:
Closing this, will try doing in smaller chunks.

DavidMStraub:
Thanks, I really appreciate it! I'll also do my best to review faster (smaller chunks should also make it easier to find the time during the week).

PR #96, opened 2020-11-29
Collection sorting and paging support (Closes #32)

### LINKED ISSUE #32: Add paging

cdhorn:
@DavidMStraub here is the sorting portion. Please see questions under issue #32 about this. I think it may make sense to treat this as phase one and phase two would change the response format but not sure how you would like things handled.

### DIFF
--- a/gramps_webapi/api/resources/base.py
+++ b/gramps_webapi/api/resources/base.py
@@ -132,6 +132,8 @@ class GrampsObjectsResource(GrampsObjectResourceHelper, Resource):
             "extend": fields.DelimitedList(fields.Str(validate=validate.Length(min=1))),
             "filter": fields.Str(validate=validate.Length(min=1)),
             "rules": fields.Str(validate=validate.Length(min=1)),
+            "offset": fields.Integer(missing=0, validate=validate.Range(min=0)),
+            "limit": fields.Integer(missing=0, validate=validate.Range(min=1)),
             "formats": fields.DelimitedList(
                 fields.Str(validate=validate.Length(min=1))
             ),
@@ -158,6 +160,8 @@ class GrampsObjectsResource(GrampsObjectResourceHelper, Resource):
             handles = apply_filter(
                 self.db_handle, args, self.gramps_class_name, handles
             )
+        if args["limit"] > 0:
+            handles = handles[args["offset"] : args["offset"] + args["limit"]]
 
         query_method = self.db_handle.method(
             "get_%s_from_handle", self.gramps_class_name
--- a/gramps_webapi/api/resources/base.py
+++ b/gramps_webapi/api/resources/base.py
@@ -161,7 +161,9 @@ class GrampsObjectsResource(GrampsObjectResourceHelper, Resource):
             obj = self.get_object_from_gramps_id(args["gramps_id"])
             if obj is None:
                 abort(404)
-            return self.response(200, [self.full_object(obj, args)], args)
+            return self.response(
+                200, [self.full_object(obj, args)], args, total_items=1
+            )
 
         locale = get_locale_for_language(args["locale"], default=True)
         query_method = self.db_handle.method("get_%s_handles", self.gramps_class_name)
@@ -189,6 +191,7 @@ class GrampsObjectsResource(GrampsObjectResourceHelper, Resource):
             200,
             [self.full_object(query_method(handle), args) for handle in handles],
             args,
+            total_items=len(handles),
         )
 
 
--- a/gramps_webapi/api/resources/base.py
+++ b/gramps_webapi/api/resources/base.py
@@ -94,23 +94,23 @@ class GrampsObjectResource(GrampsObjectResourceHelper, Resource):
 
     @use_args(
         {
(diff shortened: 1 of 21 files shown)


gramps_webapi/api/resources/emit.py
```diff
@@ -140,4 +140,7 @@ def default(self, obj: Any):
         if isinstance(obj, DbBookmarks):
             return self.api_filter(obj)
 
-        return JSONEncoder.default(self, obj)
+        try:
+            return JSONEncoder.default(self, obj)
+        except TypeError:
+            return None
```
DavidMStraub, line 157: When is this expected to happen? I think it would be good to add a comment.

gramps_webapi/api/resources/filters.py
```diff
@@ -132,13 +132,15 @@ def build_filter(filter_parms: Dict, namespace: str) -> GenericFilter:
     return filter_object
 
 
-def apply_filter(db_handle: DbReadBase, args: Dict, namespace: str) -> List[Handle]:
+def apply_filter(
+    db_handle: DbReadBase, args: Dict, namespace: str, handles: List = None
```
DavidMStraub: `List=None` is not a valid type hint. Also, I suggest never to use a bare `List`. In this case, probably `Optional[List[Handle]]`.

gramps_webapi/api/resources/base.py
```diff
@@ -58,6 +63,12 @@ def object_extend(self, obj: GrampsObject, args: Dict) -> GrampsObject:
             obj.extended = get_extended_attributes(self.db_handle, obj, args)
         return obj
 
+    def sort_objects(self, objs: List[str], args: Dict, locale) -> List:
+        """Sort the list of objects as needed."""
+        return sort_objects(
+            self.db_handle, self.gramps_class_name, objs, args, locale=locale
```
DavidMStraub, line 69: I find it confusing that the argument in `sort_objects` is sometimes `objs`, sometimes `handles`. This is where type hints and keyword arguments are very useful to improve readability (and reduce the risk of bugs).

DavidMStraub:
This is great!

DavidMStraub:
Do you plan to replace `limit` and `offset` by `page` and `page_size` (or whatever we may call them)? It's not much difference, but would be slightly more convenient to "use".

gramps_webapi/api/resources/emit.py
cdhorn: Well, it should not be expected to happen, but in playing with filters I stumbled into a situation where some PyGObject got passed into emit when I tried to retrieve the filter list and I couldn't track down the root cause. So this insures should something like that happen things don't break completely.

gramps_webapi/api/resources/filters.py
cdhorn: I can change it.

gramps_webapi/api/resources/base.py
cdhorn: Yeah I think I fixed some of the hints in a later commit not in this set.

cdhorn:
Yes this includes the switch to page and pagesize.

cdhorn:
Okay @DavidMStraub I think that should do it.

And idea why Travis isn't running?

gramps_webapi/api/resources/emit.py
DavidMStraub: OK, but if it's un unexpected error it might make sence to at least log it to allow debugging (`current_app.logger.error(...)`).

cdhorn: Good point, can I add that in next PR or want it in this one?

DavidMStraub: As you prefer.

DavidMStraub:
> And idea why Travis isn't running?

No, strange. I don't have access to the Gramps Travis account. @Nick-Hall?

Nick-Hall:
It may be because I created the build under _travis-ci.com_ instead of _travis-ci.org_.  I got a welcome email from Travis on 20th, that I was not expecting, suggesting that I sign up to an open-source plan.  Our existing repositories are already on an open-source plan, so I just created a build again via my _travis-ci.org_ login.

To confuse the issue, according to their [blog](https://blog.travis-ci.com/oss-announcement), they appear to be migrating to  _travis-ci.com_ anyway.

If we experience any more problems I'll send them an email.

cdhorn:
Hey @DavidMStraub are there others changes needed from me for this?

DavidMStraub:
No, looks good, ready to merge?

cdhorn:
Sure if you can, then I can rebase the locale support for profile= then a bit later tonight.

PR #97, opened 2020-11-30
Add password reset mechanism (fixes #89)

### LINKED ISSUE #89: Add password reset endpoint
... to enable a "forgot my password" link.

This will supply a special token that can be used to access only this endpoint. By using the current password's hash, it can be ensured that the token can only be used once while remaining stateless.

DavidMStraub:
This is still a work in progress, but almost done (e.g. API spec still missing). It adds a mechanism to reset a user password via e-mail. 

It works like this:

- Post a username to `/api/user/password/reset/trigger/` (this does *not* require a token)
- The backend sends an e-mail to the user containing a special token (this requires configuring an IMAP server via `EMAIL_HOST`, `EMAIL_PORT`, `EMAIL_HOST_USER`, `EMAIL_HOST_PASSWORD`, `EMAIL_USE_TLS`)
- The user clicks the link in the e-mail (I still need to fix the e-mail body) which takes them to `/api/user/password/reset/?jwt=<token>`
- At this endpoint (when using `GET`), a form is displayed to reset the password
- The form `POST`s the password to `/api/user/password/reset`
- Using the reset link again is *not* possible, because the token contains the hash of the old password, which allows to check that it has been used without keeping any state in the DB.

### DIFF
--- a/gramps_webapi/api/__init__.py
+++ b/gramps_webapi/api/__init__.py
@@ -63,7 +63,11 @@ from .resources.types import (
     DefaultTypesResource,
     TypesResource,
 )
-from .resources.user import UserChangePasswordResource
+from .resources.user import (
+    UserChangePasswordResource,
+    UserResetPasswordResource,
+    UserTriggerResetPasswordResource,
+)
 
 api_blueprint = Blueprint("api", __name__, url_prefix=API_PREFIX)
 
@@ -157,6 +161,12 @@ register_endpt(ExportersResource, "/exporters/", "exporters")
 register_endpt(MetadataResource, "/metadata/", "metadata")
 # User
 register_endpt(UserChangePasswordResource, "/user/password/change", "change_password")
+register_endpt(UserResetPasswordResource, "/user/password/reset", "reset_password")
+register_endpt(
+    UserTriggerResetPasswordResource,
+    "/user/password/reset/trigger",
+    "trigger_reset_password",
+)
 # Search
 register_endpt(SearchResource, "/search/", "search")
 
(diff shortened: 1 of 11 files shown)


DavidMStraub:
Done - I think! There still seem to be problems with Travis, but apart from that (and possibly failing tests), I have added everything I want.

cdhorn:
Seems to look good to me, that is good that the reset token can only ever be used once. 24 hours seems like a long time though, most sites seem to give you a half hour or so these days for that sort of thing. 

The thought occurs to me that maybe the endpoint structure should be /users to return list of users and then /users/{account}/password and so forth.  Is the {account} omission a deliberate measure to reduce potential information exposure?

DavidMStraub:
> Seems to look good to me, that is good that the reset token can only ever be used once. 24 hours seems like a long time though, most sites seem to give you a half hour or so these days for that sort of thing.

Yes we can also make it shorter (or configurable). One consideration I made is that, especially when using SMTP servers of small web hosters, it is quite likely that e-mails like that end up in the Spam folder. So when I consider some of the potential users of my database :older_woman: and how long it will take them to realize it etc., hours seemed a more likely time scale to me ;)

> The thought occurs to me that maybe the endpoint structure should be /users to return list of users and then /users/{account}/password and so forth. Is the {account} omission a deliberate measure to reduce potential information exposure?

Yes, you're absolutely right. We should change this. I was already considering doing it as part of this PR but then decided to postpone it, as it makes more sense doing it as part of #15.

This is needed in particular to allow an administrator to change other user accounts.

DavidMStraub:
> , hours seemed a more likely time scale to me

Then again, the reset link can be re-requested any time, so this is not really an argument.

cdhorn:
> Yes, you're absolutely right. We should change this. I was already considering doing it as part of this PR but then decided to postpone it, as it makes more sense doing it as part of #15.

Yep, makes sense to do it at the same time.

DavidMStraub:
BTW I changed the timeout to 1h now, so I think this is ready (even if Travis seems to have problems).

cdhorn (approved):
Looks good to me!

Issue #95, opened 2020-11-29
Hide all private records on per-request basis


DavidMStraub:
This is one of the most important missing features in my opinion: we need to add the possibility to only return non-private information on all endpoints. This can later (= soon) be enabled/disabled on a per-user basis depending on permissions (#15) but for now we can just treat it as a global option, too.

In my old backend, I solved this by simply [returning a `PrivateProxyDb`](https://github.com/DavidMStraub/gramps-webapp/blob/master/gramps_webapp/db.py#L52-L56) instead of the actual database in each request (this would be `get_dbstate` in our case). That's clean and very easy to implement. However, since the proxy DB only overloads a subset of the underlying DB's mathods ([see source](https://github.com/gramps-project/gramps/blob/master/gramps/gen/proxy/private.py)), I suspect we are currently using other, low-level, methods which return private data even when used on the `PrivateProxyDb`. So we need to carefully check all our code for such cases, find ways to make it safe, and add tests to make sure we don't leak private info.

Please feel free to use this thread to post any observations/checks you have made.

DavidMStraub:
Observation: the full-text search index contains the textual representation of all objects. Even if the search endpoint can make sure to only return objects that are not private, the textual representation can definitely contain private information. We can solve this by storing the full and the "stripped" textual representation in the index and returning the appropriate results at the search endpoint.

cdhorn:
The exporters make use of that and the other proxies in addition to filters. The thought occurred to me it might be needed elsewhere and might not cover all cases.

Stepping back, some part of me wonders if this is really the responsibility of the consuming application and not the API and whether we are not clearly delineating between the two.  

As I understand things, which may not be correct, your web front end is really a single application instance multi-user use case. As you have full control of the application the above argument could apply.

But I also recognize you would clearly want the access control in the API for a distributed multi-user use case where you do not control the application only the data presentation.  I guess this is how you are looking at things.

I still think, BTW, we need to support database management and import functionality for single user use cases with configuration options to enable/disable it when not being used in that manner.

DavidMStraub:
List of Db methods we use:

```python
db.method("get_%s_from_handle", gramps_class_name)
db.method("iter_%s_handles", gramps_class_name)
db.method("get_%s_from_gramps_id", gramps_class_name)
db.get_bookmarks()
db.get_family_bookmarks()
db.get_media_bookmarks()
db.get_event_bookmarks()
db.get_citation_bookmarks()
db.get_note_bookmarks()
[...]

PR #98, opened 2020-12-02
Add locale support for profiles (Closes #56)

cdhorn:
@DavidMStraub here is the last peice from this past weekend, the locale support for profiles. If this looks good to you please merge it.  I will add the unit tests for this and the sorting support in the next PR.

PR #99, opened 2020-12-02
Replace get_dbstate with get_db_handle

DavidMStraub:
The function `get_dbstate()` was almost always used as like this: `db_handle = get_dbstate().db`. So I now replaced it with a function that directly returns the handle.

The main reason for this change is to be able to return proxy DBs instead, to solve #95.

@cdhorn, the only non-trivial change is in metadata, which is the only place where the dbstate was used (not the db). But I replaced this with a fresh instance `DbState()` because I don't think it actually does anything that requires the DB; but it would be good if you could have a look to be sure.

### DIFF
--- a/gramps_webapi/api/file.py
+++ b/gramps_webapi/api/file.py
@@ -29,7 +29,7 @@ from gramps.gen.lib import Media
 from gramps_webapi.const import MIME_JPEG
 
 from .image import ThumbnailHandler
-from .util import get_dbstate, get_media_base_dir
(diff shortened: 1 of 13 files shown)


DavidMStraub:
The failing test is unrelated to the change & the same as in #98, some Python 3.5-only issue with searching.

cdhorn:
Yeah this makes sense, I'll look closer later tonight if I can but I'm sure it's probably fine.

cdhorn (approved):
Looks good to me!

PR #100, opened 2020-12-02
Implement role-based access control (fixes #15, fixes #95)

### LINKED ISSUE #15: Group-based user permissions

### LINKED ISSUE #95: Hide all private records on per-request basis

DavidMStraub:
This cannot be merged before #99 but is otherwise finished.

Some details on the implementation:

- The user database only stores a single integer, the user role. This was in fact already the case before.
- constants in `auth.const` map these integers to roles, e.g. `ROLE_OWNER`
- a constant dict in the same module assigns permissions, that are string constants, to roles. E.g., `PERM_VIEW_PRIVATE` is the permission to view private records
- The permissions are stored in the JWT, i.e. in case they are changed, the change is effective at the latest after 15 minutes (when the token needs to be refreshed)
- The JWT does not know anything about user roles, these are managed by the backend
- The database does not know anything about permissions, these are managed by the backend

The nice thing about this setup is that we can have a fine-grained permissions system but do not need to query the user DB for every request.

So far, I have only implemented the filtering of private records using `PrivateProxyDb` (#95), but the function `require_permissions(scope)` can be used anywhere in any method to raise a 403 if permission for a given action is not present.


DavidMStraub:
Yay, PR/issue no. :100: :tada:

cdhorn (approved):
This is very nice!

DavidMStraub:
Done I think. I had to make a slightly ugly patch in e1b69ad to make the metadata endpoint work in case a `PrivateProxyDb` is used (#95), because some of the methods are not defined for the proxy.

Nick-Hall:
Yes.  It would be neater to add these methods to the [`ProxyDbBase`](https://github.com/gramps-project/gramps/blob/master/gramps/gen/proxy/proxybase.py#L81) class.

DavidMStraub:
I can submit a PR. Once this has landed in Gramps, we can remove the patch again.

DavidMStraub:
I would merge this in a couple of hours if there are no objections (we can always improve on it).

cdhorn:
Certainly none here, nice work!

PR #103, opened 2020-12-04
Add profile/locale/strip to search endpoint (Closes #102)

### LINKED ISSUE #102: Profile in search
To display a search results page, I realized the raw objects are not really useful, e.g. it is difficult to display dates for events or names for people. So what would really be needed would be to have a `profile` option also in the objects in the search result endpoint. Unfortunately, that's a bit difficult to achieve at the moment because all the profile logic is in the individual, different gramps object endpoints. So I think we should think about refactoring this so we can use it also for the search results.

cdhorn:
Hey @DavidMStraub this is for issue #102 as the profile assembly is handled by routines in `gramps_webapi.api.resources.util` so there really is little needed to add it. 

The locale keyword affects profile of course and since strip operates recursively on the result set I threw that in as well.

I'll throw tests for these options in the JWT PR when I put that together sometime in next few days.

### DIFF
--- a/gramps_webapi/api/resources/search.py
+++ b/gramps_webapi/api/resources/search.py
@@ -25,12 +25,18 @@ from typing import Dict
 from flask import current_app
 from gramps.gen.db.base import DbReadBase
 from gramps.gen.lib.primaryobj import BasicPrimaryObject as GrampsObject
-from webargs import fields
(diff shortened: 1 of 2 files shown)


DavidMStraub:
That's brilliant, I hadn't grasped that, thanks!

PR #112, opened 2020-12-08
Add gramps_id to person and family profiles (fixes #107)

### LINKED ISSUE #107: Children Gramps IDs
We currently don't return the Gramps IDs of a person's children; the `child_ref_list` only contains the handle. I suggest to add the Gramps ID to the person profile.

DavidMStraub:
As described in #107; also added it to the family profile while I was at it.


DavidMStraub:
Need to add the API docs...

cdhorn:
Looks fine, simple enough.  We should probably add it to the event profile as well.

DavidMStraub:
Done.

cdhorn (approved):
Only suggestion would be to perhaps add gramps_id to event profile unless you haven't found a use case for it in the front end yet.

Why don't you get your 2 PR merged and then I can rebase the unit tests one since there will be conflicts to resolve?

DavidMStraub:
OK thanks, I'll merge them!

For events, I didn't find it so important as the same information is contained in `extended.events`. For children and families, it's different as the `child_ref_list` and doesn't contain a `gramps_id`. The families in `extended` do, but they are organized slightly differently (`primary_parent_family` & `parent_families` vs.`primary_parent_family` & `other_parent_families`) so that it's a bit inconvenient to extract the Gramps IDs.

PR #113, opened 2020-12-08
Add place profile (fixes #104)

### LINKED ISSUE #104: Place profile
Currently, the `/places` endpoint does not contain much "readable" information. It would be great to have a profile option with formatted place names and the entire place hierarchy.

DavidMStraub:
Looking into this, I realized again how complex places are in Gramps (even before more possible complications coming in 5.2). They can have multiple names in different languages and valid at different times; they can have multiple parents and multiple children in the hierarchy.

This PR is just a simplistic version which ignores the time dependence. It only returns the place hierarchy for the most recent names. Further details could be added later.

For an example of what this looks like, see the added unit test.

Still need to add the docs.

### DIFF
--- a/gramps_webapi/api/resources/places.py
+++ b/gramps_webapi/api/resources/places.py
@@ -21,17 +21,35 @@
 
 """Place API resource."""
 
+from typing import Dict
+
+from gramps.gen.const import GRAMPS_LOCALE as glocale
+from gramps.gen.lib import Place
+from gramps.gen.utils.grampslocale import GrampsLocale
+
 from .base import (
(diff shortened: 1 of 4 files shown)


DavidMStraub:
Done! Apart from the docs, I also added one more feature, namely latitude/longitude in the place profile. Originally I hadn't thought about this, but the problem with the lat/long in the raw object is that it can have arbitrary string format, which is not useful for consumption by apps. So in the profile I now put a standard format (float) using Gramps' own utility methods.

cdhorn (approved):
Looks good!

Issue #106, opened 2020-12-05
Bug: Sort key for sort by types not localized properly


cdhorn:
Found during unit tests. 

Need to translate type name before returning sort key.

cdhorn:
Should be closed by PR #115

PR #115, opened 2020-12-11
Refactor endpoint unit test suite and encoder (Closes #92, #105, #106)

### LINKED ISSUE #92: Refactor test_endpoints to use JWT
While we know endpoint classes derived from ProtectedResource are protected with JWT the test suite for the endpoints should actually be validating that for each endpoint from a QA standpoint.

### LINKED ISSUE #105: Bug: keys filter broken in some contexts
Found while refactoring unit tests. 

Use of keys=person_ref_list,extended in combination with extend=person_ref_list or extend=all is filtering child person objects.

cdhorn:
@DavidMStraub it took a while but this is largely for #92 and fixes some bugs found along the way.

The test fixture now enables authentication so all of the tests require the use of tokens.  By default they run with the owner role but I created a test user for each role in the database.  There should be a test for every method on every endpoint to insure a token is required to access it unless I overlooked one. 

A fetch_token helper function is available and you can specify a different role when calling it as well.  The run_test helper functions were replaced with a new set of check functions that are a little more targeted and many of those can also be passed a different role. Before the run_test ones lumped lots of tests together and I tried to break them all out as much as possible now as well.

Backlink tests were added to all major objects, not just person. Sorting and paging tests were added. Tests for the event span support were added. Tests for the parameters added to search were added. And numerous other tests were added or expanded along the way.

Finally a refactor of emit.py was required to fix #105 so one of the new tests would pass. It no longer subclasses JSONEncoder as there was no way to make it aware of the level of recursion. 

There is one test that fails for me in test_user.py but that one is currently broken in master as well. I did look at it a little, it seemed as if the token or header was not being constructed properly but I was not able to figure out the cause yet. I'm hoping you are able to track that one down.


DavidMStraub:
Great! Will review it later today. After merging #121 and #119 we can see if there is still a problem with the tests.

cdhorn:
Agree better to get the other two in first.

gramps_webapi/api/resources/bookmarks.py
```diff
@@ -20,10 +20,11 @@
 
 """Bookmark API resource."""
 
-from typing import List, Optional
+from typing import Dict, List, Optional, Union
 
 from flask import Response, abort
 from gramps.gen.db.base import DbReadBase
+from webargs.flaskparser import use_args
```
DavidMStraub: After #121, `use_args` would have to be imported from `api.util`.

DavidMStraub:
That looks great! I haven't gone over all the tests in detail, but the general approach is excellent and will also simplify things going forward.

Concerning the failing test: I don't understand why in the Travis log it says

```python
        # try empty PW!
        rv = self.client.post(
            BASE_URL + "/user/password/reset/",
            headers={"Authorization": "Bearer {}".format(token)},
            json={"new_password": ""},
        )
>       self.assertEqual(rv.status_code, 422)
E       AssertionError: 400 != 422
```

because in master it is actually `self.assertEqual(rv.status_code, 400)` :thinking:

gramps_webapi/api/resources/bookmarks.py
cdhorn: Just rebased and pushed a commit for this.  This was added in this PR and not there before.

cdhorn:
Regarding test_users.py it still fails for me, not there but here:

```
        # now that should work
        rv = self.client.post(
            BASE_URL + "/user/password/reset/",
            headers={"Authorization": "Bearer {}".format(token)},
            json={"new_password": "789"},
        )
>       self.assertEqual(rv.status_code, 201)
E       AssertionError: 422 != 201
```

When I added debugging code it looked like the third part of the token was malformed and not base64 encoded.

cdhorn:
Oh I see what you meant in master. For me before I am pretty sure it was failing as above. It fails as above in this PR.

cdhorn:
Oh because I had changed 400 to 422 in the test above it.

DavidMStraub:
> When I added debugging code it looked like the third part of the token was malformed and not base64 encoded.

Oh yes, I had that problem as well. Capturing the email using the mocked SMTP server and then extracting the token is annoyingly difficult because it seems that the message string is sometimes base64 encoded, sometimes not. I still haven't understood why, but I thought it had worked at some point.

cdhorn:
@DavidMStraub took a while to realize it but the sort failures were because pyICU was missing. It's not required but recommended for Gramps, I added to the .travis.yaml to insure consistent test results. Notice test_user.py was success this time, but it still fails on my machine at home in both branches and I have not been able to figure out why yet. Anyway if this looks good to you I'd appreciate if you could merge it.

PR #116, opened 2020-12-11
Fix regression in place profile

DavidMStraub:
I made a silly mistake, changing `alternative_names` to `alternate_names` to be more consistent with the people profile in #104, but accidentally also renamed a method that is actually called `get_alternative_names` in `gramps`.


cdhorn:
Just ran into that, please merge this too!

cdhorn:
You lost extend too!  Need to add to places.py after the profile check

Issue #117, opened 2020-12-11
extend broken for places 


cdhorn:
Adding object_extend to places.py overrides the base method so the extend= functionality was lost.

DavidMStraub:
Sorry! I will fix it tomorrow.

DavidMStraub:
Fixed by #118.

PR #118, opened 2020-12-11
Fix extend regression for places

cdhorn:
@DavidMStraub here is fix for #117


DavidMStraub:
Thanks for cleaning up after me!

I'm going to merge this inspite of some tests failing. I suspect this might be due to the new webargs 7.0.0 that was released 2 days ago. I will open a separate issue.

Issue #114, opened 2020-12-09
Travis fails on Python 3.5


DavidMStraub:
There is an issue with the unit test for the search endpoint.

cdhorn:
Should be closed by PR #115 as I removed the 3.5 tests in the travis update when I added the pyICU stuff.

PR #121, opened 2020-12-12
Fix unit tests (fixes #120)

### LINKED ISSUE #120: Tests fail - new webargs version?
In the latest commits, some tests fail. This might be due to the new webargs version: [changelog](https://webargs.readthedocs.io/en/latest/changelog.html).

DavidMStraub:
I believe this fixes #120. I subclassed `webargs.FlaskParser` (as per the documentation) to make unknown query arguments fail with a 422. This is tested in some of the tests and does not happen by default anymore in webargs 7. This should also be backwards compatible as it only sets a class attribute on the subclass.

The only failing test is Python 3.5 due to #114, but we might get rid of it anyway...

### DIFF
--- a/gramps_webapi/api/__init__.py
+++ b/gramps_webapi/api/__init__.py
@@ -24,7 +24,7 @@ from typing import Type
 
 from flask import Blueprint, current_app
 from webargs import fields, validate
-from webargs.flaskparser import use_args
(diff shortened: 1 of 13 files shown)


cdhorn:
Looks good, I assume you'll merge shortly...

PR #127, opened 2020-12-18
Add new database manager class (fixes #16)

### LINKED ISSUE #16: Multithreading
See https://github.com/DavidMStraub/gramps-webapp/issues/26

This will be solved, at list for read-only access, by using https://github.com/gramps-project/gramps/pull/1119, once merged.

DavidMStraub:
Since the PR https://github.com/gramps-project/gramps/pull/1119 seems to have stalled (and indeed I'm not sure anymore this is the best way to solve it), I went ahead and added a new DB manager class that is a combination of Gramps's `CliDbManager` and `CliManager` classes; however it is *much* more concise since we can get rid of all the error handling. It's not necessary since any error raised during the handling of the database will simply cause a 500 Server Error in flask, which is the appropriate response (but the server will not crash).

What it provides in the end is a boolean argument on `get_db` that allows to disable writing a lock file. This is disabled by default now, since we are not writing anything to the DB yet anyway. In the future, we can decide whether a given operation should write a lock file (prevent all other uses from writing to the DB). My guess is that we'll end up locking the DB for most `PUT` request (which modify a resource, and we want to prevent conflicting modifications), while I believe we can omit locking for most `POST` requests (that add a new resource), since adding a new object to the DB cannot lead to a conflict (they are unique thanks to their handle).

Anyway, for the moment all this does is fix #16: we can now run the web server with multiple threads/processes (which is badly needed for a multi-user application).

### DIFF
--- a/gramps_webapi/__main__.py
+++ b/gramps_webapi/__main__.py
@@ -44,8 +44,7 @@ def cli(ctx, config):
 def run(ctx, port):
     """Run the app."""
     app = ctx.obj
-    # threading is disabled to avoid problems with locked databases
-    app.run(port=port, threaded=False)
+    app.run(port=port, threaded=True)
 
 
 @cli.group("user")
--- a/gramps_webapi/dbloader.py
+++ b/gramps_webapi/dbloader.py
@@ -1,0 +1,148 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2001-2006  Donald N. Allingham
+# Copyright (C) 2009 Benny Malengier
+# Copyright (C) 2020 David Straub
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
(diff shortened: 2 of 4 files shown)


cdhorn:
Why write a lock file when you could use a global Lock object?

DavidMStraub:
What do you mean by global object? That wouldn't be threadsafe, would it?

cdhorn:
I probably don't know enough about how Flask works. Wouldn't a global instance of a Lock() from the multithreading module be safe to use?

DavidMStraub:
Well, I'm not an expert either, but there are many ways to run a WSGI application like flask, with multiple threads or processes ... so I don't think `multithreading.Lock` works with all of them. Moreover, if we really want to go down the multi-database road at some point (i.e. the database can differ for each user), we would anyway need a per-DB lock and not a global one.

DavidMStraub:
This can be merged, right?

cdhorn:
Yes!  You have really been on a roll, great work!  I checked out latest Gramps.js again a couple days ago too and great progress on it.

PR #128, opened 2020-12-18
Fix total size

DavidMStraub:
That's a mini-PR ... I noticed that the `X-Total-Size` header was wrong for all endpoints since the length of the handle list was taken after applying the pagination, so it was always equal to the page size. I simply moved it up.


cdhorn:
Well that's an embarassing fubar!  Sorry about that.

PR #131, opened 2020-12-19
Add timeline support (Closes #123)

### LINKED ISSUE #123: Add /people/{}/timeline and /families/{}/timeline support
I propose adding support for timelines to the people and families endpoints.   This differs from the events list in that it would include optional events for spouses, parents, children, and siblings. 

A returned timeline event would consist of the event type, description, date, place name, latitude, longitude, age of the person at time of the event.  Some optional attributes might be number of citations supporting the event, highest citation confidence rating, and links to media objects associated with the event. 

The idea is to assemble and provide all the information needed to construct a timeline view of a person in a manner similar to that supported by the FamilySearch timeline feature but perhaps with some additional options and information.

cdhorn:
@DavidMStraub I think this mostly does what I sort of wanted to try to do with it now. Still need to write unit tests but wanted to open the PR so you could take a look over it.

Note the timelines it can assemble vary by endpoint, they are not all quite the same.

For /people/{handle}/timeline it is specific to the person. With the relation events you can adjust ancestors and offspring generations to evaluate, so if a grand parent died or grand child was born that too could be included on the timeline. When anchored to a person like this only events during the life of the person are included, but you can "open" either end of the range so it could include events before their birth or after their death as well. 

For /timelines/people/ you can optionally set an anchor person. Then you can pass a list of handles, otherwise it examines everyone in the database which is a bit overkill. But you can also apply filter rules to narrow that down to a more specific set of people.

For /familes/{handle}/timeline it provides the events for all members of the family from birth to death.

For /timelines/families/ you can again pass a list of handles and it also lets you apply filters.

Note I allowed for filtering events by category and not by constructing a list of specific events. It seemed to make sense at the time but I am wondering if maybe I should add some options to make it more fine grained. I used the Gramps grouping, which does some things that do not make sense to me like treating baptism, a religious event, as a vital event and naturalization, a legal event, as a travel event.  

An event with no date is discarded by default, as there is no way to sort it or know where it belongs on the timeline.  I think event refs are stored in a specific user defined order so I suppose in theory you could look across all the people whose events you are merging into the timeline and try to fit undated ones in where it makes sense as you build up the list in sequence. I'm not sure I'm up to tackling something like that at the moment, maybe someone more enterprising will come along and take up the challenge or I will revisit it down the road. 

It is hard to tell if the ones other than /people/{handle}/timeline are really useful or not. Guess time will tell, but was not hard to add them once the base person one was working.


DavidMStraub:
That's really cool!

The only issue I have noticed is that the `label`s are not properly translated as "of" is hard-coded, which leads to strings like `Geburt of Sohn` for me. Getting this right for all languages is probably tricky, so maybe we can just be a little bit less ambitious and reder it as `Geburt (Sohn)` etc. I suspect (though I'm not sure) that this would be meaningful in all languages.

cdhorn:
Yes when I was playing with locale I noticed that and was surprised it did not seem to translate.  That is not a bad idea to use () I will probably change it to go with that.

cdhorn:
@DavidMStraub I think this is ready to merge if you don't see anything glaringly wrong.  I added fine grained event filtering and pagination support to round things out and test cases are done.

Please note I modifed emit.py so if lat or long are null they will emit as 0 for schema conformance. As far as I can tell this is not uncommon as per https://en.wikipedia.org/wiki/Null_Island

PR #132, opened 2020-12-19
New and improved user endpoints

DavidMStraub:
This improves the token and user endpoints, making them a bit more restful and adding some new features:

- The `/login/` endpoint is renamed to `/token/`. "login" was a misnomer since we have stateless auth.
- The `/refresh/` endpoint was renamed to `/token/refresh/` for consistency
- All `/user/...` endpoints have been moved to `/users/<user_name>/...` which is more consistent with the other endpoints and allows to implement editing other users (given sufficient permissions)
- The shorthand `-` is used for the currently authenticated user, which allows to apply the appropriate permissions (e.g. `EditOwnUser` and `EditOtherUser`)
- A new `GET` endpoint `/users/` returning details about all users (requires the permission `ViewOtherUser`)
- A new `GET` endpoint `/users/<user_name>/` returning details about a specific users (requires the permission `ViewOtherUser` unless `user_name` is `-`)
- A new `POST` endpoint `/users/<user_name>/` to add a new user
- A new `PUT` endpoint `/users/<user_name>/` to update an existing user

I added unit tests for all cases.

Thinks that can be improved in the future:

- There is no endpoint to change a user's role yet (need to think about how to do it best)
- The update password is still a `POST` (need to think about how to deal with the `old_password`, which is known to the user changing their own pw, but not to an owner)

I will leave these improvements to a future PR.


cdhorn:
Looks great!   You forgot to add the 2 new endpoints and their methods to the apispec.yaml

DavidMStraub:
Done!

PR #133, opened 2020-12-19
Improvements to search (fixes #110, fixes #111, fixes #130)

### LINKED ISSUE #110: Cannot search for repository and note Gramps IDs
This is actually a Gramps bug IMO: all event types except those two have `self.gramps_id` as part of their `get_text_data_list`, so can be searched for.

### LINKED ISSUE #111: Search index w/o private records
As discussed in #95, in order to prevent search hits from showing up when searching for keywords that only appear in private attribiutes, we need a separate search index for public-only text strings.

### LINKED ISSUE #130: Cannot search for surnames
Another peculiarity of Gramps's `get_text_data_list`: a person's `get_text_data_child_list` contains the `primary_name` and `alternate_name`, but these are `Name` objects that only contain the first name in their `get_text_data_list`, while the `surname_list` is another child object of this, so one would have to go down another level in the `get_text_data_child_list`. Without doing this, the indexer doesn't store the person's surname.

DavidMStraub:
This makes some improvments to full-text search and solved some existing issues.

- The issue #110 that Gramps IDs are not indexed for repos and notes is solved and a "forward-compatible" way by explicitly checking if an object's Gramps ID is part of the text data list or not, and is added if not
- The issue #130 that surnames cannot be searched for is solved by iterating one level further in child objects in case of `Name`s
- The issue #111 that private should not be searchable for unauthorized users is solved in a twofold way:
  - A new boolean field `private` in the index which is required to be `False` for in searches triggered by users without `ViewPrivate`
  - The text data list is split into a "public" and a "private" part (e.g. in case of private attributes of public objects). By using two different search schemas for users with and without `ViewPrivate`, both fields are searched or just one of them as appropriate. This makes sure no private information is leaked, but also no data has to be stored twice.

Furthermore, I added the last changed timestamp to the index which now allows date searches like
```
type:repository changed:'2012 to now'
```

### DIFF
--- a/gramps_webapi/api/resources/search.py
+++ b/gramps_webapi/api/resources/search.py
@@ -28,8 +28,9 @@ from gramps.gen.lib.primaryobj import BasicPrimaryObject as GrampsObject
 from gramps.gen.utils.grampslocale import GrampsLocale
 from webargs import fields, validate
 
-from ..util import use_args
-from ..util import get_db_handle, get_locale_for_language
+from ...auth.const import PERM_VIEW_PRIVATE
+from ..auth import has_permissions
+from ..util import get_db_handle, get_locale_for_language, use_args
 from . import ProtectedResource
 from .emit import GrampsJSONEncoder
 from .util import (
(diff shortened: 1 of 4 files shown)


cdhorn:
This too, merge it in!

PR #134, opened 2020-12-20
Implement configurable internal links in HTML notes (fixes #122)

### LINKED ISSUE #122: Internal links in HTML notes
As discussed already in #26, it would be nice to render internal links in the notes HTML output as actual links (currently they are just underlined text), but that would require making the HTML code to be rendered for the link configurable via query parameters, as it will be app-dependent.

DavidMStraub:
I implemented this now by adding a new option `format_options` on the note endpoint. It allows specifying a format string like e.g.

```
{"link_format": "{obj_class}/{gramps_id}"}
```
or
```
{"link_format": "http://myapp/objects/{obj_class}/{handle}"}
```

I decided against using `jinja2` for the time being ... it would allow additional modifications, but I think for apps that have special needs, it will anyway be necessary to do some post-processing. For instance, if an app uses a URL structure like narrative web, where event objects are under an `evt` URL, sources under `src`, etc., this would be very tricky to do even in `jinja`. But in that case I think it's simpler if the app uses a pattern like `__{obj_class}__` and then replaces `__event__` by `evt` etc.

### DIFF
--- a/gramps_webapi/api/html.py
+++ b/gramps_webapi/api/html.py
@@ -20,10 +20,15 @@
 
 """HTML backend for styled text."""
 
+from typing import Callable, Optional
+
 import bleach
 from gramps.gen.lib import Note, NoteType, StyledText
 from gramps.plugins.lib.libhtml import Html
 from gramps.plugins.lib.libhtmlbackend import HtmlBackend, process_spaces
+
+from .util import get_db_handle
+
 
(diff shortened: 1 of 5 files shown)


cdhorn:
This is perfect.  Don't forget to add the option to apispec.yaml...  Merge this up too!

PR #135, opened 2020-12-20
Add possibility for configuring a static asset folder

DavidMStraub:
To host a static HTML frontend (such as [Gramps.js](https://github.com/DavidMStraub/Gramps.js)), one can simply use a separate web server and point it to the API (using CORS). However, an even simpler possibility (e.g. for a single, self-hosted tree) that does not require CORS is to just serve the static assets directly via flask.

This PR adds a config option `STATIC_PATH` that defines a folder where static files will be served automatically by flask. By default, it points to a file with a single, empty `index.html`. When used, there is a catch-all route that serves any existing file, but serves the `index.html` in the `STATIC_PATH` for non-existing paths. This allows to use this setup for single-page apps (see [the flask docs](https://flask.palletsprojects.com/en/1.1.x/patterns/singlepageapplications/)).

To avoid routing also erroneous API calls (`/api/nonexistantendpoint`) to the static folder, the catch-all route explicitly checks whether the path starts with `api` and fails with a 404 if it does.


DavidMStraub:
The only downside is that calls to existing API endpoints with the wrong method will return a 404 instead of the appropriate 405. However I'd prefer to open a separate issue for this and solve it later.

cdhorn:
Looks good, makes sense to do this.

PR #136, opened 2020-12-21
Missing profiles in search, role in profile events, suppress indexing error

DavidMStraub:
This contains 3 smaller changes:

- It adds 3 missing profiles for the `search` endpoint: citation, media, place
- It adds a `role` property to the event object in the profiles for person and family. This is useful e.g. to display whether a given event is a Primary or just a Witness one
- It suppresses a `LockedError` raised by `whoosh` when running with multiple threads (and only one of them will be able to recreate the search index)


cdhorn (approved):
Looks good to me!

PR #137, opened 2020-12-21
Add reference profile

DavidMStraub:
This adds a "profile" version of the "extended backlinks". This seems a bit excessive and is partly redundant (the references for families are children and parents, etc.), but I wasn't able to solve some frontend issues without it, for instance:

- Show all references to a given object. For this, `backlinks` is not sufficient as it is e.g. impossible to display the primary participant of a birth event
- Show the names of a people tagged in photos. While the person object is part of the extended `media_list` (and this also contains the necessary `rect`), displaying the name would require implementing a name displayer on the frontend side, which is not easy at all, in full generality.

Unit tests are still running and I expect I might have to fix some tests, apart from that I think am done.


cdhorn (approved):
Looks good enough I suppose!  By done I guess you mean this should be the last missing peice needed for the first iteration of Gramps.js? If so that's great!

DavidMStraub:
I hope so :wink:

DavidMStraub:
@cdhorn I am a bit desparate about those missing tests ... unfortunately I cannot test this properly locally at the moment since I have different Gramps settings & locale in my development setup and it would be tedious to change it. If you have time, could you please have a look what is still missing for the two failing test methods? Thanks!

cdhorn:
Sure, I will take a look later tonight.

PR #139, opened 2020-12-22
Add /api/facts endpoint

cdhorn:
@DavidMStraub this enhancement adds a new `/api/records/` endpoint that exposes the various facts gathered by `find_records` in `gramps.plugins.lib.librecords`  That library collects the information provided in the records text report and also a gramplet. The report usually defaults to a rank of 3, I choose to default to 1, but it is of course configurable. The `apispec.yaml` has been updated and unit tests are included so if you do not see any issues this should be ready to merge.


cdhorn:
In hind sight I wonder if I should have named this /api/facts/ instead. Actually, yeah, going to change it quick.

DavidMStraub:
Just as I was wondering what to put on the front page. Nice!

Just waiting for the unit tests to complete :rocket:

PR #140, opened 2020-12-23
Add /api/living endpoint

cdhorn:
@DavidMStraub this enhancement adds `/api/living/{handle}` and `/api/living/{handle}/dates` endpoints to provide access to the `ProbablyAlive` calculator in `gramps.gen.utils.alive`  It includes the `apispec.yaml` update and unit tests so unless you notice anything wrong should be ready to merge. 

I think maybe with this most of the core Gramps functionality is now exposed through the API.  The exceptions would be related to database and configuration management and the importers.

There is some kind of holiday calculator I noticed.  Maybe I'll add that too although I'm unsure how useful it may be outside of reports.

@Nick-Hall @romjerome have we missed anything that you have noticed that might be useful for a view only / presentation style application?

### DIFF
--- a/gramps_webapi/api/__init__.py
+++ b/gramps_webapi/api/__init__.py
@@ -40,6 +40,7 @@ from .resources.exporters import (
 from .resources.facts import FactsResource
 from .resources.families import FamiliesResource, FamilyResource
 from .resources.filters import FilterResource, FiltersResource, FiltersResources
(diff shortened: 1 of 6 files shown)


cdhorn:
Okay @DavidMStraub now this one is ready. I'll rebase the other once this is done.

PR #141, opened 2020-12-23
Fix people and families profile tests

cdhorn:
@DavidMStraub this should fix the tests for you.


cdhorn:
Well that's interesting, passes fine on my home machine.  Hmmm.  Maybe we should short cut this one and just spot check some key values, after all if the schema check passes we know the structure should be correct so we would just want to verify some calculated/displayed values are correct. Is late I need to get some 😴

DavidMStraub:
Thanks!! I'll merge the other PR and then we can fix the tests in master.

PR #142, opened 2020-12-23
Add /api/holidays endpoint

cdhorn:
@DavidMStraub figured why not, this is for the holiday calculator thingy.  Includes apispec.yaml and a few unit tests so unless you see anything wrong should be ready to merge.


DavidMStraub:
Cool! This is ready to merge, right?

cdhorn:
Yes should be.

PR #144, opened 2020-12-23
Sorting search results

DavidMStraub:
This adds a new `sort` parameter that allows to sort search results by a field (any not by score). The only meaningful possibility at the moment is `changed` to sort by the last changed date. I used the same convention as for the other endpoints: `-changed` is descending.

I also fixed the unit tests (I hope) as part of this PR.

Will add API docs in a moment.

### DIFF
--- a/tests/test_endpoints/test_search.py
+++ b/tests/test_endpoints/test_search.py
@@ -223,7 +223,7 @@ class TestSearch(unittest.TestCase):
         header = fetch_header(self.client, role=ROLE_OWNER)
         rv = self.client.get(
             TEST_URL
-            + "?query={}".format(quote("type:repository changed:'2012 to now'")),
+            + "?query={}".format(quote("type:repository change:'2012 to now'")),
             headers=header,
         )
         self.assertEqual(len(rv.json), 1)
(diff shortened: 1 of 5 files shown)


tests/test_endpoints/test_search.py
```diff
@@ -239,3 +239,42 @@ def test_get_search_explicit_fields_guest(self):
...
+            TEST_URL
+            + "?sort=changed&pagesize=1&page=1&query={}".format(
+                quote("type:person changed:'1990 to now'")
+            ),
+            headers=header,
+        )
+        self.assertEqual(len(rv.json), 1)
+        self.assertEqual(rv.json[0]["object"]["gramps_id"], "I0552")
+
+    def test_get_search_oldest(self):
```
cdhorn: Same code starts repeating? Copy paste?

cdhorn (commented):
Looks good other than duplicate test and missing apispec.yaml update.

tests/test_endpoints/test_search.py
DavidMStraub: ... or testing it twice just to be sure :rofl:

DavidMStraub:
Done. While writing the docs I realized it was a bad idea to call the last changed field `changed`, since it is called `change` on all other endpoints. Changed to `change` now.

PR #146, opened 2020-12-24
Add ratings parameter

cdhorn:
@DavidMStraub this one adds a "ratings" query parm to the timeline endpoints and a ratings keyword to the profile query parms and includes ratings when profile=all.  The rating information is for event information, two keywords are provided. The citations one is the number of citations for the event, and the confidence one is the highest available confidence level among the citations. 

If you have ever played with [Rootsfinder](https://www.rootsfinder.com/) Dallan added stars next to different events using one star for each citation, the idea being to provide a visual cue as to how well sourced the information was.  So the idea is that this information could be used to do something similar possibly. I included confidence as you could have 6 sources but they could all be relatively poor, and so six stars is deceiving in that scenario. So the thinking being maybe you then color code by confidence level.

### DIFF
--- a/gramps_webapi/api/resources/base.py
+++ b/gramps_webapi/api/resources/base.py
@@ -163,6 +163,7 @@ class GrampsObjectResource(GrampsObjectResourceHelper, Resource):
                         "events",
                         "age",
                         "span",
+                        "ratings",
                         "references",
                     ]
                 ),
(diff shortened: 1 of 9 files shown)


DavidMStraub:
Sorry, took me a bit longer.

Looks good to me! That might be a good way to nudge people to be more thorough about their evidence documentation.

cdhorn:
There was no rush on this one.

PR #155, opened 2020-12-27
Catch handle error in search (fixes #145); fix missing attribute in index 

### LINKED ISSUE #145: Handle `HandleError` in `SearchResouce.get_object_from_handle`
This can occur in particular when an object has been deleted from the database after the search index has been created.

DavidMStraub:
This closes #145 by catching possible handle errors in sort (e.g. when the index returns a hit to an object that has been removed from the DB after indexing).

It also fixes another bug I noticed: the `private` attribute was not properly stored in the index so #111 did not work properly.

### DIFF
--- a/gramps_webapi/api/resources/search.py
+++ b/gramps_webapi/api/resources/search.py
@@ -24,6 +24,7 @@ from typing import Dict
 
 from flask import current_app
 from gramps.gen.db.base import DbReadBase
+from gramps.gen.errors import HandleError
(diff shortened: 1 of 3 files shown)


DavidMStraub:
In c9e44a8 I added an unrelated fix just not to have to open another PR: the password reset link in the e-mail was missing the trailing slash and so didn't work.

cdhorn:
Looks good!

PR #156, opened 2020-12-28
Add dates parameter support for events and timelines

cdhorn:
@DavidMStraub I think this is ready now.  As citations and media objects are dated it operates on those in addition to events and I also added it to the timeline endpoints but with no wildcard support as it makes no sense there. If used with a timeline and the timeline has an anchor person it takes precedence with regard to setting the boundaries.

So valid arguments are things like `*/12/28` or `1900/1/*` or `1900/*/*` as suggested, but you can also do things like `-1900/1/1` for all events before that date, `1900/1/1-` for all events after that date, and `1855/1/1-1900/1/1` for all events in that range.

The apispec.yaml is updated and threw in some unit tests.


DavidMStraub:
That's absolutely brilliant! Can't wait to build a calendar view on top of this ...

PR #160, opened 2020-12-28
Remove reindex from app factory (fixes #149)

### LINKED ISSUE #149: Remove search indexer from app factory
After deploying the app I realized putting the reindexing into the app constructor was a bad idea. It should be put into a separate service function/script. (Some of the problems: the threads that are not first to index serve the app while the other thread is not done indexing yet; the reindexing is run when running the script to add a user; it is not possible to update the index without restarting the server.)

DavidMStraub:
As argued in #149 it was not such a good idea to put the search reindexing into the app factory. I removed it now and added a service script instead, which can be run e.g. like:

```
python3 -m gramps_webapi --config config.cfg search index-full
```

The next step in a separate PR will be to implement incremental indexing.

### DIFF
--- a/gramps_webapi/__main__.py
+++ b/gramps_webapi/__main__.py
@@ -20,13 +20,18 @@
 
 """Command line interface for the Gramps web API."""
 
+import logging
 import os
 
 import click
+from whoosh.index import LockError
 
 from .app import create_app
 from .auth import SQLAuth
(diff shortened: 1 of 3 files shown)


cdhorn:
Any chance this is being merged soon?  I have not used my tree for much testing as it takes so long to index...

Issue #165, opened 2020-12-29
Person timeline for people with unknown death date


DavidMStraub:
Having fun with the timeline view, I noticed that it doesn't work as (naively) expected for people that only have a birth event but no other (dated) events, but that do have offspring. I would expect to see the birth events of their children, instead there is nothing but the birth event. I could set `last=false`, but I don't want to do that in general since for people with death event using `last=true` (discarding events after death) is preferable. @cdhorn do you think we could somehow use birth + 100 years by default for people without death (or burial) event?

cdhorn:
🤣  That is such an obvious oversight on my part!  Sorry about that.  Yes of course we ought to set some reasonable bound, I'll put a PR together later tonight if I have a chance.

cdhorn:
Fixed with #166

Issue #164, opened 2020-12-29
Very slow relationship calculation


DavidMStraub:
Starting to work on a timeline view, I hit the issue that the person timeline endpoint is *very* slow on my tree. It takes more than 20 seconds to fetch my timeline (with default arguments), which makes it unusable in practice. I am pretty sure this is a problem within Gramps, not of our timeline code...

Most of the time is spent calculating the relationship between my 2 sisters and me, which each takes almost 10 seconds (the result being `Sister`). Something must be wrong there. Snakeviz shows that, for a single such relationship calculation, it instantiates more `Person` and `Family` objects than are present in the database.

Has anyone else encountered such problems?

cdhorn:
Wow!  If you pull up example_gramps, is Lewis Garner (I0044) the same time wise?

DavidMStraub:
Now, he is well below a second.

Even though I haven't investigated the code in Gramps in detail, I suspect the function derives all possible relations even if it just returns the closest one. It's quite a deep tree, up to 14 generations, over a thousand (common) ancestors.

DavidMStraub:
Oh, I think I found the solution. If I add `calculator.set_depth(2)`, the result is near-instant...

DavidMStraub:
When not set explicitly, the default value for the generation depth it iterates over is `behavior.generation-depth=15`, that's why it takes so long for my tree. Now the only question is what the appropriate setting for `depth` is depending on the settings. I guess `set_depth(depth=args["ancestors"])`? Or perhaps `set_depth(depth=args["ancestors"] + args["offspring"])`?

cdhorn:
> Oh, I think I found the solution. If I add calculator.set_depth(2), the result is near-instant...

Ah, that makes sense, yeah. 

> Or perhaps set_depth(depth=args["ancestors"] + args["offspring"])?

If I had to guess without looking at code probably not sum but larger of the two.

cdhorn:
Fixed with #166

PR #166, opened 2020-12-30
Limit relation search depth, estimate death date if needed

cdhorn:
@DavidMStraub I hope this addresses #164 and #165  I choose to use `probably_alive` for the calculation of the death date to bound with if no event indicating death is identifiable for the anchor person. For the relationship calculator the depth is set to `max(ancestors, offspring) + 1`

### DIFF
--- a/gramps_webapi/api/resources/timeline.py
+++ b/gramps_webapi/api/resources/timeline.py
@@ -28,6 +28,7 @@ from gramps.gen.display.place import PlaceDisplay
 from gramps.gen.errors import HandleError
 from gramps.gen.lib import Date, Event, EventType, Person, Span
 from gramps.gen.relationship import get_relationship_calculator
+from gramps.gen.utils.alive import probably_alive_range
 from gramps.gen.utils.db import (
     get_birth_or_fallback,
     get_death_or_fallback,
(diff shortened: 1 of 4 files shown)


cdhorn:
This second commit fixes more issues.  The age of the person the event is for is now included in the person profile, and if the timeline is not anchored to a person also returned in that age field as well. I think that age field was not returning the age of the anchor person for their own events, if so that is fixed too. Age calculation is now done in profile assembly as well.

Are you also playing with a family timeline view?

Let me know what else you find that I overlooked or need to fix.

DavidMStraub:
Great! I can confirm that it fixes both issues!

I haven't started with the family timeline view, but I suppose it will not be difficult to do it starting from the person one.

DavidMStraub:
There is just one more small issue I noticed: for one person (where the person events are not ordered chronologically for some reason), the christening/baptism events is 3 days after the birth and Gramps correctly shows the age as "3 days". However, in the timeline for some reason the birth shows an age of "3 days" and the baptism and age of "0 days". The events are correctly ordered chronologically though. Do you know what might be going on?

DavidMStraub:
Sorry, noticed another issue: for a person without a birth event (birthdate is unknown), I get

```
gramps_webapi/api/resources/timeline.py", line 461, in profile
    self.birth_dates[self.anchor_person.handle], event[0].date
KeyError
```

cdhorn:
Don't be sorry, I'm sorry you're debugging so much! Please check if this fixes those two if you don't mind.

DavidMStraub:
Yes, both issues are fixed for me, thanks! Ready to merge?

cdhorn:
Excellent!  Sure whenever you are ready.

Issue #126, opened 2020-12-16
Add caching


DavidMStraub:
Given that some endpoints take quite a long time to compute, we should add caching using `Flask-Caching`. The only tricky part is how to decide when to clear the cache so this continues working once we add more `PUT`/`POST` endpoints.

arafangion:
Hello! I wanted to point out that implementing support for using an "etag" can make a massive difference for clients, and avoids needing to figure out "how to decide when to clear the cache so this continues working".

For my own testing on a side project, having good support for the etag seems to result in massive performance gains because when the etag is up-to-date, the server can respond with: "Yep, you already have this, and it is exactly correct".

From my own reading of the relevant specs, it seems a bit ambiguous whether clients use the cache first, or the etag first, however for iOS projects I believe they will use the etag first, so a round trip to the server still happens, but when the etag is correct, the server responds with a 302 response (rather than a 200 response), and there is no response body because the server knows that the client already has the object.

I strongly recommend implementing them.

They do have to be explicitly and manually implemented, though.  I'm not sure how your server is implemented, however, this link seems relevant, though it sounds more complex than it is: https://flask-rest-api.readthedocs.io/en/stable/etag.html

Basically, the server has to explicitly mention the etag in the response headers, and check that request etag which is provided by the 'if-none-match' header when processing a response to determine if it can short-circuit the response and straight away return the 302 status. 

As an example, this is how I test for the etag manually:

`curl -H "if-none-match: \"hey\"" http://192.168.1.2:3000/person`

There, I'm asking the server, running on port 3000, 'Please give me the data for /person, unless the etag matches "hey"', and if the computed etag for the response would have been "hey", then the server responds with a 302 response and no data.

As a result I know that the object I have (which is typically in my cache) is in fact the most up-to-date object, which means the server does not have to upload the data, the client does not have to download the data, and does not have to parse the data.  It's just all round substantially faster.

DavidMStraub:
Thanks for the suggestion! We're actually already using `ETag` for the single-object endpoints; the motivation was not really caching but to avoid "mid-air collisions" when editing an object; see e.g. the discussion here https://github.com/gramps-project/gramps-webapi/pull/196#issuecomment-889376162.

In principle, we can/should add ETags to more endpoints to facilitate better caching.

[...]

PR #170, opened 2021-01-03
Implement thumbnail caching (#126)

DavidMStraub:
In my deployment I am already in trouble with high CPU usage, most likely due to constant thumbnail regeneration. This PR addresses this issue.

I didn't use naive caching by query string as in the "old" backend since I realized the following problems:

- It also uses the JWT as key which changes every 15 minutes
- It would serve cached private media files to unauthorized users (if they know/guess the handle)
- It wouldn't detect media objects with modified file, but unchanged handle

To solve all three problems, I now use a custom cache key which uses the MD5 hash from the Gramps DB, the request path *and* the hashed query arguments but *without* the JWT. By using the MD5 hash from the DB, we can abort for unauthorized users (since the private proxy DB will not contain the handle) and at the same time not serve outdated files.


DavidMStraub:
If there are no objections, I'd like to merge this later today, to take some load off my server.

cdhorn:
Sounds good to me.

Issue #168, opened 2021-01-01
Missing MIME type


DavidMStraub:
On my system, when trying to generate a Graphviz report (with Graphviz installed), I get

```
  File "gramps_webapi/api/resources/reports.py", line 212, in get
    return send_file(buffer, mimetype=types_map[file_type])
KeyError: '.gv'
```

If I try to set the mimetype to `application/octet-stream`, I get

```
File "lib/python3.8/site-packages/flask/helpers.py", line 592, in send_file
    raise ValueError(
ValueError: Unable to infer MIME-type because no filename is available. Please set either `attachment_filename`, pass a filepath to `filename_or_fp` or set your own MIME-type via `mimetype`.
```

Not sure what's the best way to deal with this; possibly setting `appliction/octet-stream` *plus* setting an explicit filename in case the MIME type is not recognized.

cdhorn:
A couple things at play here...

First, most of this is probably a lack of knowledge on my part on proper mime handling and I made a few mistakes.  I don't know `Graphviz` and in testing I ran into:
```
>>> types_map[".dot"]
'application/msword'
```
That clearly is not a `Graphviz` file and some searches all kept turning up Microsoft Word.  So at the time I foolishly assumed that  `.gv` and `.dot` were interchangeable so remapped the one type to the other which I now realize they are not. So I need to fix that.

I also had trouble with `.gvpdf` and `.gspdf` as well as they were not in `types_map` on my system at all so I remapped them to `.pdf` and I now wonder if that was incorrect to do as well.

Now, oddly, I can't seem to reproduce your error on my system using the API but I can from a normal Python shell which does not seem to make sense. The `/usr/share/mime/text/vnd.graphviz.xml` is certainly there and I don't see why it would be recognized in one context and not the other.  And having found that I see the `.dot` mapping to `application/msword` must take precedence over the other file.

It may make sense to switch back to keeping our own dictionary of types Gramps outputs for lookups like I was originally doing. I know you said that was redundant with `types_map` and it seems it should be hence I changed it, but that would at least present consistent behaviour.  Any thoughts on why I'm seeing inconsistent behaviour?

DavidMStraub:
Yeah it's confusing, I don't understand it fully either. I saw there is [mimetypes.init()](https://docs.python.org/3/library/mimetypes.html#mimetypes.init) which might add some additional types (so calling this or not might lead to different behaviour), but I don't know if this has anything to do with the inconsistencies you're seeing.

[...]

PR #171, opened 2021-01-03
Initialize mimetypes first (For #168)

cdhorn:
@DavidMStraub let's see if adding mimetypes.init() fixes it on your end.

### DIFF
--- a/gramps_webapi/api/resources/exporters.py
+++ b/gramps_webapi/api/resources/exporters.py
@@ -23,9 +23,9 @@
 
 """Exporters Plugin API resource."""
 
+import mimetypes
 import os
 import uuid
-from mimetypes import types_map
 from pathlib import Path
 from typing import Dict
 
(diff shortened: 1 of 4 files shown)


DavidMStraub:
Now I get this:

```
  File "gramps_webapi/api/resources/reports.py", line 118, in run_report
    "Can not find {} in mimetypes.types_map" % file_type
TypeError: not all arguments converted during string formatting
```

I think you should use `" ... {} ... ".format(...)`.

cdhorn:
Yeah but if it hit that the mimetypes.init() clearly didn't resolve the issue on your system...  

Is this running in the container or externally or both?

DavidMStraub:
This was just running directly on my machine, no container.

cdhorn:
This shouldn't be related, but did it leave any work files in the Gramps temp directory?  Can you check and let me know and if so if they are .dot or .dot.gv as perhaps that was a bug later fixed and I need to special case it. That thought just occurred to me. 

I think I added REPORT_PATH and EXPORT_PATH to override that behaviour BTW.

I think you run Ubuntu right?  Focal?  Can you shoot me level and apt list --installed output and the package list from the virtual environment you're running it in?  I want to try to reproduce it.  I know we could hard code and be done with it but I'd much rather figure out why now...

DavidMStraub:
I looked into it a bit ... apparently `mimetypes` just looks into `/etc/mime.types` on my system, which does not contain an entry for Graphviz. `xdg-mime` apparently uses data in `/usr/share/mime`, which *does* include info about Graphviz.

In the end it really seems the easiest (and must robust) way would be to hardcode all types needed for Gramps. Sorry for originally suggesting `mimetypes`, it is less useful than I thought ...

cdhorn:
No need to apologize, it seems we both learned something new.  Learn something new every day!

PR #173, opened 2021-02-06
Change license to AGPL

DavidMStraub:
@Nick-Hall & @cdhorn: as discussed, it makes sense to relicense the code under AGPL. I wasn't sure whether this is possible since we reuse GPL-licensed code from Gramps. I now read up on this and found that GPLv3 explicitly allows this:

> Notwithstanding any other provision of this License, you have permission to link or combine any covered work with a work licensed under version 3 of the GNU Affero General Public License into a single combined work, and to convey the resulting work. The terms of this License will continue to apply to the part which is the covered work, but the special requirements of the GNU Affero General Public License, section 13, concerning interaction through a network will apply to the combination as such.

Source: https://www.gnu.org/licenses/gpl-3.0.de.html

While Gramps is licensed under GPLv2, not v3, in all its license headers it explicitly contains the statement 

> This program is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation; either version 2 of the License, or (at your option) **any later version**.

(emphasis mine). So this means that it's perfectly fine to relicense any code that uses parts of Gramps as GPLv3 or AGPLv3.

By the way, the AGPL contains an analogous statement that AGPL-licensed code can be used in GPL-licensed projects, and since GPLv2 is (slightly) more permissive than GPLv3, this means it will always be possible to reuse code from *this* project in Gramps itself. This is important, as otherwise new features developed here would be blocked from being ported to Gramps, which would be bad.

I've marked this PR as draft since it's still missing the necessary replacements in the license headers; I was too lazy to do this before getting both of your :+1:s

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,281 +1,620 @@
-                    GNU GENERAL PUBLIC LICENSE
-                       Version 2, June 1991
+                    GNU AFFERO GENERAL PUBLIC LICENSE
+                       Version 3, 19 November 2007
 
- Copyright (C) 1989, 1991 Free Software Foundation, Inc.,
- 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
  Everyone is permitted to copy and distribute verbatim copies
  of this license document, but changing it is not allowed.
 
                             Preamble
 
-  The licenses for most software are designed to take away your
-freedom to share and change it.  By contrast, the GNU General Public
-License is intended to guarantee your freedom to share and change free
-software--to make sure the software is free for all its users.  This
-General Public License applies to most of the Free Software
-Foundation's software and to any other program whose authors commit to
-using it.  (Some other Free Software Foundation software is covered by
-the GNU Lesser General Public License instead.)  You can apply it to
-your programs, too.
+  The GNU Affero General Public License is a free, copyleft license for
+software and other kinds of works, specifically designed to ensure
+cooperation with the community in the case of network server software.
+
+  The licenses for most software and other practical works are designed
+to take away your freedom to share and change the works.  By contrast,
+our General Public Licenses are intended to guarantee your freedom to
+share and change all versions of a program--to make sure it remains free
+software for all its users.
 
   When we speak of free software, we are referring to freedom, not
 price.  Our General Public Licenses are designed to make sure that you
 have the freedom to distribute copies of free software (and charge for
-this service if you wish), that you receive source code or can get it
-if you want it, that you can change the software or use pieces of it
-in new free programs; and that you know you can do these things.
+them if you wish), that you receive source code or can get it if you
+want it, that you can change the software or use pieces of it in new
+free programs, and that you know you can do these things.
 
-  To protect your rights, we need to make restrictions that forbid
-anyone to deny you these rights or to ask you to surrender the rights.
-These restrictions translate to certain responsibilities for you if you
-distribute copies of the software, or if you modify it.
+  Developers that use our General Public Licenses protect your rights
+with two steps: (1) assert copyright on the software, and (2) offer
+you this License which gives you legal permission to copy, distribute
+and/or modify the software.
 
-  For example, if you distribute copies of such a program, whether
-gratis or for a fee, you must give the recipients all the rights that
-you have.  You must make sure that they, too, receive or can get the
-source code.  And you must show them these terms so they know their
-rights.
+  A secondary benefit of defending all users' freedom is that
+improvements made in alternate versions of the program, if they
+receive widespread use, become available for other developers to
+incorporate.  Many developers of free software are heartened and
+encouraged by the resulting cooperation.  However, in the case of
+software used on network servers, this result may fail to come about.
+The GNU General Public License permits making a modified version and
+letting the public access it on a server without ever releasing its
+source code to the public.
 
-  We protect your rights with two steps: (1) copyright the software, and
-(2) offer you this license which gives you legal permission to copy,
-distribute and/or modify the software.
+  The GNU Affero General Public License is designed specifically to
+ensure that, in such cases, the modified source code becomes available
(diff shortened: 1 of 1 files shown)


cdhorn:
I am glad you looked into this, I am certainly all for it.

Nick-Hall:
Gramps uses the GPLv2+ license.  So we can use Gramps core code in a AGPLv3 licensed project.

Search for "FOSS license compatibility" for details.  See this [floss license slide](https://dwheeler.com/essays/floss-license-slide.html) as an example.

However, I don't think that the reverse is true.

We should probably choose AGPLv3 for this project though, and more common code into core Gramps now.

DavidMStraub:
> However, I don't think that the reverse is true.

The [AGPL](https://www.gnu.org/licenses/agpl-3.0.html) contains the clause

> Notwithstanding any other provision of this License, you have permission to link or combine any covered work with a work licensed under version 3 of the GNU General Public License into a single combined work, and to convey the resulting work. The terms of this License will continue to apply to the part which is the covered work, but the work with which it is combined will remain governed by version 3 of the GNU General Public License.

So I think it's fine.

Nick-Hall:
The problem is that the combined work becomes AGPLv3, which isn't what we want.

I've just been looking through our mailing list archive.  This was first discussed in the [licence thread](https://sourceforge.net/p/gramps/mailman/message/24049990/) for _gramps-connect_ in 2009.

Don Allingham added his thoughts in [this thread](https://sourceforge.net/p/gramps/mailman/message/24050327/).  So I think that AGPLv3 is the way to go for the _gramps-webapi_ project.

Another [relevant thread](https://sourceforge.net/p/gramps/mailman/message/34726900/) discusses licence compatibility issues.  Here a user wanted to write a plugin that depended on LGPLv3+ licensed code.  Gary emailed FSF regarding the issue.  They recommended resolving licence issues before conveying the software to the user.

DavidMStraub:
I agree we should sort this out before making a release. But I think I disagree with

> The problem is that the combined work becomes AGPLv3, which isn't what we want.

because the AGPLv3 clearly states

>  the work with which it is combined will remain governed by version 3 of the GNU General Public License.

Strictly speaking, AGPL would apply to code snippets copied from `gramps-webapi` to `gramps` (which remains GPLv3), but in practice I think that doesn't have any effect, since the only addition in AGPL is about having to disclose the source code when the software is accessed remotely through a server.

Nick-Hall:
The work with which it is combined (Gramps core) would be GPLv3, but the combined work would be AGPLv3. This is clarified slightly in the GPLv3 section 13.  Gramps core is GPLv2+ rather than GPLv3 anyway.

Therefore I don't think that we could accept AGPLv3 code in Gramps core.  Maybe @bmatherly has more knowledge of this subject.  He was more involved in the 2009 discussions.

DavidMStraub:
> The work with which it is combined (Gramps core) would be GPLv3, but the combined work would be AGPLv3. This is clarified slightly in the GPLv3 section 13.

I see, you're right.

So should we leave it at GPLv3 to remain fully compatible?

cdhorn:
> So should we leave it at GPLv3 to remain fully compatible?

We're using GPLv2 at the moment to match Gramps, not GPLv3.

Glancing over things again quick and maybe I am second guessing myself now but not sure what would make sense to add to Gramps core.

Nick could you look at the Timeline class in api/resources/timeline.c and let me know if you think something like that would be a good addition to add as is, or do you see that kind of functionality being provided differently?

The option class I used to get the exporters working I'm not sure would be used as is, you mentioned hoping to use the same setup used for the reports for that at some point perhaps. Some of the other stuff, outside of the search engine, is sort of simple like adding some extra sort key and utility methods. 

Maybe it doesn't make sense to worry about all that much and just switch the license over now.

David I can probably make some time to quick throw rest of PR together to change the headers if you want.

bmatherly:
>Therefore I don't think that we could accept AGPLv3 code in Gramps core. Maybe @bmatherly has more knowledge of this subject. He was more involved in the 2009 discussions.

Here is my understanding...

A source file can have a copyright claim at the top. All of the files in Gramps core are currently GPLv2+ (unless some GPL3 slipped in that I am unaware of). Therefore, the combined work has a copyright of GPLv2+.

If we add a single file to Gramps core that is AGPL, then all the other source files remain at GPLv2+. However, the combined work becomes AGPL.

We can do this, and it is legal, but it is difficult to go back (would require removing AGPL files). So effectively everything becomes AGPL for all of time. One drop contaminates the whole batch.

Historically, when this topic has come up in the past, we eventually gave up on it because it seemed complicated and the benefit didn't outweigh the complication.

Speaking personally, when I donate software, I do it in order to benefit as many people as possible. I do not care if someone finds a way to make money using if that means it is useful for more people. We already have people selling Gramps on Ebay and all kinds of other strange things. I do not care to concern myself with trying to avoid those situations. I would be happy for all my code contributions to be released in the public domain or released under the MIT license. But I do not feel strongly about these ideas and I am happy to follow the majority of the Gramps contributors.

Nick-Hall:
@DavidMStraub Why was this closed?  Did you make a decision?

We can change the licence to AGPL3+ if both contributors agree.

The only disadvantage would be for code that we may wish to move into Gramps core in the future.  As far as I can see there would be three approaches:

1. Change the licence of an individual module by agreement.  This is not a problem when there are only two of you, but may need to be reviewed in the future.
2. Have a mix of licences in the _gramps-webapi_.  Attach a GPL2+ licence that we may wish to move.
3. Write any general code in Gramps core.  We could fast-track the merge process when adding new functionality.

DavidMStraub:
> @DavidMStraub Why was this closed? Did you make a decision?

I had assumed it was too complicated given the above discussion, but I'm perfectly fine with using a mix of the 3 approaches you suggest (e.g. we can use approach 1 for smaller code snippets, while for larger features we develop from scratch we can go for approach 2 or 3 case by case).

So I'll go ahead and replace the license headers as well before merging this.

DavidMStraub:
Closed in favour of #179.

PR #177, opened 2021-06-28
Implement API changes of Flask-JWT-Extended 2.0 (fixes #175)

### LINKED ISSUE #175: Make compatible with Flask JWT extended v4
A couple of weeks ago, a new major version of Flask JWT extended was released that contains several breaking changes; I noticed this since it broke the Travis jobs.

https://flask-jwt-extended.readthedocs.io/en/stable/v4_upgrade_guide/

We have to sort this out before making a release (#158).

DavidMStraub:
This should fix #175.

Note an important announcement in the Flask JWT Extended [upgrade guide](https://flask-jwt-extended.readthedocs.io/en/stable/v4_upgrade_guide/):

> This has the unfortunate side effect that any existing JWTs your application is using will not work correctly if they utilize additional claims. We strongly suggest changing your secret key to force all users to get the new format of JWTs. If that is not feasible for your appilication you could build a shim to handle both the old JWTs which store additional claims in the user_claims key, and the new format where additional claims are now stored at the top level, until all the JWTs have had a chance to cycle to the new format.

### DIFF
--- a/gramps_webapi/api/auth.py
+++ b/gramps_webapi/api/auth.py
@@ -24,11 +24,7 @@ from functools import wraps
 from typing import Iterable
 
 from flask import abort, current_app
-from flask_jwt_extended import (
-    get_jwt_claims,
-    verify_jwt_in_request,
-    verify_jwt_refresh_token_in_request,
-)
+from flask_jwt_extended import get_jwt, verify_jwt_in_request
 
 
 def jwt_required_ifauth(func):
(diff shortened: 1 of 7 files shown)


DavidMStraub:
Can't find the reason for the failing tests :thinking:

DavidMStraub:
Until earlier this year, using the Gramps version from the `master` branch, the following worked:

```python
from gramps_webapi.api.util import get_locale_for_language
locale = get_locale_for_language("de")
locale.translation.sgettext("0 days")
```
which would return "0 Tage". With the current Gramps, this does not work anymore and returns "0 days" and I can't find the reason. `pyICU` is correctly installed (both on my machine and here on Travis). Perhaps some changes in Gramps or in `pyICU`? Any clue, @Nick-Hall?

DavidMStraub:
Figured it out. We should use

```python
locale.translation.ngettext("{number_of} day", "{number_of} days", 0).format(number_of=0)
```

DavidMStraub:
Can we call this a Zero-Day vulnerability? :upside_down_face:

PR #179, opened 2021-06-28
Update headers and LICENSE to AGPL for most objects

cdhorn:
As discussed this should update all the file headers and the licensing to AGPL with the exception of the following modules which will remain GPL as they include snippets from or derive from code in Gramps proper: dbloader.py, dbmanager.py, exporters.py, sort.py, and html.py  In addition I am leaving the timeline.py GPL as I am using a descendant of it elsewhere now.


DavidMStraub:
Looks good to me! The failing tests are due to #175 and can be safely ignored as you didn't change any code.

DavidMStraub:
Rebased to master

PR #196, opened 2021-07-20
Endpoints for updating and deleting objects (fixes #193, fixes #191)

### LINKED ISSUE #191: Add endpoints for updating objects
This is just a `PUT` version of the `POST` in #178, which should be a trivial generalization apart from the issue of handling concurrent changes. I think we can fix it with [optimistic locking](https://medium.com/swlh/handling-concurrent-requests-in-a-restful-api-5a25a4b81a1) using `ETag` and `If-Match`. My first thought was to use `change` as an `ETag`, but I think it's better to use a hash of the Gramps object's JSON representation, since `change` only has 1 s granularity and it could happen that two modifications happen within one second. Hash collisions are impossible as the JSON representation always includes the handle, which is unique.

We could also add `PATCH` to change individual properties.

### LINKED ISSUE #193: Add endpoints for deleting objects
This should be fairly simple. 

One questions is if we want to provide a way of undoing the delete. We cannot really do it in the backend as each request has its own database session, but we could return the JSON representation of the deleted object in the response, so a client could choose to implement an "UNDO" button that posts this object back to recreate it (which is not an "undo" from the backend perspective, but has the same effect).

A more basic question regarding the implementation of deletions in Gramps: what if I delete an event, what happens to all the event references of people and families? Looking at the DBAPI code, it looks like the `reference` table in the DB is automatically updated to remove all those references/backlinks, however the objects themselves are not updated, are they? So they still contain those "dead" references. Is there some point where Gramps cleans up those dangling references?

DavidMStraub:
This is a draft implementation of the `DELETE` endpoints. The API is very simple: just call `DELETE` on `/api/<class>/<handle>`. It will fail if the user doesn't have `DeleteObject` permission.

To enable applications to implement undo-like functionality, the return value of the call is a JSON representation of the database transaction of the form

```json
[
  {
    "_class": "<class>",
    "type": "delete",
    "handle": "<handle>",
    "old": "<JSON representation of object>",
    "new": null
  }
]
```

For the deletion of references, I needed to implement lots of functions by copying over code from classes like `DeleteEventQuery`. These generic functions in the new `delete` module could be moved to the `DbGeneric` class in Gramps at some point.


DavidMStraub:
Forgot to mention: this is rebased onto the branch in #188, so the diff is currently not super useful until this is merged.

DavidMStraub:
TODO

- [ ] For `ETag`, might want to use the same hash regardless of `GET` query parameters
- [ ] For `POST`, should return the newly created handle, e.g. in a `Location` header

DavidMStraub:
Docs are still missing.

DavidMStraub:
Done!

The API is simple enough:

- delete: `DELETE` to `/api/<class>/<handle>`
- update: `PUT` the updated JSON `{<object>}` to `/api/<class>/<handle>`

In both cases, upon success a JSONified database transaction is returned as reponse

- delete: `{"type": "delete", "_class": "<class>", "handle": "<handle>", "old": "{<object>}", "new": null}`
- update: `{"type": "update", "_class": "<class>", "handle": "<handle>", "old": "{<object>}", "new": "{<updatedobject>}"}`

This can be used by applications to implement undo-like functionality.

For `PUT`, I also implemented `ETag` as mentioned in #191:

Every `GET` to `/api/<class>/<handle>` now sets an `ETag` header that  is based on a hash of the object's JSON representation. This can be sent by applications along with the updated object in a `PUT`'s `If-Match` header. If the object has changed in the meantime, the request will fail with a 416. (The application could then offer overwriting the changes by `PUT`ing again without that header.)

cdhorn:
I wonder if ETag should also be implemented on DELETE. While not likely, what if two users get the object, and then one of them updates it and then the other deletes it?  Should the one deleting it have been made aware that it was modified, as perhaps the information from that update might have caused them to choose otherwise?

DavidMStraub:
Good point, yes!

DavidMStraub:
Let me know if you think this can be merged.

gramps_webapi/api/resources/base.py
```diff
@@ -139,6 +144,11 @@ def _parse_object(self) -> GrampsObject:
             abort(400)
         return from_json(json.dumps(obj_dict))
 
+    def has_handle(self, handle: str) -> bool:
+        """Chek if the handle exists in the database."""
```
cdhorn: Spelling

gramps_webapi/api/resources/delete.py
```diff
@@ -0,0 +1,414 @@
+"""Functions for deleting objects with references."""
```
cdhorn, line 28: Missing license preamble and copyrights if sections pulled from Gramps core plus your own.

gramps_webapi/api/resources/delete.py
```diff
@@ -0,0 +1,414 @@
...
+def delete_object(
+    db_handle: DbWriteBase, handle: str, gramps_class_name: str
+) -> List[Dict[str, Any]]:
+    """Delete the object and its references."""
+    key = gramps_class_name.lower()
+    try:
+        method = delete_methods[key]
+    except KeyError:
+        raise NotImplementedError(gramps_class_name)
+    with DbTxn(f"Delete {gramps_class_name} {handle}", db_handle) as trans:
```
cdhorn, line 438: Not sure how undo would work with things, but if like desktop Gramps the transaction message would be displayed to the user to describe the action being undone I would argue it should reference the Gramps Id and not the handle.

gramps_webapi/api/resources/util.py
```diff
@@ -807,3 +811,35 @@ def update_object(
...
+            new_data = obj_cls().unserialize(new_data)
+        item = {
+            "type": trans_dict[action],
+            "handle": handle,
+            "_class": obj_cls_name,
+            "old": json.loads(to_json(old_data)),
+            "new": json.loads(to_json(new_data)),
+        }
+        out.append(item)
+    return out
```
cdhorn, line 839: Should the message describing the action performed by the transaction also be returned for undo history functionality?

gramps_webapi/auth/const.py
```diff
@@ -56,6 +56,7 @@
         PERM_ADD_OBJ,
         PERM_DEL_OBJ,
     },
+    ROLE_CONTRIBUTOR: {PERM_EDIT_OWN_USER, PERM_VIEW_PRIVATE, PERM_ADD_OBJ,},
```
cdhorn, line 59: Maybe it seems too fine grained, but it seems natural to me you would want two versions of `ROLE_CONTRIBUTOR` and `ROLE_EDITOR`, one that is _trusted_ with `PERM_VIEW_PRIVATE` and one that is not. 

The one that is not would be for those with whom you may be working on older generations but who you are not close enough to that you would be willing to share everything.

tests/test_endpoints/test_delete.py
```diff
@@ -0,0 +1,161 @@
...
+        cls.dbman.create_new_db_cli(cls.name, dbid="sqlite")
+        with patch.dict("os.environ", {ENV_CONFIG_FILE: TEST_AUTH_CONFIG}):
+            cls.app = create_app()
+        cls.app.config["TESTING"] = True
+        cls.client = cls.app.test_client()
+        sqlauth = cls.app.config["AUTH_PROVIDER"]
+        sqlauth.create_table()
+        sqlauth.add_user(name="user", password="123", role=ROLE_GUEST)
+        sqlauth.add_user(name="contributor", password="123", role=ROLE_CONTRIBUTOR)
+        sqlauth.add_user(name="admin", password="123", role=ROLE_OWNER)
```
cdhorn, line 67: From a QA standpoint should `ROLE_MEMBER` and `ROLE_EDITOR` not also be validated?

tests/test_endpoints/test_put.py
```diff
@@ -0,0 +1,146 @@
...
+        cls.dbman.create_new_db_cli(cls.name, dbid="sqlite")
+        with patch.dict("os.environ", {ENV_CONFIG_FILE: TEST_AUTH_CONFIG}):
+            cls.app = create_app()
+        cls.app.config["TESTING"] = True
+        cls.client = cls.app.test_client()
+        sqlauth = cls.app.config["AUTH_PROVIDER"]
+        sqlauth.create_table()
+        sqlauth.add_user(name="user", password="123", role=ROLE_GUEST)
+        sqlauth.add_user(name="contributor", password="123", role=ROLE_CONTRIBUTOR)
+        sqlauth.add_user(name="admin", password="123", role=ROLE_OWNER)
```
cdhorn, line 68: Same question as with delete about validating the other roles.

cdhorn (changes requested):
The missing license/copyright on delete.py is the only must fix issue I think.

gramps_webapi/api/resources/delete.py
DavidMStraub: Right! I will add the GPL header and the various copyrights from the files defining the `Delete...View` classes in Gramps, because that's where most of the code comes from.

gramps_webapi/api/resources/delete.py
DavidMStraub: This message won't be exposed to the API user at any time, so if at all, it might appear in a debug log; in that case, the handle might even be better as it's guaranteed to be unique.

gramps_webapi/api/resources/util.py
DavidMStraub: I think the JSON itself contains all the information: e.g. `[{"type": "delete", "handle": "somehandle", ...}, ...]`

gramps_webapi/auth/const.py
DavidMStraub: I thought about that, the reason I didn't add it (so far) was that granting `AddObject` without granting `ViewPrivate` leads to some additional complications that have to be handled. For instance, should we allow such a user to create on object that is itself private? Then they would have no way to view this object after creation.

Not saying it's impossible, but I'd suggest opening an issue and tackling this separately.

DavidMStraub:
Thanks for the review! Made some changes.

gramps_webapi/api/resources/delete.py
cdhorn: Okay

gramps_webapi/auth/const.py
cdhorn: Yeah that makes sense to handle that separately as you say. It would be nice if a `ViewLiving` was available separately too, but I seem to recall there were issues with that so you decided it made more sense to mark all living people private.

cdhorn:
Looks good to merge to me!

gramps_webapi/auth/const.py
DavidMStraub: Yes, we discussed this issue [in the forum](https://gramps.discourse.group/t/gramps-web-api-first-release/1609/5?u=davidmstraub). Still not sure how to solve it.

PR #197, opened 2021-07-30
Implement file upload (fixes #189)

### LINKED ISSUE #189: Add endpoint for file upload
When we want to add media, we don't only need the media object creation endpoint (#178) but also a way to upload files. Looking at [this stackoverflow discussion](https://stackoverflow.com/questions/3938569/how-do-i-upload-a-file-with-metadata-using-a-rest-web-service), I agree it would be best to first upload the metadata (the media object, in our case) and then allow uploading the file. We can simply do that with a `/file/<handle>/` `PUT` endpoint. The file backend can then decide what to do: either store it on the local disk, or, if the backend is something fancier, like S3, directly redirect there.

However, I see some tricky security issues to avoid here: I think we cannot offer a "file dialog" to choose a subdirectory because we have to assume the directory and file names should not be accessible to all users with upload rights (the users might not have permission to view private records, and some of the files/directories might contain private information in their names). Also, what to do about file name collisions?

The cleanest solution would be to use handles as file names and a standard directory name, e.g. `$MEDIA_BASE_DIR/GrampsWebApi/<handle>.<ext>`. This could then be cleaned/reorganized by the site owner locally with Gramps Desktop, if necessary.

DavidMStraub:
API:

- `POST` of a binary stream to `/api/media/` creates a new, empty media object, stores the file in the media base directory as `<checksum>.<ext>` where `ext` is derived from the MIME type, and returns a transaction JSON (which contains, for instance, the handle of the newly created object). Note: it is no longer possible to `POST` a JSON to the media endpoint as that would create a media object without file!
- `PUT` of a JSON to `/api/media/<handle>` updates the media object (but not its file!)
- `PUT` of a binary stream to `/api/media/<handle>/file` uploads a new file and updates the `path`, `mime`, and `checksum` of the media object accordingly

This means all files uploaded via the web API will be stored as `$MEDIA_BASE_DIRECTORY/<checksum>.<ext>` and the user cannot influence the file name or location. I believe this is necessary as it solves the problems mentioned in #189. It also allows for deduplication as the files are always stored under their MD5 checksum.

The only possible issue I see is that there is no check whether the MIME type supplied by the user is correct or not. However I believe this is acceptable - we cannot really guard against malicious or corrupted files uploaded by authorized users. We do have to make sure that *unauthorized* users cannot access any private information, and I believe this is ensured by this implementation.

Docs still missing.

Edit: note this is rebased on top of #196, so the diff is not useful.

### DIFF
--- a/gramps_webapi/api/resources/file.py
+++ b/gramps_webapi/api/resources/file.py
@@ -20,9 +20,14 @@ class MediaFileResource(ProtectedResource):
 
     def get(self, handle) -> Response:
         """Download a file."""
+        db_handle = get_db_handle()
+        try:
+            obj = db_handle.get_media_from_handle(handle)
+        except HandleError:
+            abort(HTTPStatus.NOT_FOUND)
         base_dir = current_app.config.get("MEDIA_BASE_DIR")
         handler = LocalFileHandler(handle, base_dir)
-        return handler.send_file()
+        return handler.send_file(etag=obj.checksum)
 
     def put(self, handle) -> Response:
         """Upload a file and update the media object."""
@@ -32,6 +37,10 @@ class MediaFileResource(ProtectedResource):
             obj = db_handle.get_media_from_handle(handle)
         except HandleError:
             abort(HTTPStatus.NOT_FOUND)
+        checksum_old = obj.checksum
+        for etag in request.if_match:
+            if etag != checksum_old:
+                abort(412)
         mime = request.content_type
         if not mime:
             abort(HTTPStatus.NOT_ACCEPTABLE)
--- a/gramps_webapi/api/resources/file.py
+++ b/gramps_webapi/api/resources/file.py
@@ -1,0 +1,49 @@
+"""Endpoint for up- and downloading media files."""
+
+from http import HTTPStatus
+
+from flask import Response, abort, current_app, request
+from gramps.gen.db import DbTxn
+from gramps.gen.errors import HandleError
+
+from ...auth.const import PERM_EDIT_OBJ
+from ..auth import require_permissions
+from ..file import LocalFileHandler, process_file, upload_file
+from ..util import get_db_handle
+from . import ProtectedResource
+from .util import update_object
+
+
+class MediaFileResource(ProtectedResource):
+    """Resource for media files."""
+
+    def get(self, handle) -> Response:
+        """Download a file."""
+        base_dir = current_app.config.get("MEDIA_BASE_DIR")
+        handler = LocalFileHandler(handle, base_dir)
+        return handler.send_file()
+
+    def put(self, handle) -> Response:
+        """Upload a file and update the media object."""
+        require_permissions([PERM_EDIT_OBJ])
+        db_handle = get_db_handle()
+        try:
+            obj = db_handle.get_media_from_handle()
+        except HandleError:
+            abort(HTTPStatus.NOT_FOUND)
+        checksum, f = process_file(request.stream)
+        base_dir = current_app.config.get("MEDIA_BASE_DIR")
+        mime = request.content_type
+        if not mime:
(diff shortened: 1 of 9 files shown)


DavidMStraub:
Ready for review; also added the files's checksum as `ETag` for `PUT`.

gramps_webapi/api/resources/file.py
```diff
@@ -0,0 +1,66 @@
+"""Endpoint for up- and downloading media files."""
```
cdhorn, line 21: License preamble and your copyright.

gramps_webapi/api/resources/file.py
```diff
@@ -0,0 +1,66 @@
...
+        with DbTxn("Update media object", db_handle_writable) as trans:
+            update_object(db_handle_writable, obj, trans)
+            try:
+                pass  # update_object(db_handle_writable, obj, trans)
+            except ValueError:
+                abort(400)
+            trans_dict = transaction_to_json(trans)
+        return Response(
+            response=json.dumps(trans_dict), status=200, mimetype="application/json"
+        )
```
cdhorn, line 86: So unless I overlooked it the previous file is left orphaned and is not cleaned up, since the checksum will have changed and a new file will be written as a result. Which you would want if you had undo functionality I suppose, but there should probably be a means of identifying and purging them as well. (I suppose they are orphaned on a delete as well..)

cdhorn (commented):
Seems to look okay to me other than the missing license again. I need to make some time to play with the new features, you are on a roll!

gramps_webapi/api/resources/file.py
DavidMStraub: That's true, currently there is no way to delete files and there is no automatic cleanup. There is a possibility to delete *media objects* but that does not remove the corresponding file.

This is similar to how it works in Gramps: deleting a Gramps media object does not delete the corresponding file. Gramps actually never deletes files.

This is also why an automatic cleanup cannot really work: just because there is a file in the media base dir that does not have a corresponding media object doesn't mean we can delete it (I have plenty of files in my media directory that do not have associated objects).

There are two problems with this:

1. The media directory might keep filling up with outdated/unused files
2. An inadvertent upload of a file (e.g. a user uploads an image and then realizes the image is copyrighted or otherwise private, and tries to delete it again) cannot be undone

I think 1. has to be solved manually by the administrator, I don't see any other way. I don't see a way to solve 2. at all, so I guess users have to be careful.

cdhorn: For 1 you can scan the media dir and cross reference against the media objects, so you could identify orphans to provide a report or allow the admin to schedule an automated cleanup of say anything orphaned and more than 30 days old.  

I see it as different from desktop Gramps where you have control over the file name and location and you are the one adding them and deleting them.

For 2 you would need to add Contributor to the Gramps data model and track physical file contributions separate of the media object, or add such tracking to the user database on the API side.

Not something needed now but to keep in mind I guess. As I understand it I think 2 must be addressed for GDPR.

DavidMStraub: > As I understand it I think 2 must be addressed for GDPR.

I think the possibility to delete content (directly) and the right for deletion of personal data are separate - even if the user cannot delete a file they can ask the tree owner to delete anything that contains their personal data.

> For 1 you can scan the media dir and cross reference against the media objects, so you could identify orphans to provide a report or allow the admin to schedule an automated cleanup of say anything orphaned and more than 30 days old.

That sounds like a great idea for a Gramps (desktop) addon. There are already several cleanup tools for other object types. The admin could use that regularly with their synchronised database & media directory.

Btw, I also want to re-implement the S3 file handler I had in the "old" app at some point - in that case, only files with a corresponding media object will be stored on S3 with the checksum as handle, and I don't see a problem with deleting them.

PR #200, opened 2021-08-04
Update search index on PUT, POST, & DELETE (fixes #199)

### LINKED ISSUE #199: Update search index on PUT, POST, and DELETE
We can't rely on server admins doing a regular reindex as users will expect changes to be reflected immediately in functionalities that rely on search.

DavidMStraub:
After implementing this (which just required an update/add/delete call to the index in the put/post/delete methods), I realized this can lead to concurrent requests failing as the first one is obtaining an exclusive write lock on the search index. Fortunately, `whoosh` provides an `AsyncWriter` class that automatically handles the lock error by spawning a thread in which it retries commiting the changes to the index. I added a unit test where I add 10 objects while the index is manually locked; they all become searchable after a short wait.

### DIFF
--- a/gramps_webapi/api/resources/base.py
+++ b/gramps_webapi/api/resources/base.py
@@ -412,7 +412,7 @@ class GrampsObjectsResource(GrampsObjectResourceHelper, Resource):
         # get the new handle from the transaction dict, in case it was not specified
         # explicitly
         handle = trans_dict[0]["handle"]
-        with indexer.index(overwrite=False).writer() as writer:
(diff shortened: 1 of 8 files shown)


DavidMStraub:
Tests finally fixed ... this PR would be the next to merge (i.e. before #201).

DavidMStraub:
Please let me know if there are any objections to merging this. I'd like to go ahead as I'm starting to implement the object addition UI in Gramps.js.

cdhorn (approved):
Quick looked over, looks good to me!

Issue #163, opened 2020-12-29
Gramps synchronization


DavidMStraub:
Eventually, especially once we implement write operations, it would be great to support two-way synchronization with Gramps desktop. I suppose it would not be too hard to implement a plugin that uses the API XML export that we already have and builds on top of the existing Import Merge Tool.

cdhorn:
Not sure I follow your thinking and not sure how practical this is...

Nick-Hall:
Why don't we just write directly to the database?

The only obstacle I can see at the moment is the metadata which is held in memory, but that should be easy to fix.

Then we could use [JSON patch](https://en.wikipedia.org/wiki/JSON_Patch) to modify existing objects. For example:

[{ "op": "replace", "path": "/gender", "value": "1" }]

This could be used with a PATCH request to change the gender of a person.

We could implement `person_patch` and associated database methods in core Gramps.  I have already prototyped this approach for the standard editors as a route to implement multi-user access.

cdhorn:
Right, the GUI seeing changes shouldn't be an issue AFAIK.  But the REST API is stateless so how can it push notifications to the clients not knowing who they are?  Would clients poll for changes?

DavidMStraub:
OK, let me elaborate what I meant. There are a couple of separate issues in my view.

### 1. Need for a Gramps &rarr; API synchronization

Let's say there is an official hosting service "GrampsCloud" and I host my tree there. I still want to keep using Gramps (as a desktop app) on my laptop and expect it to work as I'm used to when I'm sitting in an archive in a basement without wifi, or on a plane. I wouldn't want to accept, as a user, to only be able to work when I'm online. So there will be "offline" changes that need to be synced back to the "cloud". At the moment, I'm just rsync-ing my Gramps database directory to my web server, but once we allow write access to the API, there will be conflicts and there must be a way to solve them. The logic for solving these conflicts should be mostly contained in the "import merge tool", just that this would be the other way around: not merging changes from an XML file into the local tree but merging local changes into a remote tree (the API).

Note that this is quite similar (I presume) how other online services connect with their desktop counterparts, e.g. MyHeritage and Family Tree Maker (although I've never used them).

### 2. Need for an API &rarr; Gramps synchronization

This might not be necessary at the moment, but once we allow (multi-user) write access, there must be a way for a researcher to sync those changes back to their local tree; potentially with conflicts. This is exactly what the "import merge tool" does, so I think it's not difficult to implement.

[...]

PR #201, opened 2021-08-06
Add transactions endpoint

DavidMStraub:
This PR adds a new `/transactions/` endpoint for `POST` which allows to post a raw database transaction, of the same form that is returned in the response of the existing `POST`/`PUT`/`DELETE` requests.

There are two main applications for this endpoint:

1. To undo a transaction - e.g. an application can take the response from a `DELETE`, switch `delete` with `add` and `old` with `new` - and post it back to `/transactions/` to undo the delete. We could even make this simpler in the future by adding a `reverse` query parameter that does this reversion in the backend automatically.
2. For the synchronization addon (#163). In https://github.com/DavidMStraub/gramps-addon-webapisync, I am almost finished except for the part where I have to push the changes back to the API. With this new endpoint, this is very simple: during sync, I collect all the changes in a transaction on an in-memory DB, then finally I post the JSONized transaction to the new endpoint.

### DIFF
--- a/gramps_webapi/api/__init__.py
+++ b/gramps_webapi/api/__init__.py
@@ -80,6 +80,7 @@ from .resources.user import (
     UserTriggerResetPasswordResource,
 )
 from .resources.objects import CreateObjectsResource
+from .resources.transactions import TransactionsResource
 from .util import make_cache_key_thumbnails, use_args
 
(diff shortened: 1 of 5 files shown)


DavidMStraub:
Note to self: need to add the search index update as well (as in #199).

cdhorn (approved):
Also looks good at a quick glance.

Issue #186, opened 2021-07-13
Documentation contributions


DavidMStraub:
The [question came up](https://gramps.discourse.group/t/gramps-installation-on-a-vm-on-synology-nas/1634/13?u=davidmstraub) how to allow community contributions to the documentation. I see two options:

1. Making the wiki editable for the public
2. Moving the documentation to the repository itself (and have contributions via PRs)

Opinions?

@Nick-Hall @cdhorn @vnl

cdhorn:
I would go for option #2 and keep it under a top level docs directory.

DavidMStraub:
Right! We should probably do it in a way compatbile with [Readthedocs](https://docs.readthedocs.io/en/stable/intro/getting-started-with-sphinx.html) - with .rst it will still be nicely rendered on Github but we'll have the option to deploy it to RTD as well.

emyoulation:
An observation... as an end-user, I expect the documentation on GitHub to be development oriented... and that the tool hasn't moved from the 'geek prerequisite' stage into the 'ready for Aunt Martha to try' stage.

The documentation for Gramps tools is fragmented. Some is here on GitHub in various repositories. Some is in Sphinx on the Gramps-Project domain. Some is in the MediaWiki (in the User Reference Manual,  in dedicated pages, in GEPS proposals). Some is in MantisBT. Some is spread amount the various discussion forums: maillists, Discourse, Reddit, Geneanet, Element, Facebook, etc. Some are in forks of Gramps... like that of PortableApps.

So, at least SOME thought should be given to the path towards eventual consolidation for coherent presentation.

Nick-Hall:
I'm OK with either option.

DavidMStraub:
@emyoulation I see your point, but I think there is difference between Gramps (Desktop) and this project, because Aunt Martha :older_woman:  should definitely be using Gramps, but I doubt she'll ever be setting up a Web API instance (note this is in no way meant to be a sexist comment). So I think it's fine to follow best practices for documentation of Python packages.

I would really love to have an option for Aunt Martha to create her own instance, too, but I think this would require a hosted solution where you can create an instance by clicking a button.

vnl:
I think option two is good just to make sure it's clean.

emyoulation:
I fear you're underestimating the grit & determination of Aunt Marthas. 

[...]

PR #202, opened 2021-08-07
Move documentation from wiki to /docs (#186)

DavidMStraub:
As discussed in #186; I now just copied the Markdown files. We can take care of making this compatible with Readthedocs later...


DavidMStraub:
I'll merge this so we can go ahead and improve the documentation.

PR #206, opened 2021-08-16
Fix tests with webargs 8.0.1

DavidMStraub:
A small change in [webargs 8.0.1](https://webargs.readthedocs.io/en/latest/changelog.html) lead to many tests suddenly failing. This should make them pass again.

### DIFF
--- a/tests/test_endpoints/checks.py
+++ b/tests/test_endpoints/checks.py
@@ -48,11 +48,11 @@ def check_invalid_semantics(test, url, check="none", role=ROLE_OWNER):
     """Test that invalid parameters and values are handled properly."""
     check_list = {
         "none": [""],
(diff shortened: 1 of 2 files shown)


DavidMStraub:
I'd like to merge this immediately to make the tests pass again.

@cdhorn would be great if you can have a look afterwards and let me know if you see any issues. The change in webargs was that, previously, `DelimitedList` with an empty string was translated to `['']`, now to `[]` (which makes more sense). I removed the `?=` test cases for `base` and `list` since this no longer raises a 422.

The only other change I made is to change the test for `profile` in `test_get_people_handle_parameter_backlinks_validate_semantics` from boolean to list, which I think is correct (`profile` used to be a boolean but now is a list).

PR #211, opened 2021-09-01
Allow using the same JSON format for GET, POST, and PUT

DavidMStraub:
Implementing the UI for adding & updating objects in Gramps.js, I realized there was an issue with the current implementation of putting or posting objects to the API: it is assumed that they follow the same JSON structure as Gramps' own JSON methods in `gramps.gen.lib.serialize`. However, I had forgotten that the JSON objects we return are slightly different even for default query parameters:
- The `_class` attribute is always missing from JSON objects
- Types are represented as strings rather than as `GrampsType` objects, and the string is untranslated, while Gramps uses strings in the default locale. So, for a birth event, we have `{'type': 'Birth'}`, while Gramps, on a German locale, would serialize as `{'type': {'_class': 'EventType', 'string': 'Geburt'}`

The problem with this difference is that it would be impossible to fetch an object via `GET`, modify some values, and update the object again via `PUT`, since `PUT` expects the "raw" Gramps format.

The solution in this PR is a new function `fix_object_dict` called in the `post` and `put` methods, which simply inserts the `_class` attributes in the right places and turns the type strings into localized dicts, if necessary.

This way, the user is free to post/put objects either in the "raw" Gramps format or using the simplified format that is used for `GET`, and both will work the same way.

### DIFF
--- a/gramps_webapi/api/resources/base.py
+++ b/gramps_webapi/api/resources/base.py
@@ -142,7 +142,10 @@ class GrampsObjectResourceHelper(GrampsJSONEncoder):
             obj_dict["_class"] = self.gramps_class_name
         elif obj_dict["_class"] != self.gramps_class_name:
             abort(400)
-        obj_dict = fix_object_dict(obj_dict)
+        try:
+            obj_dict = fix_object_dict(obj_dict)
+        except ValueError:
+            abort(400)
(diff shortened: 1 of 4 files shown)


DavidMStraub:
Merging this for now.

PR #219, opened 2021-09-26
Fix email confirmation template

DavidMStraub:
A tiny fix: I noticed that due to a copy-and-paste error, the email confirmation dialog was not shown but only an empty page.


DavidMStraub:
Resubmitted as part of #220.

PR #229, opened 2021-10-09
Allow configuring from environment, add docker entrypoint

DavidMStraub:
The aim of this PR is to simplify the docker deployment process. It is still marked as draft as I also need to update the documentation.

Changes:
- The most important config variables can now be specified as environment variables. This has two advantages:
    - In most cases one can set up docker compose without the need to create and mount a config file
    - The settings that always need to be the same when using docker (the paths to the directories in the container) can be set as env vars in the image itself, so no need to specify in the config anymore
- The search index is now created on container creation, if it does not exist yet
- When the environment variable `GRAMPSJS_VERSION` is set, the Gramps.js frontend (corresponding to that released version) is downloaded into the static asset directory. This means users can decide with a simple env var if they want to use the API alone or add the Gramps.js frontend.


DavidMStraub:
Ready!

PR #247, opened 2021-12-01
Add place reference date to parent place profile

DavidMStraub:
This adds the date of a parent place reference to the place profile.

Related issue: https://github.com/DavidMStraub/Gramps.js/issues/39


DavidMStraub:
Need to redo - made a mistake.

PR #250, opened 2021-12-06
Increment version numbers for v0.3.0

DavidMStraub:


DavidMStraub:
Bad idea, we should have v0.2.2 instead as there are no new features really.

PR #260, opened 2021-12-19
Add support for PostgreSQL (#255)

Rahel-A:
* Pass POSTGRES_PASSWORD environment variable to Database Manager to open PostgreSQL.
* Install the required prerequisites for PostgreSQL in the Docker
container.

PR #265, opened 2022-01-08
Add support for PostgreSQL

Rahel-A:
- Pass POSTGRESQL_PASSWORD environment variable to Database Manager to open PostgreSQL.
- Install the required prerequisites for PostgreSQL in the Docker container.

Also see PR #260 for previous comments on this feature...

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,16 @@
 FROM dmstraub/gramps:5.1.4
 
+ENV GRAMPS_VERSION=51
 WORKDIR /app
 ENV PYTHONPATH="${PYTHONPATH}:/usr/lib/python3/dist-packages"
 
-# install poppler (needed for PDF thumbnails) and ffmpeg (needed for video thumbnails)
+# install poppler (needed for PDF thumbnails)
+# ffmpeg (needed for video thumbnails)
+# postgresql client (needed for PostgreSQL backend)
 RUN apt-get update && apt-get install -y \
   poppler-utils ffmpeg libavcodec-extra \
+  unzip \
+  libpq-dev postgresql-client postgresql-client-common python3-psycopg2 \
   && rm -rf /var/lib/apt/lists/*
 
 # set locale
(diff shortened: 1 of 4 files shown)


DavidMStraub:
Sorry, but for some reason this branch still has the two previous issues:

- `POSTGRES_PW` is not a config variable
- `str or None` is used as type hint

Could it be that you pushed the wrong branch?

Rahel-A:
> Sorry, but for some reason this branch still has the two previous issues:
> 
>     * `POSTGRES_PW` is not a config variable
> 
>     * `str or None` is used as type hint
> 
> 
> Could it be that you pushed the wrong branch?

I'm confused, let's resync...

I'm using POSTGRES_PASSWORD instead of POSTGRES_PW to match the environment variable name with PostgreSQL. The gramps-webapi can read POSTGRES_PASSWORD env var as set through (docker) environment:
https://github.com/gramps-project/gramps-webapi/blob/c6a121221982db015919c70ffa37fb061fa737ec/gramps_webapi/app.py#L77
Are you asking me to add another method for gramps-webapi to read POSTGRES_PASSWORD (through some other configuration)?

I tried `password: Optional[str] = None` as your comment suggested but that didn't change any test case failures, instead I found that setting the type to be either str or None makes the test cases pass:
https://github.com/gramps-project/gramps-webapi/blob/c6a121221982db015919c70ffa37fb061fa737ec/gramps_webapi/dbmanager.py#L43

DavidMStraub:
> I tried password: Optional[str] = None as your comment suggested but that didn't change any test case failures, instead I found that setting the type to be either str or None makes the test cases pass:

Which failures? `str or None` is not a valid type hint. I think it will just evaluate as `str` (since `str` is not falsey the part after `or` is never evaluated.

> Are you asking me to add another method for gramps-webapi to read POSTGRES_PASSWORD (through some other configuration)?

Yes, it should just be a normal configuration parameter that can be optionally set from the environment, as I suggested in the previous PR:

```python
app.config["POSTGRES_PASSWORD"] = app.config["POSTGRES_PASSWORD"] or os.getenv("POSTGRES_PASSWORD")`
```
and then you can just use `app.config["POSTGRES_PASSWORD"]` in the instantiation.

Rahel-A:
> > I tried password: Optional[str] = None as your comment suggested but that didn't change any test case failures, instead I found that setting the type to be either str or None makes the test cases pass:
> 
> Which failures? `str or None` is not a valid type hint. I think it will just evaluate as `str` (since `str` is not falsey the part after `or` is never evaluated.
> 
> > Are you asking me to add another method for gramps-webapi to read POSTGRES_PASSWORD (through some other configuration)?
> 
> Yes, it should just be a normal configuration parameter that can be optionally set from the environment, as I suggested in the previous PR:
> 
> ```python
> app.config["POSTGRES_PASSWORD"] = app.config["POSTGRES_PASSWORD"] or os.getenv("POSTGRES_PASSWORD")`
> ```
> 
> and then you can just use `app.config["POSTGRES_PASSWORD"]` in the instantiation.

OK, sorry for the confusion earlier. I've addressed your comments and the test cases pass now.

DavidMStraub:
Thanks!

DavidMStraub:
Oops, I forgot that this won't work until https://github.com/gramps-project/addons-source/pull/523 is merged. Anyway I think it's fine if we don't release the next version before.

Rahel-A:
> Oops, I forgot that this won't work until [gramps-project/addons-source#523](https://github.com/gramps-project/addons-source/pull/523) is merged. Anyway I think it's fine if we don't release the next version before.

There's also the other workaround available until that gets merged, you can set the number of GUNICORN workers to 1.

PR #278, opened 2022-04-08
Fix relationship translation

DavidMStraub:
I noticed that the output of the `/relations/` endpoint was sometimes not translated and found this is due to a missing locale parameter being passed in one method which is only needed for some translations; e.g. the stepgrandaunt example in the unit tests was translated, but "husband" wasn't. This PR fixes this issue and adds 1 unit test.


DavidMStraub:
Test failing due to https://github.com/gramps-project/gramps-webapi/issues/279

PR #290, opened 2022-06-24
Restrict flask-jwt-extended version

DavidMStraub:
This PR removes two unit tests that started failing recently.

It turned out this is due to an unexpected change in the flask-jwt-extended library; since 4.4.0, it allows fetching a new refresh token also with a valid access token.


DavidMStraub:
It turned out the change was not intentional and will be reverted (https://github.com/vimalloc/flask-jwt-extended/issues/483), so I restored the tests and instead specified to not use the two affected versions of the package.

PR #303, opened 2022-08-01
Correctly center square thumbnail

DavidMStraub:
I noticed that the square thumbnail used for object views (person, family, ...) is only centered vertically, but not horizontally in the image, which looks bad for photos in portrait orientation (family photos!).


DavidMStraub:
Also noticed & fixed another issue with the UUID in the new tokens.

PR #314, opened 2022-09-16
Change default locale to en_US

DavidMStraub:
The default locale of the Docker image is currently en_GB, but I realized this was a suboptimal choice as the "untranslated" strings in Gramps all correspond to en_US. So in cases where strings are translated at import time (e.g. reports or exporter descriptions), using en_GB will make it impossible to translate (e.g. in cases the word "centre" is involved...).


Nick-Hall:
Yes.  The default in Gramps is "en" which uses the US English base strings.

PR #331, opened 2023-02-01
Enable the running as any non-root user

ludovicus3:
Good security practice is to run as non-root. Some kubernetes environments such as Openshift/OKD don't allow root by default. The [documentation](https://docs.okd.io/4.11/openshift_images/create-images.html) gives an explanation of how to do this. I have implemented it here for the webapi.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,7 +28,9 @@ RUN mkdir /app/static && touch /app/static/index.html
 RUN mkdir /app/db && mkdir /app/media && mkdir /app/indexdir && mkdir /app/users
 RUN mkdir /app/thumbnail_cache
 RUN mkdir /app/tmp
-RUN mkdir -p /root/.gramps/gramps$GRAMPS_VERSION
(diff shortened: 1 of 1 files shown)


DavidMStraub:
Hi,

thanks for your contribution, much appreciated.

However we can't merge it like this since it would break all existing deployments that mount the database volume into the root folder. So please consider the Gramps Web [docs](https://gramps-project.github.io/web/) and add simlinks to the image to make sure the change is backwards compatible.

The commit with redis caching for the limiter was removed again, right? Please open an issue for that in case you are working on it, there are some upcoming changes to configuration handling in #330.

ludovicus3:
Yes, the redis was removed. That's the first and last time I used the github web IDE. I ended up having to move that commit around to get it to the right branch.

I'll have to give it some more thought. The symlink would work to access the plugins. But for the data base and recent-files-gramps.xml, not so much. The USER_DB_URI only affects the sqlite database not the recent files xml.

PR #350, opened 2023-03-28
Use task queue for exporter

DavidMStraub:
This uses the celery task queue for database exports. In addition to the existing `/api/exporters/<extension>/file` GET endpoint (which I think could be removed later) there is a new `/api/exporters/<extension>/file` POST endpoint for starting the export, which will return a filename that can then be downloaded from `/api/exporters/<extension>/file/processed/<filename>`, either immediately after receiving status 201 (without task queue) or after the task has completed (check `/api/tasks/<id>`).


DavidMStraub:
I also disabled gpkg export as exporting media files currently does not work for S3 and for local files I think it is a security risk at present.

PR #353, opened 2023-03-29
Add transliterated strings to search index (fixes #352)

### LINKED ISSUE #352: Allow searching with ASCII characters to show results which include non-ASCII characters
When I'm searching for a place name to add to an event, I need to write non-ASCII characters if I want to find a place name includes non-ASCII characters in its name.

Take this example: I want to add "Trąbczyn" as a place to an event. In the "Select Place" dialogue window, if I type "Trabczyn" then no result shows. In order to get around this, I need to edit the place and include an alternative name of "Trabczyn" - which I think is not the best solution.

I would argue that the expected user interaction is that I can type "Trabczyn" and see results which includes "Trąbczyn". Is it possible to do this? Otherwise I always need to add an ASCII-friendly version of a place name as an alternative name.

(I'm using the term ASCII here because it seems like this is the issue, but please correct me if it's not the proper way to describe this problem)

DavidMStraub:
This adds transliterated strings to the search index if the strings contain non-ASCII characters. This allows to search for Mr. Шестаков by searching for Shestakov, for instance. See #352.

### DIFF
--- a/gramps_webapi/api/search.py
+++ b/gramps_webapi/api/search.py
@@ -22,11 +22,12 @@
 from collections import OrderedDict
 from datetime import datetime
 from pathlib import Path
-from typing import Any, Dict, Generator, List, Optional, Tuple
+from typing import Any, Dict, Generator, List, Optional, Sequence, Tuple
 
 from flask import current_app
 from gramps.gen.db.base import DbReadBase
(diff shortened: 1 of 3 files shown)


takaji10:
Fantastic! Thanks very much for your work on Gramps. Looking forward to trying this out. :)

PR #362, opened 2023-04-25
Allow multiple trees

### LINKED ISSUE #346: Changes to TREE config option
Currently, TREE contains the database name (not ID!) and selects the tree to serve.

With multi-tree support, it is no longer strictly needed, but it makes sense to keep it to make it simpler serving a single tree.

An empty value would then indicate that multi-tree support is enabled.

### LINKED ISSUE #348: Add tree ID to user registration endpoint
The user registration endpoint is for new users to self-register to a given tree. When there are multiple trees, we need the tree ID to be provided as an additional input.

To clean up registered accounts which never validated their email address (spam accounts), it might make sense to add a timestamp column to the user database.

DavidMStraub:
Fixes #346, fixes #348.

This allows running an instance that gives access more than one tree. This is enabled by setting the `TREE` config option to `"*"`.


DavidMStraub:
Documentation PR: https://github.com/gramps-project/web/pull/12

PR #418, opened 2023-08-15
Speed up iter handles in private proxy db

### LINKED ISSUE #417: Performance issues with PrivateProxyDb
Users are reporting poor performance for "guest" users on some trees, which is realted to slowness of the PrivateProxyDb. Investigation is needed what the root cause of these speed issues is and how they can be remedied.

DavidMStraub:
In a first effort to fix #417, this PR drastically reduces the time taken by the iter handle methods of Gramps' private proxy db. Speedups of a factor of 30 and more are possible.

The existing implementation executed a database query for each object just to check whether it has private: true in its raw data. But we have a database column "private" and can simply filter for it in a single query!

This obviously only works when the base DB is a DBAPI instance, so if it is not, we fall back to the existing implementation.

@Nick-Hall, I could backport this to Gramps core as well if you think it makes sense.

### DIFF
--- a/gramps_webapi/api/util.py
+++ b/gramps_webapi/api/util.py
@@ -36,12 +36,25 @@ from gramps.cli.clidbman import NAME_FILE, CLIDbManager
 from gramps.gen.config import config
 from gramps.gen.const import GRAMPS_LOCALE
 from gramps.gen.db.base import DbReadBase
-from gramps.gen.db.dbconst import DBBACKEND
+from gramps.gen.db.dbconst import (
+    DBBACKEND,
+    KEY_TO_NAME_MAP,
+    PERSON_KEY,
+    FAMILY_KEY,
+    SOURCE_KEY,
+    EVENT_KEY,
+    MEDIA_KEY,
+    PLACE_KEY,
+    REPOSITORY_KEY,
+    NOTE_KEY,
+    CITATION_KEY,
+)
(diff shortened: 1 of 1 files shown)


Nick-Hall:
We don't allow database code in the proxies in Gramps core.  I seem to remember that an approach discussed previously was to pass a filter to the iterators and/or get_handle methods which would allow the backend to apply optimisations where appropriate.

DavidMStraub:
Makes sense. For Web API, it's OK as a fix for now, but in core I can work on a PR to add an object property filter method that falls back to iter_objects.

Nick-Hall:
There were a lot of discussions related to this back in 2016 on the `gramps-devel` mailing list.

A couple of threads jump out at me:
* [Filters and Seraches](https://sourceforge.net/p/gramps/mailman/message/35063922/)
* [Optimising filters to use SQL indexes](https://sourceforge.net/p/gramps/mailman/message/35020033/)

Many of our developers contributed to the discussion.  I suggest that you read the archives and restart the conversation.

I seem to remember that we were leaning towards passing a Filter object into the iterators.

emyoulation:
The Isotammi group invited review of how their Neo4j based website approached filtering ... noting the close relationship to the Gramps database and it being open source.

https://gramps.discourse.group/t/collaborate-on-optimizing-a-new-custom-rule/3063/10

PR #422, opened 2023-08-18
Version Conflict Fix Resolves issue 421 

Samurai336:
pinned version not needed and prevents conflict on dependency install

https://github.com/gramps-project/gramps-webapi/issues/421


DavidMStraub:
Thanks! Actually without the pin, the whole line is redundant. I removed it, let's see if unit tests pass.

PR #425, opened 2023-08-31
Test with Gramps 5.2

DavidMStraub:
See what happens when running unit tests with Gramps 5.2.

Expected to fail.


DavidMStraub:
Surprisingly few errors!

- `Heirat` vs `Hochzeit`: translation change
- different date fromat: `1760-02-00` vs `1760-02` etc.
-  'could not find foreign type FontOptions'

PR #484, opened 2024-02-03
Allow disabling registration, treat admins like owners in single-tree setup

### LINKED ISSUE #475: Disable registration
Apologies if this is already a feature and I'm just somehow  missing it - my family hopes to have a private (but publicly available) Gramps installation. This would mean disabling the Registration form in the login page. Is this possible? If not, please consider it as a feature, as I imagine this is a significant use-case.

### LINKED ISSUE #482: Registration disabled in single-tree setup if only administrator exists
I got a fresh docker-compose setup and I want to allow registration. Currently a click on a registration submit tells me, that registration is disabled.

In the first sentence of the [documentation](https://www.grampsweb.org/user-guide/registration/) it just says
`it will need to be enabled manually by a site owner before it can be used to log in`

But how do I do that exactly? I can't find any configuration possiblity for that

Gramps 5.1.6
Gramps Web API 1.5.1
Gramps Web Frontend 24.1.2
locale: en
multi-tree: false
task queue: true

DavidMStraub:
This PR fixes several issues related to new user registration:

- It allows explicitly disabling registration with a new config option `REGISTRATION_DISABLED` - fixes #475 
- It does not automatically disable registration in a single-tree setup if an administrator, but no owner exists - fixes #482 
- It sends the new user notification e-mail also to administrators in case of a single-tree setup

### DIFF
--- a/gramps_webapi/api/resources/user.py
+++ b/gramps_webapi/api/resources/user.py
@@ -558,12 +558,15 @@ class UserConfirmEmailResource(LimitedScopeProtectedResource):
             # otherwise it has been confirmed already
             modify_user(name=username, role=ROLE_DISABLED)
             tree = get_tree_from_jwt()
+            is_multi = current_app.config["TREE"] == TREE_MULTI
             run_task(
                 send_email_new_user,
                 username=username,
                 fullname=current_details.get("full_name", ""),
                 email=claims["email"],
                 tree=tree,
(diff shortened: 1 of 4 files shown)


DavidMStraub:
```yaml
REGISTRATION_DISABLED: "True"
```
&rarr;
```yaml
GRAMPSWEB_REGISTRATION_DISABLED: "true"
```

PR #489, opened 2024-02-04
fix mails over tls

Shablone:
Send mails using the flag use_tls was only following the workflow of sending via ssh (not tls), but tls uses another workflow then ssh. This change allows using tls

### DIFF
--- a/gramps_webapi/api/util.py
+++ b/gramps_webapi/api/util.py
@@ -396,13 +396,15 @@ def send_email(
     use_tls = get_config("EMAIL_USE_TLS")
     try:
         if use_tls:
-            smtp = smtplib.SMTP_SSL(host=host, port=port, timeout=10)
-        else:
-            smtp = smtplib.SMTP(host=host, port=port, timeout=10)
-            smtp.ehlo()
-            if port != 25:
+            if port == 465:
+                smtp = smtplib.SMTP_SSL(host=host, port=port, timeout=10)
+            else:
+                smtp = smtplib.SMTP(host=host, port=port, timeout=10)
+                smtp.ehlo()
                 smtp.starttls()
                 smtp.ehlo()
+        else:
+            smtp = smtplib.SMTP(host=host, port=port, timeout=10)
         if user:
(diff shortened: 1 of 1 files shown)


DavidMStraub:
Sorry, but that's not correct. Please see the [smtplib docs](https://docs.python.org/3/library/smtplib.html):

> An [SMTP_SSL](https://docs.python.org/3/library/smtplib.html#smtplib.SMTP_SSL) instance behaves exactly the same as instances of [SMTP](https://docs.python.org/3/library/smtplib.html#smtplib.SMTP). [SMTP_SSL](https://docs.python.org/3/library/smtplib.html#smtplib.SMTP_SSL) should be used for situations where SSL is required from the beginning of the connection and using starttls() is not appropriate.

The existing code works perfectly fine with TLS.

Shablone:
See [discussion](https://github.com/gramps-project/gramps-web-api/discussions/488#discussion-6178287)
After change of this code, the tls configuration from this discussion works. So there definetly is an issue with that.

> [SMTP_SSL](https://docs.python.org/3/library/smtplib.html#smtplib.SMTP_SSL) should be used for situations where SSL is required from the beginning of the connection and using starttls() is not appropriate`

doesn't this exactly tell, that SMTP_SSL shouldn't be the default? From my personal test I can tell you that with gmail on port 587 is does not work with an instance of SMTP_SSL and only works with SMTP and a call of starttls()

DavidMStraub:
> doesn't this exactly tell, that SMTP_SSL shouldn't be the default? From my personal test I can tell you that with gmail on port 587 is does not work with an instance of SMTP_SSL and only works with SMTP and a call of starttls()

Yes. Because it uses STARTTLS. From our [documentation](https://www.grampsweb.org/Configuration/):

> `EMAIL_USE_TLS` Boolean, whether to use TLS for sending e-mails. Defaults to True. **When using STARTTLS, set this to False and use a port different from 25.**

Shablone:
Ok. I missunderstood.

Issue #285, opened 2022-05-15
Improve search


DavidMStraub:
I realized it's currently difficult to search for events since the name of the event participants is not part of the `text_data_list` that gets indexed.

Another issue is that type names (e.g. Baptism) are only indexed in the main locale, so it's not possible to search for a localized event type ("Taufe"), for instance.

DavidMStraub:
Another one: it's almost impossible to search for families, as the names of the parents to not appeat in the text data list.

DavidMStraub:
> I realized it's currently difficult to search for events since the name of the event participants is not part of the text_data_list that gets indexed.

> Another one: it's almost impossible to search for families, as the names of the parents to not appeat in the text data list.

Those two will be fixed as part of #525.

> Another issue is that type names (e.g. Baptism) are only indexed in the main locale, so it's not possible to search for a localized event type ("Taufe"), for instance.

Not fixed yet, but anyway can't be fixed by changing what is indexed, but by translating the search query.

PR #528, opened 2024-07-23
Replace whoosh search backend by sifts

### LINKED ISSUE #525: Replace whoosh by sifts
Although whoosh works ok as our full-text search engine, I've been unhappy with it for a while, for two reasons:

1. It is simply not maintained anymore. Last commit 2 years ago, last release more than 5 years ago.
2. It can only store the index in the local file system, which makes scaling with containers more challenging.

But we are also using only a small subset of whoosh's feature set, and moving to an even more powerful (& difficult to deploy) server-client based engine would be overkill.

So, eventually I simply implemented a minimal new full-text search library that leverages the built-in full-text search in SQLite and PostgreSQL. Both backends are supported, so SQLite can be used by default and Postgres can be used for scaling multi-tree deployments.

The new library is called [sifts](https://github.com/DavidMStraub/sifts).

DavidMStraub:
Fixes #525 - see there for the rationale.

This changes the allowed search syntax and API slightly:

- The search syntax no longer supports `type:` and `change:` - this was not properly documented anyway
- Instead, there is a new `type` query argument. To only return people and families, use `?type=person,family`.
- AND/OR are now case insensitive
- The wildcard * only works at the end of a word

To Do:

- [x] API docs
- [x] Add sifts version to metadata


DavidMStraub:
This is ready to merge; however I'd like to work a little bit on https://github.com/gramps-project/gramps-web-api/issues/285 before merging this, as users will anyway have to recreate their search index.

PR #530, opened 2024-08-04
Speed up incremental search reindex

### LINKED ISSUE #436: GEDCOM import is too slow
I'm using GrampsHub (premium version). I was not able to import my GEDCOM file containing about 12,000 people and 4,500 families. There is no error message and I cannot find any logfile.

It was possible to import a smaller file, but it needed hours. The performance has to be improved.

DavidMStraub:
The incremental search reindex was *very* inefficient in case of many updates. And that is what is used after importing a file. Imprting `example.gramps` locally using a SQLite index, before:

```
[2024-08-04 17:43:40,381: INFO/ForkPoolWorker-6] Task
gramps_webapi.api.tasks.import_file[48c173e5-0d0f-4cd6-b682-aef48bbdda9e]
succeeded in 2314.4943576940022s: None
```

... and after:

```
[2024-08-04 22:24:36,800: INFO/ForkPoolWorker-6] Task 
gramps_webapi.api.tasks.import_file[57839f70-15a8-4540-9d46-9d3e63f6d451]
succeeded in 77.58003538799676s: None
```

:open_mouth: 

I believe this fixes #436.

### DIFF
--- a/gramps_webapi/_version.py
+++ b/gramps_webapi/_version.py
@@ -18,4 +18,4 @@
 #
 
 # make sure to match this version with the one in apispec.yaml
-__version__ = "2.4.0"
+__version__ = "2.4.1"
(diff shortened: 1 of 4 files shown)


DavidMStraub:
Added another fix: the search index wasn't updated after using the delete objects feature.

PR #533, opened 2024-08-08
fix: dirty workaround create search index in docker-entrypoint.sh just for single tree

genofire:
…t for single tree

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -14,7 +14,7 @@ then
 fi
 
 # Create search index if not exists
-if [ -z "$(ls -A /app/indexdir/*.db)" ]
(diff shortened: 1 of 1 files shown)


DavidMStraub:
Thanks for your efforts, really appreciated!

There is still one problem though, we cannot assume that there is always a `TREE` env var (in fact it's deprecated and should be `GRAMPSWEB_TREE`, the tree can also be configured in a config file...

DavidMStraub:
Sorry for keeping this open but I had to think about it a bit more. I realized that, since v2.4, we report on the `/metadata/` endpoint whether the search index is up to date, and the Gramps Web frontend displays a notification if it isn't. With that in place, I start to think the cleanest solution wouldn't be just getting rid of this entirely.

PR #538, opened 2024-08-27
Generate app/secret only if SECRET_KEY is unset

pwalkr:
If we already have a secret key, we don't need to create another.

I caught this trying to reduce the number of volume mounts (https://github.com/gramps-project/gramps-web/discussions/478) while simultaneously running as non-root. Compose failed to start anything:
```
grampsweb_celery  | mkdir: cannot create directory ‘/app/secret’: Permission denied
grampsweb-1       | mkdir: cannot create directory ‘/app/secret’: Permission denied
```
Injecting this `docker-compose.sh` let me move my secret to env and the service starts and runs normally.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -1,15 +1,15 @@
 #!/bin/sh
 set -e
 
-# create random flask secret key
-if [ ! -s /app/secret/secret ]
-then
-    mkdir -p /app/secret
-    python3 -c "import secrets;print(secrets.token_urlsafe(32))"  | tr -d "\n" > /app/secret/secret
-fi
 # use the secret key if none is set (will be overridden by config file if present)
 if [ -z "$GRAMPSWEB_SECRET_KEY" ]
 then
+    # create random flask secret key
+    if [ ! -s /app/secret/secret ]
+    then
+        mkdir -p /app/secret
+        python3 -c "import secrets;print(secrets.token_urlsafe(32))"  | tr -d "\n" > /app/secret/secret
+    fi
     export GRAMPSWEB_SECRET_KEY=$(cat /app/secret/secret)
 fi
 


DavidMStraub:
Thanks! The problem with this is that using env variables prefixed with `GRAMPSWEB_` is only one of several configuration options; if the secret key is set in a config file, this is not detected here.

pwalkr:
If the secret key is set in a config file, it's likely `mkdir -p /app/secret` still won't be doing the right thing. In that case one could set `GRAMPSWEB_SECRET_KEY=unused` to avoid any `mkdir` errors on start. But it's not the cleanest solution.

What's ideal? Something like `python3 -m gramps_webapi --config /app/config/config.cfg export_config` that bash could grok? Move mkdir into python? I'd be interested to apply it in resolving similar indexdir warnings:
```
grampsweb_celery  | ls: cannot access '/app/indexdir/*.db': No such file or directory
grampsweb-1       | ls: cannot access '/app/indexdir/*.db': No such file or directory
```

DavidMStraub:
Your solution is in any case better than the status quo, so let's merge. Force-pushing to try to get tests to run through.

pwalkr:
I've been tinkering with a python-enhanced alternative, would this be closer to ideal?
```
#!/bin/sh

# mkdirs and pick up any variables from flask config
eval "$(python3 - <<EOC
import os
import sys
from flask import Flask
from gramps_webapi.config import DefaultConfig
from gramps_webapi.const import ENV_CONFIG_FILE
import secrets

# Merge config
app = Flask('config_dump')
app.config.from_object(DefaultConfig)
app.config.from_envvar(ENV_CONFIG_FILE)
app.config.from_prefixed_env(prefix="GRAMPSWEB")

# Handle secret - prints "export" so shell can handle~env before exec
SECRET_PATH = '/app/secret/secret'
if not app.config["SECRET_KEY"] and not os.path.isfile()
    #secrets.token_urlsafe(32) ... write to file
    # Export for shell to eval
    print("export GRAMPSWEB_SECRET_KEY={}".format(cdir))

# mkdirs
for cdir in [
        app.config["REPORT_DIR"],
        app.config["EXPORT_DIR"],
        app.config["THUMBNAIL_CACHE_CONFIG"]["CACHE_DIR"],
        app.config["MEDIA_BASE_DIR"]]:
    if not os.path.isdir(cdir):
        print("+ mkdir {}".format(cdir), file=sys.stderr)
        os.makedirs(cdir)
EOC
)"

# TBD user migration and index here in shell or in python, whichever cleaner

exec "$@"
```
Could be cleaned up with config-merge logic consolidated in a `gramps_webapi` method. I'm stuck on how best to set `SECRET_KEY` and avoid race conditions between api and celery containers (maybe shell's atomic `mkdir`)

DavidMStraub:
Let's start with the simple version for now :slightly_smiling_face:

PR #541, opened 2024-08-30
python.linting settings are deprecated

JohnRDOrazio:
Just a simple fix for the updated Python extensions in VSCode, seeing that the `settings.json` for VSCode is checked into the repo... I was wondering where the warnings were coming from until I realized that the `settings.json` was in the repo itself.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,6 +1,7 @@
 {
   "recommendations": [
     "ms-python.python",
+    "ms-python.black-formatter",
     "editorconfig.editorconfig"
   ]
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,11 +1,4 @@
 {
-
-  "python.formatting.provider": "black",
-  "python.linting.pydocstyleEnabled": true,
-  "python.linting.pylintEnabled": true,
-  "python.linting.mypyEnabled": true,
-  "python.linting.flake8Enabled": false,
-  "python.linting.pylintArgs": ["--rcfile=setup.cfg"],
   "[python]": {
     "editor.formatOnSave": true
   }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,6 @@
 {
   "[python]": {
+    "editor.defaultFormatter": "ms-python.black-formatter",
     "editor.formatOnSave": true
   }
 }


DavidMStraub:
Good catch, but then we should either migrate it to the new preferred way of doing things (adding the MS pylint etc. extensions as recommended extensions) or remove the `.vscode` folder entirely.  Maybe having pre-commit in place makes editor settings less relevant.

JohnRDOrazio:
There are a few choices here, as indicated [here](https://code.visualstudio.com/docs/python/linting):

> Microsoft publishes the following linting extensions for Python:
> 
> | Linter	| Extension |
> |--------|------------|
> | Pylint	| https://marketplace.visualstudio.com/items?itemName=ms-python.pylint |
> | flake8	| https://marketplace.visualstudio.com/items?itemName=ms-python.flake8 |
> | mypy	| https://marketplace.visualstudio.com/items?itemName=ms-python.mypy-type-checker |
>
> Linting extensions offered by the community:
> 
> | Linter	| Extension |
> |--------|------------|
> | Ruff	| https://marketplace.visualstudio.com/items?itemName=charliermarsh.ruff |
> | mypy	| https://marketplace.visualstudio.com/items?itemName=matangover.mypy |
>

JohnRDOrazio:
Seems that the `black` formatter can also be reinstated with https://marketplace.visualstudio.com/items?itemName=ms-python.black-formatter .

JohnRDOrazio:
I'm not very familiar with the python linters, but I'm seeing `flake8` being used in the source code on the one hand while it was disabled as a linter for VSCode. Should `flake8` be among the recommended linters? Should it be enabled in VSCode?
```diff
-  "python.linting.pylintEnabled": true,
-  "python.linting.mypyEnabled": true,
-  "python.linting.flake8Enabled": false,
```
The `pylint` and `flake8` linters are now extensions offered by `ms-python`, while the `mypy` linter is offered by `matangover.mypy`. Which do you think should be recommended and / or enabled?

It seems that once `ms-python.pylint` is installed it will be used automatically:
> https://marketplace.visualstudio.com/items?itemName=ms-python.pylint
> Once this extension is installed in Visual Studio Code, Pylint is automatically executed when you open a Python file

Same with `my-python.flake8`:
> https://marketplace.visualstudio.com/items?itemName=ms-python.flake8
> Once this extension is installed in Visual Studio Code, Flake8 is automatically executed when you open a Python file
> You can skip linting with Flake8 for specific files or directories by setting the `flake8.ignorePatterns` setting.

I also see `mypy` pulled in among the source code dependencies. I'm testing development in a virtual environment and I see on the extension page:
> https://marketplace.visualstudio.com/items?itemName=matangover.mypy
> Some people prefer to have mypy installed in each project's virtual environment rather than in a global location. To do this, enable `mypy.runUsingActiveInterpreter` (either globally or for a specific workspace).
Should I add this setting too?

There is also a microsoft extension for `mypy` type checking. The `matangover` extension states that:
> Runs on your entire workspace. (This is different from Microsoft's Python extension's mypy functionality which only lints each file separately, leading to incomplete type checking.)

However the microsoft extension states that:
> Workspace-wide type checking: By default, this extension will only report errors and warnings for files open in the editor. However, you can configure it to report errors and warnings for the entire workspace through the mypy-type-checker.reportingScope setting.

PR #542, opened 2024-08-30
silence warning gtk version

JohnRDOrazio:
Gtk was imported without specifying a version first. Use gi.require_version('Gtk', '3.0') before import to ensure that the right version gets loaded.


DavidMStraub:
Awesome thanks, this warning keeps confusing users.

Issue #545, opened 2024-09-02
Migrate from setup.py to pyproject.toml


DavidMStraub:

DavidMStraub:
Fixed by #547.

PR #547, opened 2024-09-03
Fix issue #545: migrate to pyproject.toml

JohnRDOrazio:
In order to succeed in passing the pre-commit hook, I had to update with `pre-commit autoupdate`, and then I had to remove a bunch of deprecated settings.

I'm using Python 3.12 within a virtual environment.

The pre-commit hook still wasn't succeeding in importing `setuptools`, probably because I'm working within a virtual environment. Easiest way to fix / skip that error was to add `# pylint: disable=E0401` in `setup.py`. Not sure if there's a better way (for example using `init-hook` under `[MASTER]`, I tried a few solutions but none seemed to work).


DavidMStraub:
Awesome, thanks! 

I asked my artificial friend for a review :wink: below is its feedback. Can you please check whether it's right? Thanks!

---


Your `pyproject.toml` file is mostly valid, but there are a few points that could be checked or improved:

1. **`dependencies` Section:**
   - The field `dependencies` should be listed under `[project]` but as `requires = [...]` instead of `dependencies = [...]`. If your toolchain (like `hatch`, `poetry`, or `flit`) supports `dependencies`, it's fine, but generally, the correct term in the PEP 621 standard is `requires`.

2. **License Specification:**
   - The license should ideally follow a standard identifier. While you used `"AGPL v3 or greater"`, it's recommended to use `"AGPL-3.0-or-later"` as per SPDX identifiers.

3. **Optional Dependencies:**
   - If there are optional dependencies, they should be listed in a `[project.optional-dependencies]` table. However, it looks like all your dependencies are required.

4. **Order and Grouping:**
   - It's generally good practice to group related keys together. For example, `readme`, `keywords`, and `classifiers` are fine, but placing them right after `name`, `version`, `authors`, and `license` may improve readability.

5. **`dynamic` Field:**
   - You have the `dynamic = ["version"]` field which indicates that the version is dynamically computed. Ensure that this is handled correctly in your setup, especially with `setuptools_scm`.

### Improved Version of `pyproject.toml`

Here's a slightly revised version that follows these suggestions:

```toml
[project]
name = "gramps-webapi"
description = "A RESTful web API for the Gramps genealogical database."
authors = [
  { name = "Gramps Development Team" },
  { name = "David M. Straub", email = "straub@protonmail.com" }
]
license = { text = "AGPL-3.0-or-later" }
readme = "README.md"
dynamic = ["version"]
keywords = ["RESTful", "web API", "genealogy", "Gramps"]
classifiers = [
    "Programming Language :: Python :: 3",
    "License :: OSI Approved :: GNU Affero General Public License v3 or later (AGPLv3+)",
    "Operating System :: OS Independent",
]
requires = [
    "gramps-desktop[GUI,i18n]==5.2.*",
    "Click>=7.0",
    "Flask>=2.1.0",
    "Flask-Caching>=2.0.0",
    "Flask-Compress",
    "Flask-Cors",
    "Flask-JWT-Extended>=4.2.1,!=4.4.0,!=4.4.1",
    "Flask-Limiter>=2.9.0",
    "Flask-SQLAlchemy",
    "marshmallow>=3.13.0",
    "webargs",
    "SQLAlchemy",
    "pdf2image",
    "Pillow",
    "bleach[css]>=5.0.0",
    "jsonschema",
    "ffmpeg-python",
    "boto3",
    "alembic",
    "celery[redis]",
    "Unidecode",
    "pytesseract",
    "gramps-ql>=0.3.0",
    "sifts>=0.8.3",
]

[project.urls]
homepage = "https://github.com/gramps-project/gramps-web-api"
repository = "https://github.com/gramps-project/gramps-web-api"

[build-system]
requires = ["setuptools>=64.0", "setuptools_scm[toml]>=6.2", "wheel"]
build-backend = "setuptools.build_meta"

[tool.isort]
profile = "black"

[tool.setuptools_scm]
write_to = "gramps_webapi/_version.py"

[tool.setuptools.packages.find]
include = ["gramps_webapi", "gramps_webapi.*"]
```

This version corrects the `license` identifier and renames `dependencies` to `requires`, ensuring the structure is compliant with the PEP 621 standard. Make sure your build tool supports all these fields as specified.

DavidMStraub:
The failing unit test can be ignored - it's because we currently specify the version in the API spec manually, which won't work with SCM.

JohnRDOrazio:
I double checked the documentation. `requires` is used only in `[build-system]`, while `dependencies` is the keyword under `[project]`.
See https://setuptools.pypa.io/en/latest/userguide/dependency_management.html .

JohnRDOrazio:
And reading the [pep 621 specification](https://peps.python.org/pep-0621/#dependencies-optional-dependencies), I see that the keyword `requires` is specific to [Flit](https://flit.readthedocs.io/).

> For `dependencies`, it is a key whose value is an array of strings. Each string represents a dependency of the project and MUST be formatted as a valid [PEP 508](https://peps.python.org/pep-0508/) string. Each string maps directly to a `Requires-Dist` entry in the [core metadata](https://packaging.python.org/specifications/core-metadata/).
> 
> For `optional-dependencies`, it is a table where each key specifies an extra and whose value is an array of strings. The strings of the arrays must be valid [PEP 508](https://peps.python.org/pep-0508/) strings. The keys MUST be valid values for the `Provides-Extra` [core metadata](https://packaging.python.org/specifications/core-metadata/). Each value in the array thus becomes a corresponding `Requires-Dist` entry for the matching `Provides-Extra` metadata.
> 
> [Setuptools](https://setuptools.readthedocs.io/): `install_requires` for required dependencies, `extras_require` for optional dependencies ([link](https://setuptools.readthedocs.io/en/latest/setuptools.html#metadata))

JohnRDOrazio:
Another solution to the pre-commit hook problem could be found here: https://stackoverflow.com/a/61238571/394921

The author of the pre-commit hook suggests configuring pylint as a [local hook](https://pre-commit.com/#repository-local-hooks).

But I'm not very experienced with all this, perhaps it could be considered a separate issue to be worked on.

JohnRDOrazio:
Perhaps we can even remove the `license` field altogether, seeing we are using `Classifier`. From the python packaging docs:

> If you are using a standard, well-known license, it is not necessary to use this field. Instead, you should use one of the [classifiers](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#classifiers) starting with `License ::`.

DavidMStraub:
Ha, AI lied to me. Thanks for checking!

JohnRDOrazio:
My redocly linter is complaining that `version` is a required element under `info`.
Now that the test is no longer present perhaps we can leave the `version` element in `apispec.yml`?
I removed `gramps_webapi/_version.py` from the git cache in the meantime, seeing that it has been added to `.gitignore`

DavidMStraub:
The problem is that we would have to manually update the version in apispec each time *before* tagging a release...

DavidMStraub:
But let's attack one problem at a time, the toml looks good. Thanks!

PR #553, opened 2024-10-07
Added Object Query Language

dsblank:
This PR adds a Python Object Query Language to the gramps-web-api. This allows using standard Python on Gramps primary objects. This is based on @DavidMStraub 's "Gramps Query Language".

The QL lives here: https://github.com/dsblank/object-ql and can work on any Python objects. It contains adopted copies of the GQL tests, and should be a drop-in replacement for GQL.

Example queries include:

Find the person with a particular gramps_id:

```python
person.gramps_id == 'person001'
```

Find all of the people with notes:

```python
person.get_note_list()
```

Find all of the people with notes that mention 'vote':

```python
any([('vote' in str(get_note(handle).text)) for handle in person.get_note_list()])
```
If you don't know what the type is, you can use obj, like:

```python
"23" in obj.gramps_id
```

### DIFF
--- a/gramps_webapi/api/resources/base.py
+++ b/gramps_webapi/api/resources/base.py
@@ -24,6 +24,7 @@ from abc import abstractmethod
 from typing import Dict, List
 
 import gramps_ql as gql
+import pythonish_ql as pql
 from flask import Response, abort, request
 from gramps.gen.const import GRAMPS_LOCALE as glocale
 from gramps.gen.db import DbTxn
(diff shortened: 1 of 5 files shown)


DavidMStraub:
Thanks! I think it's always good to have multiple options to choose from, so I would be happy to include this even if there is overlap with GQL. Just some comments:

- The reason for the failing checks is that, currently, Gramps Web API still supports Python 3.8. I think it'll work if you just include `from __future__ import annotations` in your library in the relevant files. I know 3.8 reaches EOL this month, so we might drop support at some point in the near future, but not for the next release.
- Please include some unit tests. Ideally also with some malicious query attempt :slightly_smiling_face: 
- It would be good to settle on a name for your library. Are you sure it will stay "Pythonish QL"?

dsblank:
I think I have addressed all of the above, now with a new name `object-ql`. I'll continue to test on https://github.com/dsblank/object-ql and explore any possible security issues.

dsblank:
object_ql now supports a security infrastructure, and resource limitations. Updated minimum version to 0.1.2.

DavidMStraub:
Great, thanks! The only thing that's missing is some simple unit tests that the query with oql works as expected. For GQL, I was lazy and just added a couple of cases in `test_people.py`, you could just copy & paste them for OQL.

dsblank:
I'll finish this up with tests. But now that it is looking like we'll have a nice json repr, the prospect of having a query language that looks like python but queries like SQL is possible. Next challenge!

PR #557, opened 2024-10-14
Provide more insights in emails to admin and self-registered user

hgohel:
Changed email sent to self-registered user so they know that after they confirm their email address, an admin still needs to approve the account. User will also see this message when they confirm their email.

Changed admin notification to let them know that they need to set the self-registered user's role to allow them to log in.

Documentation updated with [gramps-web-docs PR #24](https://github.com/gramps-project/gramps-web-docs/pull/24)

### DIFF
--- a/gramps_webapi/api/emails.py
+++ b/gramps_webapi/api/emails.py
@@ -52,14 +52,15 @@ def email_confirm_email(base_url: str, token: str):
     """Confirm e-mail address e-mail text."""
     intro = (
         _(
-            "You are receiving this e-mail because you (or someone else) "
-            "have registered a new account at %s."
+            "You are receiving this email because this e-mail address "
+            "was used to register a new account at %s."
         )
         % base_url
     )
     action = _(
         "Please click on the following link, or paste this into your browser "
-        "to complete the process:"
+        "to confirm your email address. You will be able to log on once an "
+        "administrator reviews and approves your account."
     )
 
     return f"""{intro}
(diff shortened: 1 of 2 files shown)


gramps_webapi/api/emails.py
```diff
@@ -52,14 +52,15 @@ def email_confirm_email(base_url: str, token: str):
     """Confirm e-mail address e-mail text."""
     intro = (
         _(
-            "You are receiving this e-mail because you (or someone else) "
-            "have registered a new account at %s."
+            "You are receiving this email because this e-mail address "
```
DavidMStraub: email and e-mail in one line :slightly_smiling_face:

gramps_webapi/api/emails.py
```diff
@@ -52,14 +52,15 @@ def email_confirm_email(base_url: str, token: str):
...
+            "was used to register a new account at %s."
         )
         % base_url
     )
     action = _(
         "Please click on the following link, or paste this into your browser "
-        "to complete the process:"
+        "to confirm your email address. You will be able to log on once an "
+        "administrator reviews and approves your account."
```
DavidMStraub: I guess we should also use "tree owner" everywhere here

gramps_webapi/api/emails.py
```diff
@@ -52,15 +52,15 @@ def email_confirm_email(base_url: str, token: str):
     """Confirm e-mail address e-mail text."""
     intro = (
         _(
-            "You are receiving this email because this e-mail address "
+            "You are receiving this message because this e-mail address "
```
DavidMStraub, line 55: I just meant the inconsistent spelling (email vs e-mail), but good like this :blush:

hgohel: Doh! Didn't pick up on that :)

PR #570, opened 2024-11-14
Add ability to change Gunicorn timeout

Illia-M:
To process long requests on not powerful machines with a lot of data it will be helpful to be able extend timeouts


DavidMStraub:
Sure, why not. Thx

PR #576, opened 2024-11-21
Make server more flexible

dsblank:
This PR does the following:

1. Refactors the logging so there is only one (the app logger)
2. Adds flags to allow more flexibility
  * `--use-wsgi` to add a wsgi wrapper
  * `--host` for wsgi listener
  * `--debug-level` set the log level
  * `--log-file` if you want to log to file
  * `--open` open the browser on host:port
  * `--max-workers` for wsgi server
3. The default flag values should allow server to run exactly as before (although the log format is now more standard and colorized)


DavidMStraub:
Thanks, good ideas! The only problem I see is that if somebody pip-installs Gramps Web API and uses the CLI to run it with the new `--browser` option, it will simply open an empty page in the browser. I see two options to improve on this:

1. Check if `{STATIC_PATH}/index.html` exists and is non-empty. If not, issue a warning that for running Gramps Web, downloading the frontend is required, and refuse opening the browser.
2. Replace `static/index.html` shipped with the Python package by a placeholder file containing something like 
```html
<h1>Welcome to Gramps Web API!</h1>
<p>If you see this page, Gramps Web API is up and running.</p>
<p>Gramps Web API provides a REST API to work with a Gramps family tree database.
It also servers as the backend of Gramps Web, an integrated web app.
To run Gramps Web, <a href="https://github.com/gramps-project/gramps-web/releases">download
the frontend</a> and set the Gramps Web API `STATIC_PATH` configuration variable.</p>
```

What do you think?

gramps_webapi/__main__.py
```diff
@@ -58,12 +59,71 @@ def cli(ctx, config):
...
+def run(ctx, port, tree, host, open, debug_level, log_file, max_workers, use_wsgi):
     """Run the app."""
     app = ctx.obj["app"]
-    app.run(port=port, threaded=True)
+    debug_level = debug_level.upper()
+    if max_workers is None:
+        max_workers = min(32, os.cpu_count() + 4)
+
+    def open_webbrowser():
+        time.sleep(1.0)
```
DavidMStraub, line 102: why sleep?

gramps_webapi/__main__.py
```diff
@@ -58,12 +59,71 @@ def cli(ctx, config):
...
+    if open != "no":
+        thread = Thread(target=open_webbrowser)
+        thread.start()
+
+    if log_file:
+        file_handler = logging.FileHandler(log_file, "w+")
+        app.logger.addHandler(file_handler)
+        app.logger.setLevel(debug_level)
+
+    print("Running gramps-web...")
```
DavidMStraub: Gramps Web API

gramps_webapi/translogger.py
```diff
@@ -0,0 +1,137 @@
...
+"""
+Middleware for logging requests, using Apache combined log format
+"""
+
+import logging
+import time
+import urllib.parse
+
+
+class TransLogger(object):
```
DavidMStraub: `(object)` is not needed (I know this is copy-pasted from somewhere else, but we can modernize a bit)

gramps_webapi/__main__.py
```diff
@@ -58,12 +59,71 @@ def cli(ctx, config):
...
     """Run the app."""
     app = ctx.obj["app"]
-    app.run(port=port, threaded=True)
+    debug_level = debug_level.upper()
+    if max_workers is None:
+        max_workers = min(32, os.cpu_count() + 4)
+
+    def open_webbrowser():
+        time.sleep(1.0)
+        new = {"tab": 0, "window": 1}[open]
```
DavidMStraub: This will raise a `KeyError` if the user provides any unexpected string. I think it would be better to use a [click choice option](https://click.palletsprojects.com/en/stable/options/#choice-options) for the browser flag.

DavidMStraub:
One thing I'm unsure about: which things will break when using threads. So far, Web API was either served by the flask development server (which is single-threaded) or by gunicorn using worker processes (not threads). Did you investigate that already?

gramps_webapi/__main__.py
dsblank: That gives the server time to start. If you open the page too quickly, you'll get an error.

gramps_webapi/__main__.py
dsblank: Yes, I debated that. On one hand, the gramps-web-api is just the REST server. Thinking of suggesting renaming gramps-web to gramps-web-frontend,  plus gramps-web-api together as gramps-web (and I'm thinking about a package just called gramps-web). What do you think?

gramps_webapi/translogger.py
dsblank: I can fix a couple of other things there too. Now that I know you are open to this PR, I will do that.

gramps_webapi/__main__.py
dsblank: I'll take a look.

dsblank:
> Did you investigate that already?

I assume that we will not run those two paths with the other flags. I'm developing a third flow that runs the server + has frontend preinstalled and runs locally.

dsblank:
First, thanks for being open to changes! 

Regarding the:

> Check if {STATIC_PATH}/index.html exists and is non-empty. If not, issue a warning that for running Gramps Web, downloading the frontend is required, and refuse opening the browser.

I think that is good in general. Currently, one doesn't know if the frontend is available or not, right? It would be good to show something like:

```
Warning: the gramps-web-frontend was not found. Starting gramps-web-api as just
a server.
```

dsblank:
Does it ever make sense to run the gramps-web-api server without a frontend? I guess for testing. Anything else? Maybe that is just an error, without something like `--server-only`?

gramps_webapi/__main__.py
DavidMStraub: > Thinking of suggesting renaming gramps-web to gramps-web-frontend, plus gramps-web-api together as gramps-web

That's exactly the naming we have. The frontend used to have a separate name (Gramps.js, "the Gramps Web frontend"), but we renamend the repository to `gramps-web` because it lead to confusion.

From the point of view of the combined web application, it would make sense to have a monorepo and combined packaging, but so far we still have them separated. Since the API can be used for other things as well, I think it kind of makes sense. And for proper web-based deployments, it's anyway more complicated (celery, nginx, etc.).

What we could do is to add a convenience command to download a frontend.

DavidMStraub:
> Does it ever make sense to run the gramps-web-api server without a frontend? I guess for testing. Anything else? Maybe that is just an error, without something like --server-only?

In principle, it could serve as backend of various different apps. Somebody could develop some tailored frontends, say with some dedicated forms for entering census data; or for native mobile apps. That was the original idea when we started the API without a frontend.

It can also be used as a way to do batch operations on a Gramps tree without using Python code, just HTTP request.

gramps_webapi/__main__.py
dsblank: I added a note to explain.

gramps_webapi/__main__.py
dsblank: I'd like to make a gramps-web-desktop that is pip installable, and runs from the command line with easy options, ready to go. The frontend could be included in repo, or downloaded. Which would you prefer?

gramps_webapi/translogger.py
dsblank: Updated!

gramps_webapi/__main__.py
dsblank: Done! Also did that to `--debug-level`

dsblank:
> In principle, it could serve as backend of various different apps. Somebody could develop some tailored frontends, say with some dedicated forms for entering census data; or for native mobile apps. That was the original idea when we started the API without a frontend.

I see. As long as they have `GRAMPSWEB_STATIC_PATH` set, then the frontend could be anything.

> It can also be used as a way to do batch operations on a Gramps tree without using Python code, just HTTP request.

Whoa! Interesting!

gramps_webapi/__main__.py
dsblank: Changed the message to use gramps-web-api.

dsblank:
Ok, I think this PR is ready for final review. Feel free to make suggestions, or pushes onto this PR.

gramps_webapi/__main__.py
DavidMStraub: > The frontend could be included in repo, or downloaded. Which would you prefer?

With the current development setup and release cycles, I think downloading is the only option as the frontend is released more frequently, but let's keep this for a separate PR.

By the way - in my opinion, a usable desktop-like Gramps Web needs a background task worker (but that's also a separate discussion).

gramps_webapi/__main__.py
DavidMStraub: Did you forget to commit those changes? I don't see them in the final PR.

dsblank: I used the [click.Choice](https://github.com/gramps-project/gramps-web-api/pull/576/files#diff-fdcef86582d5df066774fb9feb6f20ca30750f97d63064aebec9e4246c7d0e3cR69), but I didn't want to make the choices be "1" or "2" but "window" or "tab", so still need the dict. But the value is guaranteed to be in dict.

dsblank:
> With the current development setup and release cycles, I think downloading is the only option as the frontend is released more frequently, but let's keep this for a separate PR.

I have an automated the downloading (to appear in a new repo probably today). We can put it in the repo as a build step, or in the start-up command. Either way.

> By the way - in my opinion, a usable desktop-like Gramps Web needs a background task worker (but that's also a separate discussion).

I tried this version (using waitress) and a large family tree (40,000 people) and the indexing worked without crashing. I hope to find a solution to this in general though through either an architecture redesign, or a easy task-server. I don't know enough about the issue yet.

gramps_webapi/__main__.py
DavidMStraub: Must have overlooked it, thanks

DavidMStraub:
> I tried this version (using waitress) and a large family tree (40,000 people) and the indexing worked without crashing. I hope to find a solution to this in general though through either an architecture redesign, or a easy task-server. I don't know enough about the issue yet.

We have celery - no redesign needed. My point was just that it's not a simple `python -m gramps_webapi ...` to start the app in that case because it's more like a microservice architecture.

DavidMStraub:
It would be great if you could update the documentation according to the new features at https://github.com/gramps-project/gramps-web-docs.

dsblank:
> My point was just that it's not a simple python -m gramps_webapi ... to start the app in that case

That was my point too :)

dsblank:
> It would be great if you could update the documentation

Done!

PR #590, opened 2024-12-13
Add metadata_encoding to unzip when import media archive

shizacat:
If the imported media archive contains non-ascii characters in the file names, we get error
`OSError: [Errno 36] File name too long: '/tmp/tmpuattkcyw/Any/╨£╨░╤Ç'`

It might be better to explicitly set the encoding for the archive.
With this fix, the import was successful.


DavidMStraub:
This parameter was added in Python 3.11, so your fix will not run on all supported Python versions (3.9+).

Can you check if there is a setting in you ZIP software to correctly specify the encoding?

DavidMStraub:
Thanks a lot thought for investigating and proposing a fix - let me know if you find a different way.

PR #593, opened 2024-12-28
Optionally process transactions in background

### LINKED ISSUE #592: Optionally process transactions in the background
The transactions endpoint is used for raw database transactions, which is used in particular by the sync addon (but is also required for the future revert option #511). Right now, this is done synchronously. But it includes updating the full-text and semantic search indexes. For a large number of transactions (e.g. syncing very out-of-sync trees), this can take very long, in particular with semantic search enabled, so we should process it in the background. For backwards compatibility, background processing should be optional.

DavidMStraub:
Fixes #592.


dsblank:
Nice! I'll take a look at this and try it.

Issue #583, opened 2024-12-08
Speed up relationship calculation


DavidMStraub:
Relationship calculation is currently too slow because it uses a large number of database queries. It should be relatively easy to get rid of this bottleneck by loading the relevant part of the database into memory.

DavidMStraub:
Fixed by #598 - further improvement likely requires caching.

PR #598, opened 2025-01-02
Add cached people/families proxy DB to speed up relationship calculation

DavidMStraub:
This might be enough to solve #583: this adds a proxy database class that allows to fetch all people and families in the database and cache them, making subsequent get-by-handle calls much faster, since they don't require a roundtrip to the database. The cost for fetching all people and families is relatively low since it uses a single database call each on SQLite and Postgres backends.

This strategy can/should lead to an enormous speedup when calculating relationships, especially when using PostgreSQL, because without this caching, the Gramps implementation of the relationship calculation can issue thousands of individual database calls.

### DIFF
--- a/gramps_webapi/api/resources/dna.py
+++ b/gramps_webapi/api/resources/dna.py
@@ -3,7 +3,7 @@
 #
 # Copyright (C) 2020       Nick Hall
 # Copyright (C) 2020-2023  Gary Griffin
-# Copyright (C) 2023       David Straub
+# Copyright (C) 2023-2025  David Straub
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as published by
@@ -32,6 +32,8 @@ from gramps.gen.relationship import get_relationship_calculator
 from gramps.gen.utils.grampslocale import GrampsLocale
 from webargs import fields, validate
 
+from gramps_webapi.api.people_families_cache import CachePeopleFamiliesProxy
+
 from ...types import Handle
 from ..util import get_db_handle, get_locale_for_language, use_args
 from .util import get_person_profile_for_handle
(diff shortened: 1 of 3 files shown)


dsblank:
Does this have an impact on memory consumption?

DavidMStraub:
It does consume more memory, but negligible in practice. (Even trees with tens of thousands of people will consume only a couple of MB of RAM.)

DavidMStraub:
... by the way, the way I've written the proxy class, the objects are not cached on instantiation - the idea is that this could be used in different contexts and the cache is only created when it is assumed that a large number of individual `get_X_by_handle` calls would otherwise be made (which is the case for relationship calcuations or sorting/filtering).

dsblank:
Does the cache need to be cleared when an item is edited or deleted? Is this similar to the [cache proxy in core gramps](https://github.com/gramps-project/gramps/blob/master/gramps/gen/proxy/cache.py)? I didn't add cache breaking to it, but could.

DavidMStraub:
No, it's not necessary because it only lives during a single request. It also operates on a read-only `DbReadBase` which cannot be written to. We write to the database typically in `POST`, `PUT` or `DELETE` requests, while this is meant for speeding up slow read queries.

dsblank:
> while this is meant for speeding up slow read queries.

Yeah, that is what the `CacheProxyDb` was created for too.

DavidMStraub:
Yes, this is very much motivated by `CacheProxyDb`. The reason it didn't help here is that it caches repeated calls to the same objects, but that's not the bottleneck here, the bottleneck is having many individual `SELECT`s for individual objects rather than a single `SELECT` to read the whole table.

dsblank:
I see... yeah, makes sense. Nice!

gramps_webapi/api/resources/dna.py
```diff
@@ -3,7 +3,7 @@
 #
 # Copyright (C) 2020       Nick Hall
 # Copyright (C) 2020-2023  Gary Griffin
-# Copyright (C) 2023       David Straub
+# Copyright (C) 2023-2025  David Straub
```
wrenix, line 6: Why not the reuse Format?
https://reuse.software/spec-3.3/#comment-headers

DavidMStraub: Because I wasn't aware of it - this is borrowed from how the Gramps project does it. To me (as someone who prefers more permissive licensess) this is mostly a nuisance - if we can automate it, I would be happy.

PR #600, opened 2025-01-06
Remove import from torch in export.py

dsblank:
This PR removes a spurious import from torch, which crashes if torch is not installed.


DavidMStraub:
Thanks!

I blame copilot, it sometimes does that :grimacing: should probably configure the type checker to fail on unused imports.

PR #601, opened 2025-01-06
Add "has association type" person filter rule

DavidMStraub:
Related to https://github.com/gramps-project/gramps-web/issues/544: this allows filtering all people that have an association of type "DNA", which is used for DNA match data.


DavidMStraub:
Rebase issue, will try again

PR #608, opened 2025-01-17
Format and return alternate_place_names

cajturner:
Return place name value and formatted date string

Added based on discussion here: https://github.com/gramps-project/gramps-web/pull/566

### DIFF
--- a/tests/test_endpoints/test_places.py
+++ b/tests/test_endpoints/test_places.py
@@ -504,6 +504,7 @@ class TestPlacesHandle(unittest.TestCase):
             rv["profile"],
             {
                 "alternate_names": [],
+                "alternate_place_names": [],
                 "gramps_id": "P0860",
                 "lat": 33.6259414,
                 "long": -97.1333453,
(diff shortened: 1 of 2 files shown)


tests/test_endpoints/test_places.py
```diff
@@ -513,13 +513,15 @@ def test_get_places_handle_parameter_profile(self):
             rv["profile"],
             {
                 "alternate_names": [],
+                "alternate_place_names": [],
```
cajturner, line 507: I wasn't sure how the test data is created. How can I create a Place with alt names?

cajturner: @DavidMStraub can you point me in the right direction for setting up new test data?

cajturner: Ah from the gramps example db. I will update the tests

DavidMStraub:
Can you please ping me when it's ready to merge? I'll convert to draft in the meantime.

cajturner:
> Can you please ping me when it's ready to merge? I'll convert to draft in the meantime.

Yes, will do. I have a new born baby that has just arrived so could be a few weeks before I can look

DavidMStraub:
Congrats on that family tree extension! :laughing: :tada:

cajturner:
@DavidMStraub Finally got some time to add a test 😅

Issue #614, opened 2025-02-02
Link in Mail-Adress confirmation message incomplete on Apple mail


DanielMorsch:
Hi,

just noticed that the link for the mail-adress confirmation message or the Password reset message is incomplete.

it seems as if the Apple mail app on macOS or iOS gets stuck at a point in the jwt token and thus only half of the link is really detected as a link.

Like this one:

<img width="446" alt="Image" src="https://github.com/user-attachments/assets/81070394-7123-4d2d-b513-579a714ecedf" />

On Gmail the same mail looks good. The whole link is detected.

Would it be possible to design the e-mail with a little html code? Then a button could be created to go directly to the link. At the moment I find the emails a bit too “technical”.

DavidMStraub:
Yes I think that's a good idea!

Do you want to give it a shot? Actually, I have some code for Grampshub confirmation e-mails that we can directly use.

We need to add to the `send_email` function:

- new argument `body_html: str | None = None`
- Replace `msg.set_content(body)` by

```python
    msg.set_content(body)
    if body_html:
        msg.add_alternative(body_html, subtype="html")
```

And then the various functions like `email_confirm_email` have to be adapted - probably we can return a tuple of `body, body_html`.

A snippet to reuse for an HTML button:

```python

def url_as_button(url: str, button_label: str) -> str:
    return f"""
        <a href="{url}" style="background-color: #6D4C41; color: white;
            padding: 10px 20px; text-decoration: none; border-radius: 4px;
            font-family: sans-serif;">
            {button_label}
        </a>
    """

body = f"{text}\n\n{url}"
body_html = f"""
<p>{text}</p>
<p style="text-align: center;">
{url_as_button(url, "some_button_label")}
</p>
"""
```

DavidMStraub:
Fixed in #615

PR #615, opened 2025-02-05
New mail layout and fix incomplete link in mail

DanielMorsch:
Hi,

I have completely revised the layout for the two mails “Confirm e-mail” and “Reset Password”. Corresponding to #614 
The mails now also contain the user names of the users.

I have also added a new cli cmd for testing (it was a little quicker to test during the development phase). But we can also throw it out again.

The new task `send_dummy_email` in `tasks.py` could be adapted a little and then placed behind an API endpoint. Then nothing would stand in the way of a mail test section in the frontend.

### DIFF
--- a/gramps_webapi/api/emails.py
+++ b/gramps_webapi/api/emails.py
@@ -21,86 +21,6 @@
 
 from gettext import gettext as _
 
-
-def email_reset_pw(base_url: str, user_name: str, token: str):
-    """Reset password e-mail text."""
-    intro = _(
-        "You are receiving this e-mail because you (or someone else) "
-        "have requested the reset of the password for your account."
-    )
-
-    action = _(
-        "Please click on the following link, or paste this into your browser "
-        "to complete the process:"
-    )
-
-    end = _(
-        "If you did not request this, please ignore this e-mail "
-        "and your password will remain unchanged."
-    )
-
-    url = f"""{base_url}/api/users/-/password/reset/?jwt={token}"""
-    body = f"""{intro}
-
-{action}
-
-{url}
-
-{end}
-"""
-
-    body_html = email_htmlbody_reset_pw(user_name=user_name, url=url)
-    return (body, body_html)
-
-
-def email_confirm_email(base_url: str, user_name: str, token: str):
-    """Confirm e-mail address e-mail text."""
-    intro = (
-        _(
-            "You are receiving this message because this e-mail address "
-            "was used to register a new account at %s."
-        )
-        % base_url
-    )
-    action = _(
-        "Please click on the following link, or paste this into your browser "
-        "to confirm your email address. You will be able to log on once a "
-        "tree owner reviews and approves your account."
-    )
-
-    url = f"""{base_url}/api/users/-/email/confirm/?jwt={token}"""
-
-    body = f"""{intro}
-
-{action}
-
-{url}
-"""
-    body_html = email_htmlbody_confirm_email(user_name=user_name, url=url)
-    return (body, body_html)
-
-
-def email_new_user(base_url: str, username: str, fullname: str, email: str):
-    """E-mail notifying owners of a new registered user."""
-    intro = _("A new user registered at %s.") % base_url
-    next_step = _(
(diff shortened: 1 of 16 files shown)


gramps_webapi/api/emails.py
```diff
@@ -63,20 +68,23 @@ def email_confirm_email(base_url: str, token: str):
...
+    return (body, body_html)
 
 
 def email_new_user(base_url: str, username: str, fullname: str, email: str):
     """E-mail notifying owners of a new registered user."""
     intro = _("A new user registered at %s.") % base_url
     next_step = _(
-        "Please review this user registration and assign a role to "
-        "enable access:"
+        "Please review this user registration and assign a role to " "enable access:"
```
dstraub-pf, line 87: We can make this a single string now that it's on a single line - not sure why it wasn't.

gramps_webapi/api/emails.py
```diff
@@ -91,3 +99,139 @@ def email_new_user(base_url: str, username: str, fullname: str, email: str):
 
 {user_details}
 """
+
+
+def email_html_styles() -> str:
```
dstraub-pf, line 104: Let's make this a constant `EMAIL_HTML_STYLES`

dstraub-pf: CSS styles perhaps...

gramps_webapi/api/emails.py
DavidMStraub, line 86: We can make this a single string now that it's on a single line - not sure why it wasn't.

gramps_webapi/api/emails.py
DavidMStraub: Let's make this a constant `EMAIL_CSS_STYLES`

gramps_webapi/api/emails.py
```diff
@@ -91,3 +99,139 @@ def email_new_user(base_url: str, username: str, fullname: str, email: str):
...
+def email_htmlbody_reset_pw(user_name: str, url: str) -> str:
+    header = _("Reset Your Password")
+    greeting = _("Hi %s,") % user_name
+    descMail = _(
+        "You are receiving this e-mail because you (or someone else) have requested the reset of the password for your account."
+    )
+    descAction = _("Click the button below to set a new password:")
+    buttonLabel = _("Reset Password")
+    descIgnore = _(
+        "If you did not request a password reset, please ignore this email and your password will remain unchanged."
```
DavidMStraub, line 161: This string is different from the one used in the text-only e-mail. Can you please avoid repetition and make the HTML e-mail a styled version, but with exactly the same text? (Except click the button...)

gramps_webapi/api/emails.py
```diff
@@ -91,3 +99,139 @@ def email_new_user(base_url: str, username: str, fullname: str, email: str):
...
+    )
+    return """
```
DavidMStraub: Let's Make the HTML boilerplate a string constant as well.

DavidMStraub:
Beautiful! :tada: 

Concerning the CLI API: rather than `python -m gramps_webapi system sendMail user@example.com`, what I would find much more useful is:

- `python -m gramps_webapi email reset-pw user@example.com someusername`
- `python -m gramps_webapi email confirm-email user@example.com someusername`

This would be a very simple change of your code but would make this both a test utility (with dummy data) and an administration tool (to manually trigger such emails for whatever reason). What do you think?

gramps_webapi/api/emails.py
DanielMorsch: If that is ok, I would then adapt the unstyled version.

DavidMStraub:
Can you ping me when you think it's ready? Thanks!

DanielMorsch:
Yes, of course. The unit tests have not yet been adapted. 
I have a few problems running the unit tests on my Mac. But I think I have now discovered what I need to adjust. 
I'll let you know when I've adapted the unit tests.

DanielMorsch:
Hi @DavidMStraub,

the unit tests should now run again. From my side it is now ready.

DavidMStraub:
Looks good! :rocket:

Issue #616, opened 2025-02-05
Update JSON serialization methods for Gramps 6.0


DavidMStraub:
@dsblank would you mind summarizing briefly the changes made in https://github.com/gramps-project/gramps/pull/1876 to Gramps core?

Gramps Web API uses the existing `gramps.gen.lib.serialize` which has been removed, so it will be some work changing it.

dsblank:
Can you point to an example in the gramps-web-api that uses the methods? I don't mind doing the update myself.

BTW, `gramps.gen.lib.serialize` hasn't been removed, and all of the functions are still there (renamed) if you have a DB instance. Otherwise, you can use the underlying methods in `gramps.gen.lib.json_utils`.

DavidMStraub:
Thanks!

Here are some examples:

- UndoDb https://github.com/gramps-project/gramps-web-api/blob/master/gramps_webapi/undodb.py#L82-L93
- Transactions as JSON https://github.com/gramps-project/gramps-web-api/blob/master/gramps_webapi/undodb.py#L82-L93
- Parsing JSON objects https://github.com/gramps-project/gramps-web-api/blob/master/gramps_webapi/api/resources/base.py#L152-L167

dsblank:
Thanks! I'll start a PR. From what I've seen, these bits should run faster.

PR #617, opened 2025-02-05
Update serialize operations for Gramps 6.0

dsblank:
WIP: this is a first draft of updating the serialization code to new Gramps 6.0 data representation.

I don't understand all of the code changed here, and I don't know what kinds of exceptions were caught (and why) and probably the new json_utils will raise different exceptions.

### DIFF
--- a/gramps_webapi/api/resources/tasks.py
+++ b/gramps_webapi/api/resources/tasks.py
@@ -19,14 +19,14 @@
 
 """Background task resources."""
 
-import json
+#import json
 from http import HTTPStatus
 
 from celery.result import AsyncResult
 from flask import abort
 
 from . import ProtectedResource
-
+from gramps.gen.lib.json_utils import object_to_string
 
 class TaskResource(ProtectedResource):
     """Resource for a single task."""
@@ -39,13 +39,14 @@ class TaskResource(ProtectedResource):
 
         def serialize_or_str(obj):
             try:
-                return json.dumps(obj)
+                return object_to_string(obj) # json.dumps(obj)
             except TypeError:
                 return str(obj)
 
         def serializable_or_str(obj):
             try:
-                json.dumps(obj)
+                #json.dumps(obj)
+                object_to_string(obj)
                 return obj
             except TypeError:
                 return str(obj)
--- a/gramps_webapi/api/resources/base.py
+++ b/gramps_webapi/api/resources/base.py
@@ -31,7 +31,8 @@ from gramps.gen.db import DbTxn
 from gramps.gen.db.base import DbReadBase
 from gramps.gen.errors import HandleError
 from gramps.gen.lib.primaryobj import BasicPrimaryObject as GrampsObject
-from gramps.gen.lib.serialize import from_json
+#from gramps.gen.lib.serialize import from_json
+from gramps.gen.lib.json_utils import dict_to_string
 from gramps.gen.utils.grampslocale import GrampsLocale
 from pyparsing.exceptions import ParseBaseException
 from webargs import fields, validate
@@ -164,7 +165,7 @@ class GrampsObjectResourceHelper(GrampsJSONEncoder):
             abort_with_message(400, f"Error while processing object: {exc}")
         if not validate_object_dict(obj_dict):
             abort_with_message(400, "Schema validation failed")
-        return from_json(json.dumps(obj_dict))
+        return dict_to_string(obj_dict) # json.dumps(obj_dict))
 
     def has_handle(self, handle: str) -> bool:
         """Check if the handle exists in the database."""
--- a/gramps_webapi/api/resources/objects.py
+++ b/gramps_webapi/api/resources/objects.py
@@ -27,7 +27,8 @@ from flask_jwt_extended import get_jwt_identity
 from gramps.gen.db import DbTxn
 from gramps.gen.lib import Family, Person
 from gramps.gen.lib.primaryobj import BasicPrimaryObject as GrampsObject
-from gramps.gen.lib.serialize import from_json
+#from gramps.gen.lib.serialize import from_json
+from gramps.gen.lib.json_utils import data_to_object
 from webargs import fields, validate
 
 from ...auth.const import PERM_ADD_OBJ, PERM_DEL_OBJ_BATCH, PERM_EDIT_OBJ
@@ -68,7 +69,7 @@ class CreateObjectsResource(ProtectedResource):
                 abort_with_message(400, "Error processing objects")
             if not validate_object_dict(obj_dict):
                 abort_with_message(400, "Validation error while processing objects")
-            obj = from_json(json.dumps(obj_dict))
+            obj = data_to_object(obj_dict) # json.dumps(obj_dict))
             objects.append(obj)
         return objects
 
--- a/gramps_webapi/api/resources/util.py
+++ b/gramps_webapi/api/resources/util.py
@@ -54,7 +54,8 @@ from gramps.gen.lib import (
     Span,
 )
 from gramps.gen.lib.primaryobj import BasicPrimaryObject as GrampsObject
-from gramps.gen.lib.serialize import to_json
+#from gramps.gen.lib.serialize import to_json
+from gramps.gen.lib.json_utils import string_to_dict, object_to_string
 from gramps.gen.plug import BasePluginManager
 from gramps.gen.relationship import get_relationship_calculator
 from gramps.gen.soundex import soundex
@@ -1122,16 +1123,17 @@ def transaction_to_json(transaction: DbTxn) -> TransactionJson:
             continue  # this happens for references
         trans_dict = {TXNUPD: "update", TXNDEL: "delete", TXNADD: "add"}
         obj_cls = getattr(gramps.gen.lib, obj_cls_name)
-        if old_data:
-            old_data = obj_cls().unserialize(old_data)
-        if new_data:
-            new_data = obj_cls().unserialize(new_data)
+        # In Gramps 6.0, the data repr is a string
+        #if old_data:
+        #    old_data = obj_cls().unserialize(old_data)
+        #if new_data:
+        #    new_data = obj_cls().unserialize(new_data)
         item = {
             "type": trans_dict[action],
             "handle": handle,
             "_class": obj_cls_name,
-            "old": json.loads(to_json(old_data)),
-            "new": json.loads(to_json(new_data)),
+            "old": string_to_dict(old_data), # json.loads(to_json(old_data)),
+            "new": string_to_dict(new_data), # json.loads(to_json(new_data)),
         }
         out.append(item)
     return out
@@ -1155,7 +1157,7 @@ def reverse_transaction(transaction_list: TransactionJson) -> TransactionJson:
 
 def hash_object(obj: GrampsObject) -> str:
     """Generate a SHA256 hash for a Gramps object's data."""
-    data = to_json(obj).encode()
+    data = object_to_string(obj).encode() #to_json(obj).encode()
     return sha256(data).hexdigest()
 
 
--- a/gramps_webapi/api/tasks.py
+++ b/gramps_webapi/api/tasks.py
@@ -30,7 +30,8 @@ from celery import shared_task, Task
 from celery.result import AsyncResult
 from flask import current_app
 from gramps.gen.db import DbTxn
-from gramps.gen.lib.serialize import from_json, to_json
+#from gramps.gen.lib.serialize import from_json, to_json
+from gramps.gen.lib.json_utils import string_to_object, object_to_dict
 from gramps.gen.db.base import DbReadBase
 from gramps.gen.errors import HandleError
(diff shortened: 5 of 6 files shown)


dsblank:
In reference to #616

gramps_webapi/api/resources/tasks.py
```diff
@@ -39,13 +39,14 @@ def get(self, task_id: str):
 
         def serialize_or_str(obj):
             try:
-                return json.dumps(obj)
+                return object_to_string(obj) # json.dumps(obj)
```
DavidMStraub: This module is quite sloppy because it doesn't use any type hints - in this function, `obj` is not a Gramps object, so there is no need to change it, we can leave `json.dumps`.

DavidMStraub:
There seems to still be a `from _future__ import annotations` missing somewhere in Gramps where pipes are used in annotations.

DavidMStraub:
Found the cause and will submit a PR to Gramps core soon.

DavidMStraub:
Fixed in https://github.com/gramps-project/gramps/pull/1930

gramps_webapi/api/resources/tasks.py
DavidMStraub:
```suggestion
                return json.dumps(obj)
```

gramps_webapi/api/resources/tasks.py
```diff
@@ -19,14 +19,14 @@
...
-import json
+#import json
 from http import HTTPStatus
 
 from celery.result import AsyncResult
 from flask import abort
 
 from . import ProtectedResource
-
+from gramps.gen.lib.json_utils import object_to_string
```
DavidMStraub, line 29:
```suggestion

```

gramps_webapi/api/resources/tasks.py
dsblank: All of the json methods have been replaced with optimized versions. If obj is actually a dict, then you should use `dict_to_string` (which uses orjson in this case, and is much faster). But anyway, we want to hide the details in these methods as they are all optimized and will always be the best methods of conversion, even if the internal representation changes going forward.

gramps_webapi/api/resources/tasks.py
dsblank: Probably want to change that to `dict_to_string`

pyproject.toml
```diff
@@ -16,7 +16,8 @@ classifiers = [
 ]
 requires-python = ">=3.9"
 dependencies = [
-    "gramps[GUI,i18n]==5.2.*",
+    "gramps[GUI,i18n]==6.0.0b1",
```
DavidMStraub: We can try with `"gramps[all] @ git+https://github.com/gramps-project/gramps.git@maintenance/gramps60"` now that the fixes have been merged into this branch.

DavidMStraub:
@dsblank would you mind giving me push access to the PR branch? I'd like to try if tests pass with the fixes upstream in Gramps.

dsblank:
@DavidMStraub done!

DavidMStraub:
I fixed it by releasing a new version of `gramps-ql`, now it fails due to `object-ql` :laughing: @dsblank the ball is in your court again.

Btw, in `gramps-ql` I decided to stay compatible with Gramps 5 and 6 for now:

https://github.com/DavidMStraub/gramps-ql/blob/main/src/gramps_ql/gql.py#L18-L29

dsblank:
object-ql is now updated to version 0.1.3

DavidMStraub:
@dsblank I see to problems with `object_to_string`:

1. The type hint is wrong. It claims to return `str | bytes` but always returns `bytes`. If the type hint is taken at face value, all callers should handle both types, which doesn't make sense.
2. The name is misleading. IMO it should either return `str` or be named `object_to_byte_string`

DavidMStraub:
What could cause these errors in unit tests?

```python
    def has_handle(
        db_handle: DbWriteBase,
        obj: GrampsObject,
    ) -> bool:
        """Check if an object with the same class and handle exists in the DB."""
        obj_class = obj.__class__.__name__.lower()
        method = db_handle.method("has_%s_handle", obj_class)
>       return method(obj.handle)
E       AttributeError: 'Person' object has no attribute 'handle'
```

dsblank:
> What could cause these errors in unit tests?

That is a weird one. No idea. (If that is proxy code, I removed those methods in #1952 for Gramps 6.1).

DavidMStraub:
> However, the blob serializer does still return a string for these methods.

I don't think that's true. The blob serializer uses `pickle.dumps` which returns `bytes`.

Question: why does this actually work when using `orjson`?

https://github.com/gramps-project/gramps/blob/maintenance/gramps60/gramps/plugins/db/dbapi/dbapi.py#L669-L675

You are now writing a `bytes` obejct to the JSON field of type `TEXT`, aren't you?

DavidMStraub:
> Oh wow... relying on such low-level details seem like poor design. But point me to the code and I'll see what is the issue.

Well - what the code relies on is that Proxy databases and non-proxy databases have a common set of methods and functions. You can look at
https://github.com/search?q=repo%3Agramps-project%2Fgramps-web-api%20dbreadbase&type=code
to see how `DbReadBase` is used as a type hint.

Of course as long as Gramps isn't properly type hinted, it's easy to just remove something... :roll_eyes: but equally easy to run into refactoring bugs.

dsblank:
> What I meant is that the _to_string methods should be consistently either return str or bytes, and the DBAPI can cast to the appropriate type if/where needed.

As far as I know, no change in type is needed.

dsblank:
> `DbReadBase` is used as a type hint.

Probably using the "protocol" API is better than using a class anyway, don't you think?

If it is a blocker, we can reinstate `ProxyBase` as a subclass of `DbReadBase` ... it just means implementing a few (dozen) methods that do nothing but call the instance methods.

DavidMStraub:
> Probably using the "protocol" API is better than using a class anyway, don't you think?

Yes.

> As far as I know, no change in type is needed.

Ok, I was just surprised passing a `byte` into an `execute` placeholder that expects a string works. For `psycopg2` it's not clear to me from the [documentation](https://www.psycopg.org/docs/usage.html#adapt-string) either. 

For a Python f-string, it would not work:

```python
bs = "bs".encode()
print(f"{bs}")
# will print:
"b'bs'"
```

DavidMStraub:
Investigated the remaining CI failures.

Unfortunately, `dict_to_object` is not a replacement for the removed `from_json` function in Gramps. It expects all the object properties to be present as dict keys, otherwise it raises a `KeyError`. That doesn't make any sense for Web API as of course objects usually only contain the non-empty keys.

So what I will do is I'll add back the Gramps 5.0 `from_json` function in Web API. It would be better in core, but ok.

@Nick-Hall for visibility.

dsblank:
> So what I will do is I'll add back the Gramps 5.0 from_json function in Web API. It would be better in core, but ok.

What would be better is to have a better understanding of what the issue is, and fix that.

DavidMStraub:
Agreed.

Minimal example:

```python
from gramps.gen.lib.json_utils import data_to_object

data= {
    "_class": "Note",
    "gramps_id": "N0001",
    "text": {"_class": "StyledText", "string": "Hello Gramps."},
}
data_to_object(data)
```

```
File ~/.venv/gramps6/lib/python3.12/site-packages/gramps/gen/lib/styledtext.py:112, in StyledText.set_object_state(self, attr_dict)
    105 def set_object_state(self, attr_dict):
    106     """
    107     Set the current object state using information provided in the given
    108     dictionary.
    109 
    110     We override this method to handle the `tags` and `string` properties.
    111     """
--> 112     self._tags = attr_dict["tags"]
    113     self._string = attr_dict["string"]

KeyError: 'tags'
```

dsblank:
That seems like a valid error, based on the design. Here `data` isn't a fully formed data dict of a Note. What else could be wrong or missing? The conversion to/from objects doesn't have any error checking, as these are created for fast internal conversions.

Perhaps you want something like:

```python
note = Note()
data = object_to_data(note)
data["gramps_id"] = "N0001"
data["text"]["string"] = "Hello Gramps."
```

Note that object_to_data() may return a "_object" key with the original object cached. If you don't want that, use `object_to_dict()`

Also note that you can't do assignment using the dot attr API.

```python
# DO NOT DO THIS!
data.gramps_id = "N0001"
data.text.string = "Hello Gramps."
```

DavidMStraub:
> The conversion to/from objects doesn't have any error checking, as these are created for fast internal conversions.

Yes, but that doesn't apply to the old `from_json`, `to_json`.

Gramps Web API is a REST API that relies on passing JSON objects around. Converting objects to/from JSON is one of the core tasks. Until Gramps 5.2, this relied on functions in Gramps core. Now these functions have been removed. So I copied them over. Still can't see what's wrong with that, although I am always unhappy moving things from core to Web API, as it increases the maintenance burden (such as for this upgrade).

dsblank:
> Gramps Web API is a REST API that relies on passing JSON objects around. Converting objects to/from JSON is one of the core tasks. 

I think you can say that about Gramps core now, even more so!

> Until Gramps 5.2, this relied on functions in Gramps core. Now these functions have been removed. So I copied them over. 

@Nick-Hall how do you feel about creating objects from dicts that have missing keys? I think it is fine for `gramps-web-api` to have specialized JSON functions for that. Although not sure how such objects will behave (if interacting with gramps core) if they are missing attributes?

DavidMStraub:
Down to 22 errors now.

Do you know what could cause this one, related to filters? Perhaps related to the filter refactoring?

```
gramps_webapi/api/resources/filters.py:180: in apply_filter
    filter_object = build_filter(filter_parms, namespace)
gramps_webapi/api/resources/filters.py:138: in build_filter
    filter_object.set_logical_op(filter_parms["function"])
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <gramps.gen.filters._genericfilter.GenericEventFilter object at 0x7f7338d9e450>
val = 'xor'

    def set_logical_op(self, val):
        if val in GenericFilter.logical_functions:
            self.logical_op = val
        else:
>           raise Exception("invalid operator: %r" % val)
E           Exception: invalid operator: 'xor'
```

Nick-Hall:
> how do you feel about creating objects from dicts that have missing keys? I think it is fine for gramps-web-api to have specialized JSON functions for that. Although not sure how such objects will behave (if interacting with gramps core) if they are missing attributes?

The objects shouldn't be missing any attributes.  The legacy function created the object and then called `setattr` in a loop to assign the attributes.  This was identified as a performance bottleneck by @prculley in my original prototype.

There may be some subtle differences since the new code calls `set_object_state` and the legacy code will call setters such as `Person.set_gender`.  You will also need to check if tuples and lists are handled differently now.

dsblank:
> The objects shouldn't be missing any attributes. 

That is clarifying. The issue is in code like:

```python
def set_object_state(self, attr_dict):
    self._tags = attr_dict["tags"]
    self._string = attr_dict["string"]
```

where it doesn't check attr_dict before accessing "tags". Should we fix those?

Nick-Hall:
No.  If you use the newer functions the problem is that [set_object_state](https://github.com/gramps-project/gramps/blob/master/gramps/gen/lib/baseobj.py#L100) requires a full dictionary so it can assign every attribute in a single update.

The legacy code can accept a partial dictionary because it first creates the object and then sets the attributes in a loop with `setattr`.

We traded off the flexibility of being able to use partial dictionaries for a performance gain.

DavidMStraub:
> I knew that the Gramps UI had removed the xor option, so I did too. Do you need it?

Not sure. What was the reason for removing it?

dsblank:
> Not sure. What was the reason for removing it?

There used to be a logical connection named "xor" but I think they renamed it "one" (only one rule should apply) which replaces it.

DavidMStraub:
But in `_genericfilter.py`, in Gramps 5.2 we have

```python
logical_functions = ["and", "or", "xor", "one"]
```
while in Gramps 6.0 we now have

```python
logical_functions = ["and", "or", "one"]
```

So "one" and "xor" are distinct and "xor" was removed.

I don't mind too much, but its clearly a breaking change.

Nick-Hall:
It looks like "xor" is just a deprecated synonym for "one".  I see no problem with removing it.  We could add an upgrade to the filter parser to be safe.

DavidMStraub:
I removed xor.

There are still several tests failing.

Also the undo database will need a schema upgrade.

Is end of Feb still a realistic date for Gramps 6.0 release?

DavidMStraub:
There is an annoying error now in the undo database because `string_to_data` fails if the undo object is a reference rather than a Gramps object, which is serialized to a list `[obj_handle, obj_class, ref_handle, ref_class]`. `string_to_data` passes this to `DataDict` which throws an error.

DavidMStraub:
I commented out the last failing test (database migration) and rebased onto a new branch to be able to iterate in the main repo.

Issue #673, opened 2025-08-11
Add name format parameter to people endpoint


DavidMStraub:
... which will determine the name format to use in the person profile `name_display` key.

@vktimofeev it would be great if you could implement this!

vktimofeev:
My first question is what would be the profile["name_display"] value if the name format parameter is absent in the query?

Btw the name format is contained in the Name class (gramps.gen.lib.name) instances as "display_as" field which value is not a format string itself but rather an index of desirable format. In the recent versions of Gramps, the list of desirable formats is kept in a record within the **metadata** table of the database having the **settings** field value equal to "name_formats" (formerly name formats configuration was stored by default in a GUI config file). So the displayer  - a NameDisplay class instance instantiated in the gramps.gen.display.name, can produce the name in the format predefined in the database, and therefore name format query string parameter is not even necessary to produce a reasonable value for  name_display key.

Maybe it wasn't necessary to pronounce all this, but just code it right away...

DavidMStraub:
> Maybe it wasn't necessary to pronounce all this, but just code it right away...

I think it's super useful to pronounce all this, thanks a lot! 🙂 

Good point to distinguish between the numbered, predefined formats and the format strings.

What about the following approach, which mimics how Gramps desktop handles it:

1. `name_display` defaults to a name format defined in the database (answering your first question)
2. in the future, the Web API can support changing name formats, so the tree owner can influence the default display format
3. the Web API will support a query parameter that is a format string
4. the frontend will support a setting that let's users either select from one of the predefined formats in the database, or provide their own query string. In both cases, this is converted to the appropriate query string in API calls.

What do you think? For your use case, 1. & 2. might even be enough if you expect a single format to be prevalent in your tree?

DavidMStraub:
I'm away from the keyboard at the moment, so I cannot check but: is the current behaviour of the people view the same as or different from the people list in Gramps Desktop?

DavidMStraub:
Ok, if you want you can open a draft pull request with your work in progress code, then I can add comments.

vktimofeev:
Done! I hope, it's all right with the pull request itself: never did this before here on Github.

PR #674, opened 2025-08-12
A straightforward implementation of #673 : added webarg "name_format"…

vktimofeev:
…, a "name_display" field to the person profile, and initialization code using gramps.gen.display.name displayer

### DIFF
--- a/gramps_webapi/api/resources/util.py
+++ b/gramps_webapi/api/resources/util.py
@@ -334,6 +334,7 @@ def get_birth_profile(
     person: Person,
     args: Union[list, None] = None,
     locale: GrampsLocale = glocale,
+    name_format: str | None = None,
 ) -> tuple[dict, Union[Event, None]]:
     """Return best available birth information for a person."""
     event = get_birth_or_fallback(db_handle, person)
@@ -341,7 +342,9 @@ def get_birth_profile(
         return {}, None
     args = args or []
     return (
-        get_event_profile_for_object(db_handle, event, args=args, locale=locale),
+        get_event_profile_for_object(
+            db_handle, event, args=args, locale=locale, name_format=name_format
+        ),
         event,
     )
 
@@ -351,6 +354,7 @@ def get_death_profile(
     person: Person,
     args: Union[list, None] = None,
     locale: GrampsLocale = glocale,
+    name_format: str | None = None,
 ) -> tuple[dict, Union[Event, None]]:
     """Return best available death information for a person."""
     event = get_death_or_fallback(db_handle, person)
@@ -358,7 +362,9 @@ def get_death_profile(
         return {}, None
     args = args or []
     return (
-        get_event_profile_for_object(db_handle, event, args=args, locale=locale),
+        get_event_profile_for_object(
+            db_handle, event, args=args, locale=locale, name_format=name_format
+        ),
         event,
     )
 
@@ -368,6 +374,7 @@ def get_marriage_profile(
     family: Family,
     args: Union[list, None] = None,
     locale: GrampsLocale = glocale,
+    name_format: str | None = None,
 ) -> tuple[dict, Union[Event, None]]:
     """Return best available marriage information for a couple."""
     event = get_marriage_or_fallback(db_handle, family)
@@ -375,7 +382,9 @@ def get_marriage_profile(
         return {}, None
     args = args or []
     return (
-        get_event_profile_for_object(db_handle, event, args=args, locale=locale),
+        get_event_profile_for_object(
+            db_handle, event, args=args, locale=locale, name_format=name_format
+        ),
         event,
     )
 
@@ -385,6 +394,7 @@ def get_divorce_profile(
     family: Family,
     args: list | None = None,
     locale: GrampsLocale = glocale,
+    name_format: str | None = None,
 ) -> tuple[dict, Event | None]:
     """Return best available divorce information for a couple."""
     event = get_divorce_or_fallback(db_handle, family)
@@ -392,7 +402,9 @@ def get_divorce_profile(
         return {}, None
     args = args or []
     return (
-        get_event_profile_for_object(db_handle, event, args=args, locale=locale),
+        get_event_profile_for_object(
+            db_handle, event, args=args, locale=locale, name_format=name_format
+        ),
         event,
     )
 
@@ -531,24 +543,35 @@ def get_person_profile_for_object(
                 label="age",
                 locale=locale,
                 role=locale.translation.sgettext(event_ref.get_role().xml_str()),
+                name_format=name_format,
             )
             for event_ref in person.event_ref_list
         ]
     if "all" in args or "families" in args:
         primary_parent_family_handle = person.get_main_parents_family_handle()
         profile["primary_parent_family"] = get_family_profile_for_handle(
-            db_handle, primary_parent_family_handle, options, locale=locale
+            db_handle,
+            primary_parent_family_handle,
+            options,
+            locale=locale,
+            name_format=name_format,
         )
         profile["other_parent_families"] = []
         for handle in person.parent_family_list:
             if handle != primary_parent_family_handle:
                 profile["other_parent_families"].append(
                     get_family_profile_for_handle(
-                        db_handle, handle, options, locale=locale
+                        db_handle,
+                        handle,
+                        options,
+                        locale=locale,
+                        name_format=name_format,
                     )
                 )
         profile["families"] = [
-            get_family_profile_for_handle(db_handle, handle, options, locale=locale)
+            get_family_profile_for_handle(
+                db_handle, handle, options, locale=locale, name_format=name_format
+            )
             for handle in person.family_list
         ]
     return profile
@@ -654,6 +677,7 @@ def get_family_profile_for_object(
                 base_event=marriage_event,
                 label="span",
                 locale=locale,
+                name_format=name_format,
             )
             for event_ref in family.event_ref_list
         ]
--- a/gramps_webapi/api/resources/search.py
+++ b/gramps_webapi/api/resources/search.py
@@ -103,7 +103,11 @@ class SearchResource(GrampsJSONEncoder, ProtectedResource):
                 )
             elif class_name == "event":
                 obj.profile = get_event_profile_for_object(
-                    self.db_handle, obj, args["profile"], locale=locale
+                    self.db_handle,
+                    obj,
+                    args["profile"],
+                    locale=locale,
+                    name_format=args.get("name_format"),
                 )
             elif class_name == "citation":
                 obj.profile = get_citation_profile_for_object(


gramps_webapi/api/resources/people.py
```diff
@@ -49,9 +49,18 @@ def object_extend(
         """Extend person attributes as needed."""
         db_handle = self.db_handle
         if "profile" in args:
-            obj.profile = get_person_profile_for_object(
-                db_handle, obj, args["profile"], locale=locale
-            )
+            if "name_format" in args:
```
DavidMStraub: There is no need for an if statement here. Simply pass `name_format=args.get("name_format")` to the function.

gramps_webapi/api/resources/util.py
```diff
@@ -453,13 +453,16 @@ def get_place_profile_for_handle(
 
 
 def get_person_profile_for_object(
-    db_handle: DbReadBase, person: Person, args: list, locale: GrampsLocale = glocale
+    db_handle: DbReadBase,
+    person: Person,
+    args: list,
+    locale: GrampsLocale = glocale,
+    name_format: str = None,
```
DavidMStraub: This type hint is wrong since `None` is not a string. `str | None`, and please make sure that `from __future__ import annotations` is imported at the top of the file.

gramps_webapi/api/resources/util.py
```diff
@@ -453,13 +453,16 @@ def get_place_profile_for_handle(
...
 ) -> Person:
     """Get person profile given a Person."""
     options = []
     if "all" in args or "ratings" in args:
         options.append("ratings")
-    name_display = NameDisplay(xlocale=locale)
```
DavidMStraub, line 462: You are now using the default `displayer` rather than a locale-specific `NameDisplay` instance. Do I understand correctly that this has no effect, since `display_given` is locale-independent?

gramps_webapi/api/resources/util.py
vktimofeev: > This type hint is wrong since `None` is not a string. `str | None`, and please make sure that `from __future__ import annotations` is imported at the top of the file.

Done as you advised!

I wanted to use :

`name_format: Optional[str] = None,`

to avoid problems with python 3.9, but I hope 

`from __future__ import annotations`

makes it unnecessary.

str |None = None

My local mypy produces no errors when being ran with the same arguments as from previously failed command launched from the Github actions workflow file.

gramps_webapi/api/resources/people.py
vktimofeev: > There is no need for an if statement here. Simply pass `name_format=args.get("name_format")` to the function.

Rewritten as you advised!

gramps_webapi/api/resources/util.py
vktimofeev: > You are now using the default `displayer` rather than a locale-specific `NameDisplay` instance. Do I understand correctly that this has no effect, since `display_given` is locale-independent?

Your remark seems to be very valid!  I believe that profile.name_display must be locale dependent. But default displayer is initialized when `gramps.gen.display.name` is imported,  that is it uses server locale - $LANG or $LC_ALL environment variables. But the `get_person_profile_for_object()` receives its `locale` variable originally from the `get` HTTP request handler, and thus it comes from the client.

I just wanted to make an optimization, and I noticed that code that I needed had been already executed in the gramps_webapi/dbloader.py , and it also executes after the `get` handler, but the NameDisplay class has no instance methods to override the locale defined in the constructor when the object had been created (during import mentioned above).

So if you consider that name_display must be locale independent nevertheless, it is not difficult to return to my previous code using default displayer.

DavidMStraub: Looking at the Gramps source code, it seems to me it's indeed *not* locale dependent. So we can use your optimization, right?

https://github.com/gramps-project/gramps/blob/master/gramps/gen/display/name.py#L1089-L1090

vktimofeev: > Looking at the Gramps source code, it seems to me it's indeed _not_ locale dependent. So we can use your optimization, right?
> 
> https://github.com/gramps-project/gramps/blob/master/gramps/gen/display/name.py#L1089-L1090

In fact, there is one line in the Gramps source where the locale is used, namely the line 393 of the file you've just mentioned:

```
        # Translators: needed for Arabic, ignore otherwise
        COMMAGLYPH = xlocale.translation.gettext(",")
```

DavidMStraub: But that's not relevant for the given name I guess?

vktimofeev: > But that's not relevant for the given name I guess?

`profile.name_display` is not specifically for given names: this is application of either default format for Gramps installation, or format specified for the individual `Name` object in the database or the format specified by user of the Gramps Web and passed via the `name_display` webarg. It is like full name formatted according to the rules specific to a particular culture and national language, and for now it can only be configured via Gramps GUI.

My initial idea was to use the same NameDisplay instance methods to extract surname and given name from the Name object and then substitute them to the profile.name_surname and profile.name_given . But I believe that `name_display` that I've just implemented is also usable and probably would replace some of occurrencies of name generation like in gramps-web/src/components/GrampsjsPerson.js, now:

```
return html`${given} ${surname} ${suffix}`
```

what produces "strange and unusual" display string, and it would be:

```
return html`${name_display}`
```

that is configured by the user either via Gramps GUI - it is possible now, and in the future via Gramps Web also.

Sorry for repeating myself, but I found it opportune to precise my entire vision of what I'm doing.

The only purpose for using locale in the `NameDisplay` class is to handle COMMAGLYPH character in Arabic language correctly, and it would not work if the displayer instance is initialized with `xlocale` corresponding to en_US, for example, before a request from the client has come, and then the client is using Arabic language.

Maybe it is a good idea not to use the default displayer from `gramps.gen.display.name` module but rather to initialize the displayer in the gramps_webapi/dbloader.py - its code executes when the client request containing proper locale has arrived, because now this code can modify the default displayer initialized by the `gramps.gen.display.name` code by loading the name formats from the database and the default format from the config, but it cannot change the its initial locale. Eventual problems would occur only when the client uses the Arabic language though - otherwise it must be no difference.

DavidMStraub: My point was the following: the only use of the `name_display` instance in the `get_person_profile_for_object` functoin is for `name_display.display_given`. I claim that `display_given` does not depend on the locale. If you agree, we can use your optimized version (using the default name displayer) and resolve this conversation. If you disagree, please show why `display_given` is locale dependent.

vktimofeev: I could admit that `name_display.display_given()`  is locale-independent: it would return a single world and Gramps makes no name inclinations, etc... that could be locale-dependent.

But now I'm using the same `name_display` object once more, now using its `format_str() `and `display()` methods, that may be locale-dependent: at least, when using Arabic locale it redefines the comma character. I don't know whether this is really important in practice, but we are using an object initialized in one locale (server's default), its code do use the locale data, and then we using this object to produce an output in different locale, now specified by the client. Our code can modify this object at the moment when we know the client's locale (namely in the gramps_webapi/dbloader.py ) , but it can override the locale of this object which is set up in the constructor, and the class `NameDisplay` contains no methods to alter the locale of an object that is already initialized. So my decision is to stay on the safe side: the creation of a new `NameDisplay` instance is not expensive at all. However two solutions are possible: either we re-create the default displayer now using the locale from the request, or we initialize it there where we use it like I did in my latest version.

The `display` name from the `gramps.gen.display.name` module which refers to an instance of `NameDisplay` class from the same module is also used being imported also as `name_displayer` in the gramps_webapi/api/report.py source file.

DavidMStraub: I see, thanks for investigating! Then we keep the original version for the name displayer instantiated with the locale, agreed?

vktimofeev: You are welcome!

>  we keep the original version for the name displayer instantiated with the locale, agreed?

Do you mean to adopt my last commit version.  For the moment I would consider it good. 

At the moment there are left to update the test case(s) and to pass all the tests, aren't there?

DavidMStraub: Yep, agreed!

DavidMStraub:
I forgot to write - please also add the new key to `apispec.yaml`, thanks!

DavidMStraub:
Let's keep it as a function argument, analogously to `locale`.

vktimofeev:
> Let's keep it as a function argument, analogously to `locale`.

Following this way I had to pass the `name_format` argument starting from HTTP GET handlers from one to another. The `get_%s_profile_for_object()` function  from gramps_webapi/api/resources/util.py call one another, but they can't pass the query string arguments. 

(There are sources of confusion in these functions: the "args" argument supposed to be a list to represent the list from the profile argument value from the query string).

The `name_display` field is rendered only by the `get_person_profile_for_object() ` function, so we could obtain the `name_format` argument with only one line from the request context.

Please review the changes that I've commit. The tests has passed locally at my workstation, but I'm not pretty sure that some modifications weren't necessary for `get_citation_profile_for_object()`, `get_note_profile_for_object()` etc. I will continue if you approve the changes from the last commit - otherwise it makes no sense, since I've already spent much time for this approach, and about the same amount of time would be necessary to accomplish the code.

I also added a couple of tests to verify that `name_format` query string argument produces the necessary effect. These tests were extremely helpful: I found some cases when `name_format` wasn't passed to the `get_person_profile_for_object() `

Please also take a look at the apispec.yaml

DavidMStraub:
Excellent work, I think it's good like this if tests pass!

> (There are sources of confusion in these functions: the "args" argument supposed to be a list to represent the list from the profile argument value from the query string).

`args` will be a `dict` - apparently there were wrong type hints in some places in the existing code base. Can you please make sure the type hint is `dict` everywhere and the empty case is `{}`, not `[]`? Then we're done I think!

vktimofeev:
> Excellent work, I think it's good like this if tests pass!
> 
> > (There are sources of confusion in these functions: the "args" argument supposed to be a list to represent the list from the profile argument value from the query string).
Thank you!

> `args` will be a `dict` - apparently there were wrong type hints in some places in the existing code base. Can you please make sure the type hint is `dict` everywhere and the empty case is `{}`, not `[]`? Then we're done I think!

The value of of the expression `args["profile"]` in the `get_person_profile_for_object() ` calls and similar `get_*_profile_for_object function` calls is apparently a list of str. This is defined in the `@use_args` decorator, namely `"profile": fields.DelimitedList(...` in 

https://github.com/gramps-project/gramps-web-api/blob/6739fa353efc64cca951f49aa47dd1acffdb4f82/gramps_webapi/api/resources/base.py#L210

or 

https://github.com/gramps-project/gramps-web-api/blob/6739fa353efc64cca951f49aa47dd1acffdb4f82/gramps_webapi/api/resources/base.py#L353

This is just a list of values for the "profile" key in the query string. But the name of the argument in the `def get_person_profile_for_object(...)` is "args", so the type hint `list` is correct and `dict` is not. This is not an error in the code - there is nothing to correct, but this is not very good choice of the argument name, in my opinion. 

I corrected a single initialization of this argument in a function call from ` {}` to `[]`, but I found no other occurrencies of such incorrect usage.

I made the last update by adding some `name_format` arguments in functions calls, a minor one, and probably the last one for now. My local tests have passed.

DavidMStraub:
Oh good catch - you're right, this is a poor choice, but nothing that needs to be fixed in this PR, thanks!

PR #678, opened 2025-08-22
Endpoint for Y chromosome clade information

DavidMStraub:
This adds a new endpoint `/api/people/<handle>/ydna` that takes raw Y chromosome SNP information from a person attribute of type `Y-DNA` (in a format compatible with [yclade](https://github.com/DavidMStraub/yclade)) and returns the most likely clade from the Y full tree as well as all ancestor clades with age information (date format and date of the MRCA).


DavidMStraub:
TODO: handling of Y tree cache dir and precaching in docker image.

Issue #681, opened 2025-08-25
Remote code execution issue of name format implementation


DavidMStraub:
@vktimofeev I realized that, by allowing arbitrary strings for `name_format`, we are introducing a remote code execution vulnerability since Gramps' name displayer uses `exec` for the formatting function.

So we need to add input sanitation to *all* endpoints that accept a name format. We can abstract it into a utility function.

It would be great if you could look into that.

vktimofeev:
@DavidMStraub ok, I'm looking into that.

How do you think, could the input sanitation code be invoked like a verify function in '@use_args' decorator like here:

https://github.com/gramps-project/gramps-web-api/blob/fdb0d4b4157494994acc67572c8fc6312a12f06c/gramps_webapi/api/resources/base.py#L213

?

vktimofeev:
I'm reading a lots of opinions that it were impossible to write a safe code which passes the user input to the exec, like in this conversation, for example:

https://www.reddit.com/r/Python/comments/6b3zzy/sanitizing_code_for_an_exec_command/

The use of exec within the NameDisplay is motivated as it follows:
       
       _The job of building the name from a format string is rather
        expensive and it is called lots and lots of times. So it is worth
        going to some length to optimise it as much as possible.
        This method constructs a new function that is specifically written
        to format a name given a particular format string. This is worthwhile
        because the format string itself rarely changes, so by caching the new
        function and calling it directly when asked to format a name to the
        same format string again we can be as quick as possible._

Is it really a concern for us? So, in my opinion, one option is to rewrite the `NameDisplay` code and include the new code as an utility function in gramps-web-api, and another one - is to rewrite this code in JavaScript to use it in the browser. 

Maybe as a first step we could just remove the `name_format` query string argument, but use the `profile.name_display` field anyway which is produced as an application of either default name format that for the moment can only be set up from the GUI, or the format set up for a `Name` class instance individually. I'm using this feature this way for now, and it works fine for me. 

In my opinion, the owner of a family tree should be able to manipulate the name format not only via GUI, but also via Gramps Web frontend - the latest feature is to be implemented. But the read-only user should be able to manipulate the name formats either by setting up name_format query string, or by the frontend code that duplicates the `gramps.gen.display.name`.

I would try to implement everything I've just talked about, although these are a little bit difficult tasks for me, and they would require some time like 2-3 weeks.

[...]

PR #682, opened 2025-08-26
#681 validate name_format query parameter with regexp to prevent mali…

vktimofeev:
…cious code injection

I suppose, the regexp for accepted name display format and the name of the constant NAME_FORMAT_REGEXP may be changed.

Even the current regexp should pass any reasonable combinations of Name instance components with various punctuation added to them, but it prohibits any "words" including python identifiers and keywords, and therefore makes impossible malicious code injections.

This is, of course, a kind of paranoiac approach which is very restrictive. I'm ready to adopt a better one, not necessary with regexp, but maybe with some custom validate function.

### DIFF
--- a/gramps_webapi/const.py
+++ b/gramps_webapi/const.py
@@ -200,3 +200,7 @@ DISABLED_EXPORTERS = ["gpkg"]
 TELEMETRY_ENDPOINT = "https://telemetry-cloud-run-442080026669.europe-west1.run.app"
 TELEMETRY_TIMESTAMP_KEY = "telemetry_last_sent"
 TELEMETRY_SERVER_ID_KEY = "telemetry_server_uuid"
+
+NAME_FORMAT_REGEXP = (
+    r'^(%[%tTfFlLcCxXiImMyYoOrRpPqQsSnNgG]|%[0-2][mMyY]|[ "\',.:;\]\]\(\)\{\}\&\@])*$'
+)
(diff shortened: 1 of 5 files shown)


gramps_webapi/const.py
```diff
@@ -200,3 +200,7 @@
 TELEMETRY_ENDPOINT = "https://telemetry-cloud-run-442080026669.europe-west1.run.app"
 TELEMETRY_TIMESTAMP_KEY = "telemetry_last_sent"
 TELEMETRY_SERVER_ID_KEY = "telemetry_server_uuid"
+
```
DavidMStraub, line 203:
```suggestion
# Regular expression for allowed values of the `name_format` query parameter.
```

gramps_webapi/const.py
```diff
@@ -200,3 +200,7 @@
...
+NAME_FORMAT_REGEXP = (
+    r'^(%[%tTfFlLcCxXiImMyYoOrRpPqQsSnNgG]|%[0-2][mMyY]|[ "\',.:;\]\]\(\)\{\}\&\@])*$'
```
DavidMStraub: `\]\]` should be `\]\[` I guess?

DavidMStraub: Let's please disallow " and '. Not 100% sure these can be exploited, but they make me uneasy. While quotes can be useful for nicknames or callnames, users can always use Unicode typographic quotes instead.

DavidMStraub:
Thanks, I think this is the right approach! Let's merge it after addressing the comments.

gramps_webapi/const.py
vktimofeev: > `\]\]` should be `\]\[` I guess?

done!

> Let's please disallow " and '. Not 100% sure these can be exploited, but they make me uneasy. While quotes can be useful for nicknames or callnames, users can always use Unicode typographic quotes instead.

used \u0022 and \u0027 instead of ASCII double and single quotes in the pattern.

vktimofeev:
> Thanks, I think this is the right approach! Let's merge it after addressing the comments.

You are welcome! 

I made some amendments to address the comments. I hope it's all right now.

DavidMStraub:
Looks good!

PR #691, opened 2025-09-27
Add OIDC authentication support with multi-provider configuration

AlexBocken:
Adds comprehensive OIDC authentication support to Gramps Web API with multi-provider configuration and automatic role mapping.

## Features

  - Multi-provider support: Google, Microsoft, GitHub, and custom OIDC providers
  - Automatic role mapping: Map OIDC groups/claims to Gramps user roles
  - Flexible configuration: Support for various claim structures and nested groups
  - Backward compatibility: Preserves existing authentication when OIDC is disabled
  - Security: Proper token validation and user management

## Configuration

  All OIDC settings use GRAMPSWEB_OIDC_* environment variables for consistency. Role mapping automatically assigns permissions based on group membership from OIDC tokens.

 ## Testing

 Complete testing environment available at: https://github.com/AlexBocken/gramps-oidc-testing
 
 There are some unit-tests included but to be honest, these still need improving.
 
 ## Misc
 
 Happy to improve the codebase based on feedback, I tried to stick to utility functions and similar coding style to not waste effort.

 **Related**: Requires companion frontend PR for full OIDC integration. (PR [gramps-web/#763)(https://github.com/gramps-project/gramps-web/pull/763)


DavidMStraub:
Thanks a lot! :partying_face:  Will review ASAP.

Fixes #355

AlexBocken:
Noticed some `os.getenv` were able to get through (using the default names in configs is not helpful for debugging...)

AlexBocken:
Further investigation with OIDC Provider alternatives to keycloak revealed some issues. Authentik does not seem to want to cooperate in the token exchange with the current setup. Investigating.

README.md
```diff
@@ -12,6 +12,32 @@ Gramps Web API is the backend of [Gramps Web](https://www.grampsweb.org/), a gen
 - Developer documentation for Gramps Web API: https://www.grampsweb.org/dev-backend/
 - Documentation for Gramps Web: https://www.grampsweb.org
 
+## OIDC Configuration
```
DavidMStraub: This is not the right place to document configuration. Please submit a PR to https://github.com/gramps-project/gramps-web-docs.

gramps_webapi/api/resources/oidc.py
```diff
@@ -0,0 +1,202 @@
+#
+# Gramps Web API - A RESTful API for the Gramps genealogy program
+#
+# Copyright (C) 2020-2024      David Straub
```
DavidMStraub: I guess you want to update that header in all files you added

gramps_webapi/api/resources/oidc.py
```diff
@@ -0,0 +1,202 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+#
+
+"""OIDC authentication endpoint blueprint."""
```
DavidMStraub: This is not a blueprint

gramps_webapi/api/resources/oidc.py
```diff
@@ -0,0 +1,202 @@
...
+    @use_args(
+        {
+            "provider": fields.Str(required=True),
+        },
+        location="query",
+    )
+    def get(self, args):
+        """Redirect to OIDC provider for authentication."""
+        if not is_oidc_enabled():
+            abort_with_message(404, "OIDC authentication is not enabled")
```
DavidMStraub: I would return 405 here

gramps_webapi/api/resources/oidc.py
```diff
@@ -0,0 +1,202 @@
...
+            "session_state": fields.Str(required=False),
+            "error": fields.Str(required=False),
+            "error_description": fields.Str(required=False),
+        },
+        location="query",
+    )
+    def get(self, args):
+        """Handle OIDC callback and create JWT tokens."""
+        if not is_oidc_enabled():
+            abort_with_message(404, "OIDC authentication is not enabled")
```
DavidMStraub: 405

gramps_webapi/api/resources/oidc.py
```diff
@@ -0,0 +1,202 @@
...
+            if provider_id == "github":
+                # GitHub OAuth 2.0 - get user info from API
+                resp = oidc_client.get("user", token=token)
+                userinfo = resp.json()
+            else:
+                # Standard OIDC - get userinfo from userinfo endpoint
+                userinfo = oidc_client.userinfo(token=token)
+
+        except Exception as e:
+            logger.error(f"OIDC callback error for provider '{provider_id}': {e}")
```
DavidMStraub: What about

```python
logger.exception(f"OIDC callback error for provider '{provider_id}'")
```

gramps_webapi/auth/oidc.py
```diff
@@ -0,0 +1,360 @@
...
+    """Create or update a user based on OIDC userinfo.
+
+    Args:
+        userinfo: User information from OIDC provider
+        tree: Tree identifier (optional)
+        provider_id: OIDC provider identifier
+
+    Returns the user GUID.
+    """
+    from flask import current_app
```
DavidMStraub: Please import this at the top level.

gramps_webapi/auth/oidc.py
```diff
@@ -0,0 +1,360 @@
...
+            break
+
+    logger.info(f"User {role_claim} {user_groups} mapped to role {highest_role}")
+    return highest_role
+
+
+def create_or_update_oidc_user(
+    userinfo: Dict,
+    tree: Optional[str] = None,
+    provider_id: str = "custom"
```
DavidMStraub: Why are there defaults for `tree` and `provider_id` that only make sense in specific scenarios? Explicit is better than implicit here IMO.

gramps_webapi/auth/oidc.py
```diff
@@ -0,0 +1,360 @@
...
+    # Get username from provider-specific claim with fallback to 'sub'
+    username_claim = provider_config.get("username_claim", "preferred_username")
+    username = userinfo.get(username_claim) or userinfo.get("sub")
+
+    email = userinfo.get("email", "")
+    full_name = userinfo.get("name", "")
+
+    # For GitHub, get email from separate API call if not in userinfo
+    if provider_id == "github" and not email:
+        email = userinfo.get("email") or f"{userinfo.get('login', 'user')}@users.noreply.github.com"
```
DavidMStraub: Will emails to @users.noreply.github.com be forwarded to the user?

gramps_webapi/auth/oidc.py
```diff
@@ -0,0 +1,360 @@
...
+        ROLE_EDITOR: current_app.config.get("OIDC_GROUP_EDITOR", ""),
+        ROLE_CONTRIBUTOR: current_app.config.get("OIDC_GROUP_CONTRIBUTOR", ""),
+        ROLE_MEMBER: current_app.config.get("OIDC_GROUP_MEMBER", ""),
+        ROLE_GUEST: current_app.config.get("OIDC_GROUP_GUEST", ""),
+    }
+
+    # Check if any role mapping is configured
+    has_role_mapping = any(group_name.strip() for group_name in role_mapping.values())
+    if not has_role_mapping:
+        logger.info("No OIDC role mapping configured (no GRAMPSWEB_OIDC_GROUP_* environment variables set). Preserving existing user roles.")
```
DavidMStraub: Not environment variables but generally config options (without `GRAMPSWEB_`)

gramps_webapi/api/resources/oidc.py
```diff
@@ -0,0 +1,202 @@
...
+                user_id=user_id,
+                permissions=permissions,
+                tree_id=tree_id,
+                include_refresh=True,
+                fresh=True,
+            )
+
+            # Redirect to frontend with tokens in URL fragment (more secure than query)
+            from flask import redirect
+            frontend_url = current_app.config.get("FRONTEND_URL", "http://localhost:8001")
```
DavidMStraub: There is no such config option. If we add it - which I guess makes sense - I think we should account for the fact that, in production, most people serve frontend and backend from the same host using a reverse proxy and we should use `BASE_URL` as the fallback. (So `FRONTEND_URL` would be the URL of the frontend IFF it is different from `BASE_URL`).

We can add `http://localhost:8001` as the frontend URL in `.devcontainer/docker-compose.yml`.

gramps_webapi/api/resources/oidc.py
AlexBocken: adjusted to "resources" instead of  "endpoint blueprint". Follows a similar pattern to other files in the repo.

gramps_webapi/auth/oidc.py
AlexBocken: Good catch. No, they won't. Can only use this email for signing of commits on GitHub, my mistake. Removing this special case then.

README.md
AlexBocken: https://github.com/gramps-project/gramps-web-docs/pull/46

AlexBocken:
Alright, I have addressed all your raised points. Besides this the following items are imo still open and will be addressed by me in the coming days:

1. proper logout handling for OIDC (otherwise we do not log out on OIDC provider side if we click "Logout" in gramps-web)
2. unit-testing cleanup and full coverage
3. Test with other OIDC providers (e.g. Authentik) and fix any problems

.devcontainer/docker-compose.yml
```diff
@@ -27,6 +27,7 @@ services:
       GRAMPSWEB_CELERY_CONFIG__broker_url: redis://redis:6379/0
       GRAMPSWEB_CELERY_CONFIG__result_backend: redis://redis:6379/0
       GRAMPSWEB_LOG_LEVEL: DEBUG
+      GRAMPSWEB_FRONTEND_URL: http://localhost:8001
```
DavidMStraub: Let's make this empty by default rather than the development value

DavidMStraub:
Thanks for the changes and the explanations!

Just one more issue I noticed: defaulting to `ROLE_GUEST` means the entire Gramps Web instance (except private records) is opened up to everyone on the internet with a Google account. This cannot be the default! IMO we should default to `ROLE_DISABLED` and dispatch `send_email_new_user` exactly like in the registration endpoint.

gramps_webapi/auth/__init__.py
```diff
@@ -201,6 +202,16 @@ def _get_user_detail(user, include_guid: bool = False):
     }
     if include_guid:
         details["user_id"] = user.id
+    if include_oidc_accounts:
+        oidc_accounts = get_user_oidc_accounts(user.id)
+        details["oidc_accounts"] = oidc_accounts
+        # Add a simplified account source summary for frontend display
+        if oidc_accounts:
+            from ..api.util import get_config
```
DavidMStraub: Let's import this at the top please

gramps_webapi/api/resources/oidc.py
```diff
@@ -0,0 +1,281 @@
...
+            tokens = get_tokens(
+                user_id=user_id,
+                permissions=permissions,
+                tree_id=tree_id,
+                include_refresh=True,
+                fresh=True,
+            )
+
+            # Redirect to frontend with secure HTTP-only cookies
+            from flask import redirect, make_response
```
DavidMStraub: Top-level import please

gramps_webapi/api/resources/oidc.py
```diff
@@ -0,0 +1,281 @@
...
+            frontend_url = get_config("FRONTEND_URL") or get_config("BASE_URL")
+            response = make_response(redirect(f"{frontend_url}/oidc/complete"))
+
+            # Set HTTP-only cookies (secure=False for localhost development)
+            import os
+            is_development = os.getenv('FLASK_ENV') == 'development' or 'localhost' in frontend_url or '127.0.0.1' in frontend_url
```
DavidMStraub: 1. This should be a function, it's repeated below
2. `FLASK_ENV` has been removed in Flask 2.3.0, use `current_app.debug`
3. If `frontend_url` is `None`, this will raise a TypeError

gramps_webapi/api/resources/oidc.py
```diff
@@ -0,0 +1,281 @@
...
+                secure=not is_development,  # Allow HTTP in development
+                samesite='Lax',
+                path='/'
+            )
+
+            logger.info(f"Set OIDC cookies, redirecting to {frontend_url}/oidc/complete")
+            return response
+
+        except ValueError as e:
+            logger.error(f"Error creating/updating OIDC user for provider '{provider_id}': {e}")
```
DavidMStraub: `logger.exception`?

gramps_webapi/api/resources/oidc.py
```diff
@@ -0,0 +1,281 @@
...
+                fresh=True,
+            )
+
+            # Redirect to frontend with secure HTTP-only cookies
+            from flask import redirect, make_response
+            frontend_url = get_config("FRONTEND_URL") or get_config("BASE_URL")
+            response = make_response(redirect(f"{frontend_url}/oidc/complete"))
+
+            # Set HTTP-only cookies (secure=False for localhost development)
+            import os
```
DavidMStraub: Top level

gramps_webapi/api/resources/oidc.py
```diff
@@ -0,0 +1,281 @@
...
+
+
+class OIDCTokenExchangeResource(Resource):
+    """Resource for securely exchanging OIDC tokens from cookies."""
+
+    @limiter.limit("10/minute")
+    def get(self):
+        """Exchange HTTP-only cookies for tokens that can be stored in localStorage."""
+        from flask import request as flask_request, make_response, jsonify
+        import os
```
DavidMStraub: All top level

gramps_webapi/api/resources/oidc.py
```diff
@@ -0,0 +1,281 @@
...
+        if not access_token or not refresh_token:
+            logger.error("No OIDC tokens found in cookies")
+            abort_with_message(400, "No OIDC tokens found in cookies")
+
+        # Return tokens and clear cookies
+        response = make_response(jsonify({
+            'access_token': access_token,
+            'refresh_token': refresh_token,
+            'token_type': 'Bearer'
+        }))
```
DavidMStraub: I'm pretty sure you can omit `make_response` *and* `jsonify` in modern Flask, that makes this more readable (also valid in other places)

DavidMStraub:
Great work, thanks for your efforts, I think we are close.

But I think I found one more security issue.

If a user exists with email address A and tree X, they can sing up for tree Y with an OIDC account that has the same email address, and the code will update the existing user to use the new tree, but preserving the role. So, say, any tree owner on a given multi tree server could escalate to become tree owner of any other tree.

Although we could close this whole with appropriate checks, I think we should remove the email matching feature from the scope of this PR to keep it simpler. Can you please remove it? We can always add it back later. Sorry for suggesting that initially.

Apart from that, two questions:

1. Would it be easy to add an endpoint to associate an *existing* user account with an OIDC account manually? I'm thinking of something like a "associate Google login" button in the user profile. Again this would be out of scope of the present PR but I want to make sure the architecture can be modified to allow it in the future (without more user db schema changes)
2. What is the rationale for the `last_login`? I know it's "good to know", but we don't track it for "normal" user accounts either. Is it needed for any security or functional purpose? If not, I suggest removing it.

gramps_webapi/api/resources/oidc.py
AlexBocken: Good catch. Removed the make_response as it really is unnecessary but kept the jsonify. Without the jsonify the cookie setting becomes too unreadable imo.

gramps_webapi/auth/oidc.py
```diff
@@ -0,0 +1,451 @@
...
+
+    # Step 3: Create new user account
+    logger.info(f"Creating new OIDC user for {provider_id}:{subject_id}")
+
+    # Generate a clean username - for custom providers use the display username as-is,
+    # for others prefix with provider to avoid conflicts
+    if provider_id == "custom":
+        final_username = display_username or f"user_{uuid.uuid4().hex[:8]}"
+    else:
+        final_username = f"{provider_id}_{display_username or uuid.uuid4().hex[:8]}"
```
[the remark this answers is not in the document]
AlexBocken: logic too trivial to extract into function imo. Keeping current logic.

AlexBocken:
Oh wow, good catch with the email matching privilege escalation! Apologies, I did not consider the multi-tree scenario when setting this up. Removed. Happy to create a PR in the future to re-introduce it in a cleaner way that prevents this issue.

1. Yes that should be possible, the schema already allows for one to many associations, I don't see any database schema issues. Can also propose that in a future PR, definitely useful. Currently we have the issue that a locally registered user can not use an OIDC account with the same email. This is not a limitation of the oidc_accounts table but the existing User schema. Imo this is fine as long as we handle the OIDC association properly in the account settings like you proposed.

2. purely speculative as a nice to have. Removed as per your valid points.

Besides these points I have also just pushed a commit concerning your remaining points.

gramps_webapi/api/resources/oidc.py
```diff
@@ -0,0 +1,295 @@
...
+            logger.exception(f"OIDC callback error for provider '{provider_id}'")
+            abort_with_message(401, f"OIDC authentication failed for {provider_id}")
+
+        tree = args.get("tree")
+        if (
+            tree
+            and current_app.config["TREE"] != TREE_MULTI
+            and tree != current_app.config["TREE"]
+        ):
+            abort_with_message(403, "Not allowed in single-tree setup")
```
DavidMStraub: Let's make sure thus fails if tree is not provided in a multi-tree setup.

```suggestion
        if (
            !tree
            and current_app.config["TREE"] == TREE_MULTI
        ):
            abort_with_message(403, "Tree is required")
```

tests/test_oidc_auth.py
```diff
@@ -0,0 +1,335 @@
...
+                        with patch("gramps_webapi.auth.oidc.add_user") as mock_add_user:
+                            with patch("gramps_webapi.auth.oidc.get_guid", return_value="test-guid"):
+                                create_or_update_oidc_user(case["userinfo"])
+
+                    mock_add_user.assert_called_once()
+                    args, kwargs = mock_add_user.call_args
+                    self.assertEqual(kwargs["name"], case["expected_username"])
+                    self.assertEqual(kwargs["role"], case["expected_role"])
+
+if __name__ == "__main__":
```
DavidMStraub: We don't need this as unit tests are run with `pytest`

gramps_webapi/auth/oidc.py
```diff
@@ -0,0 +1,407 @@
...
+            )
+
+        return existing_user_id
+
+    # Step 2: Create new user account (email linking removed for security)
+    logger.info(f"Creating new OIDC user for {provider_id}:{subject_id}")
+
+    # Generate a clean username - for custom providers use the display username as-is,
+    # for others prefix with provider to avoid conflicts
+    if provider_id == "custom":
```
DavidMStraub: Can we please make `"custom"` a string constant?

DavidMStraub (changes requested):

AlexBocken:
Getting lots of unit test fails for the Python 3.9 build, even outside of my added tests. Should this be the case?

DavidMStraub:
I had the same problem and solved it by unpinning sentence transformers, it should work again after you've rebased on `master`.

AlexBocken:
rebased, thanks for that

AlexBocken:
Can confirm that with the rebase the tests seem to run smoothly both for 3.9 and 3.13, at least locally. Let's hope the same also happens once you run them via Action :smile:

DavidMStraub (approved):

PR #702, opened 2025-10-16
Fixes to the OIDC endpoint

DavidMStraub:
@AlexBocken while testing the OIDC endpoint with Google login, I noticed two issues:

1. The callback fails because Google sends additional query parameters. I changed it now to ignore them rather than throw an exception
2. The `send_user_new_email` failed because I merged #694 last week which changed the API. Fixed now.

I also have a third problem, but still figuring out whether it's just my config :slightly_smiling_face:

### DIFF
--- a/gramps_webapi/api/resources/oidc.py
+++ b/gramps_webapi/api/resources/oidc.py
@@ -207,7 +207,7 @@ class OIDCCallbackResource(Resource):
 
             # Redirect to frontend with secure HTTP-only cookies
             frontend_url = get_config("FRONTEND_URL") or get_config("BASE_URL")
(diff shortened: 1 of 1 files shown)


DavidMStraub:
Lots of style changes - I have to check whether black is properly set up in pre-commit, but not a big problem apart from making the diff difficult.

DavidMStraub:
Found the third problem. `BASE_URL` defaults to `http://localhost/` and `FRONTEND_URL` defaults to `BASE_URL`, such that there is a double slash in the redirect URL. Will add an `rstrip` to fix.

AlexBocken:
Many thanks for the debugging and fixing! Much appreciated

Issue #706, opened 2025-10-22
Add time period of place references to the place profile


Ziesie1:
This issue belongs to gramps-project/gramps-web/issues/748.

To display the assigned time period in the place hierarchy view, the profile of /api/places must be adapted.
Currently, the profile of a place doesn't contain information about the time period of the referenced parent places.

PR #707, opened 2025-10-22
add place_references with date information to the place profile (#706)

Ziesie1:
This PR implements the changes requested in issue #706.

It adds the information about the time period for the place references to a place profile.
Here is an example of the new profile output. The Property `place_references` is new.
```json
"profile": {
            "alternate_names": [
                "Leningrad",
                "Petrograd"
            ],
            "alternate_place_names": [
                {
                    "date_str": "zwischen 1924-01-26 und 1991-09-06",
                    "value": "Leningrad"
                },
                {
                    "date_str": "zwischen 1914 und 1924",
                    "value": "Petrograd"
                }
            ],
            "gramps_id": "P0443",
            "lat": 0,
            "long": 0,
            "name": "Saint Petersburg",
            "parent_places": [
                {
                    "alternate_names": [],
                    "alternate_place_names": [],
                    "gramps_id": "P0442",
                    "lat": 0,
                    "long": 0,
                    "name": "Russia",
                    "type": "Land"
                }
            ],
            "place_references": [
                {
                    "date_str": "von 1930-05-06 bis 1950-09-08",
                    "value": "Russia"
                },
                {
                    "date_str": "ab 2010",
                    "value": "Decatur"
                },
                {
                    "date_str": "",
                    "value": "Greece"
                }
            ],
            "references": {},
            "type": "Stadt"
        },
```

Remarks:
- `place_references` contains the reference information for the next place parents
- `parent_places`, on the other hand, contains the complete place hierarchy (only the first place is used here).

An alternative approach could be to only have one representation of `parent_places`, for example:
```json
"parent_places": [
      {
           "date_str":"von 1930-05-06 bis 1950-09-08",
           "place": {
                "alternate_names": [],
                "alternate_place_names": [],
                "gramps_id": "P0442",
                "lat": 0,
                "long": 0,
                "name": "Russia",
                "parent_places": [],
               "type": "Land"
           }
      },
      {
           "date_str":"ab 2010",
           "place": {
                "alternate_names": [],
                "alternate_place_names": [],
                "gramps_id": "P0XXX",
                "lat": 0,
                "long": 0,
                "name": "Decatur",
                "parent_places": [ ], // If Decatur also has a parent location, it would be displayed here.
               "type": "Land"
           }
      },
     //...
]
```
However, this solution has the disadvantage that the output can be very large if there are many nested locations. To do this, you also have to adjust the display of the location hierarchy in the UI, as the locations now have to be read from this structure.

### DIFF
--- a/gramps_webapi/api/resources/util.py
+++ b/gramps_webapi/api/resources/util.py
@@ -451,31 +451,53 @@ def get_place_profile_for_object(
                 break
             if handle is None or handle in parent_places_handles:
                 break
-            _place = db_handle.get_place_from_handle(handle)
+            _place = None
+            try:
+                _place = db_handle.get_place_from_handle(handle)
+            except HandleError:
+                break
             if _place is None:
                 break
             parent_places_handles.append(handle)
-        profile["parent_places"] = [
-            get_place_profile_for_object(
-                db_handle=db_handle,
-                place=db_handle.get_place_from_handle(parent_place),
-                locale=locale,
-                parent_places=False,
-            )
-            for parent_place in parent_places_handles
-        ]
-        profile["direct_parent_places"] = [
-            {
-                "place": get_place_profile_for_object(
-                    db_handle=db_handle,
-                    place=db_handle.get_place_from_handle(place_ref.ref),
-                    locale=locale,
-                    parent_places=False,
-                ),
-                "date_str": locale.date_displayer.display(place_ref.date),
-            }
-            for place_ref in place.get_placeref_list()
-        ]
+
+        parent_places_value = []
+        for parent_place in parent_places_handles:
+            try:
+                place_value = db_handle.get_place_from_handle(parent_place)
+                if place_value is None:
+                    continue
+                parent_places_value.append(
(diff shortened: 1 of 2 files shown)


Ziesie1:
I don't think so.

Consider the following example:
Hanover -> Lower Saxony  (>1946)-> Germany
Hanover -> Province of Hanover (1866 bis 1946) -> German Empire
Hanover  -> Kingdom of Hannover (< 1866)

With the current implementation, `parent_places` will only contains a list with [Lower Saxony, Germany]. This is also what the frontend currently displays: it only shows each first parent entry.
What we need for the issue is list of all next parent places [Lower Saxony, Province of Hanover, Kingdom of Hannover] with the corresponding date string (`date_str`).

I would also like to point out that `date_str` has nothing to do with the place itself. It belongs to the reference between a child and a parent place. For this reason, I have separated it from the `place` itself.

DavidMStraub:
This is ready to merge now, right?

Ziesie1:
Yes, I think so.
I've implemented both of your points 🙃. If everything looks good here, it can be merged.

Issue #708, opened 2025-10-26
Extend the development environment for debugging and testing


Ziesie1:
After spending some minutes on the development project, I noticed that a few things were missing.

1. There is no default launch.json for debugging the API in VSCode. 
Currently I created one for me locally, but wouldn't that be something that could be added to the project?

2. The [Dockerfile ](https://github.com/gramps-project/gramps-web-api/blob/master/.devcontainer/Dockerfile) for the devcontainer doesn't contain all dependencies like the main [Dockerfile](https://github.com/gramps-project/gramps-web-api/blob/master/Dockerfile).   
For exmaple, when working with PDF files, I encountered some problems because the tool `poppler-utils` wasn't installed in the devcontainer environment.   
Wouldn't it be better to have a base container with all dependencies on which the development and production containers could be based? This is possible with [Docker Multi-stage builds](https://docs.docker.com/build/building/multi-stage/#use-a-previous-stage-as-a-new-stage).

What do you think?

PR #723, opened 2025-11-01
add debugging configurations for the web API and tests (#708)

Ziesie1:
This pull-request implements parts of #708.

- It adds launch.json for debugging the web API and tests
- It adds a new environment variable `DISABLE_CACHES`.
I had the problem that my breakpoints were never reached. It took me a long time to realize that it was due to the caching of the request. For this reason, I disable all caching for debugging.
- It adds `poppler-utils `to the devcontainer environment which is needed for debugging with .pdf files (similar to the production [Dockerfile](https://github.com/gramps-project/gramps-web-api/blob/ad04ca3950fdf9691b248ae95a49f867820882de/Dockerfile#L18))

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -12,6 +12,11 @@
       },
       "runOptions": {
         "instanceLimit": 1
+      },
+      "options": {
+        "env": {
+          "DISABLE_CACHES": "1"
+        }
       }
     },
     {


Ziesie1:
Yes, I will.
I just have a few additional questions.

1. Should I change the environment variable DISABLE_CACHES to an configurable parameter that can also be changed via config file? If so, I would also include GRAMPSWEB_DISABLE_CACHES in the chapter [optional settings chapters in the docs](https://www.grampsweb.org/install_setup/configuration/#optional-settings).
2. What do you think about disabling caching for the “Serve Web API” tasks in vscode as well? Here, too, I had the problem that my code changes had no effect on my requests due to caching.

DavidMStraub:
> Should I change the environment variable DISABLE_CACHES to an configurable parameter that can also be changed via config file? If so, I would also include GRAMPSWEB_DISABLE_CACHES in the chapter [optional settings chapters in the docs](https://www.grampsweb.org/install_setup/configuration/#optional-settings).

I thought about - but actually it's already possible to disable all caches by setting the cache type to `NullCache`. So I think adding an additional configuration parameter is redundant. Having the environment variable could still make sense to "quickly" switch all caches off and on again.

> What do you think about disabling caching for the “Serve Web API” tasks in vscode as well? Here, too, I had the problem that my code changes had no effect on my requests due to caching.

Yes, it might make sense to have this as default. We could simply make the disable cache env var true in the compose file - then whoever doesn't like it can set it to false again.

PR #739, opened 2025-12-06
Add EMAIL_USE_SSL and EMAIL_USE_STARTTLS parameters

ericmoret:
Adds `EMAIL_USE_SSL` and `EMAIL_USE_STARTTLS` parameters for clearer SMTP configuration.

## Problem

`EMAIL_USE_TLS` is confusing:
- `EMAIL_USE_TLS=true` → implicit SSL (port 465) 
- `EMAIL_USE_TLS=false` → STARTTLS (port 587)

Users expect `EMAIL_USE_TLS=true` to work with port 587, but it doesn't.

## Solution

New explicit parameters with priority fallback:
1. `EMAIL_USE_SSL` - for port 465 (implicit SSL)
2. `EMAIL_USE_STARTTLS` - for port 587 (STARTTLS)
3. `EMAIL_USE_TLS` - legacy (still works, logs deprecation warning)

## Examples

**New way:**
```bash
EMAIL_PORT=587
EMAIL_USE_STARTTLS=true
```

**Old way (still works):**
```bash
EMAIL_PORT=465
EMAIL_USE_TLS=true
```

## Changes

- Added `EMAIL_USE_SSL = None` and `EMAIL_USE_STARTTLS = None` to `config.py`
- Updated `send_email()` in `util.py` with new logic
- Backward compatibility maintained

Documentation at grampsweb.org will need updating - happy to do that in a follow-up PR.

### DIFF
--- a/gramps_webapi/api/util.py
+++ b/gramps_webapi/api/util.py
@@ -522,19 +522,19 @@ def _resolve_smtp_config(
 ) -> tuple[bool, bool]:
     """Helper to resolve SMTP encryption settings.
 
-    Return tuple[bool, bool]: (use_ssl, use_starttls)
-    """
-    # If new params are set, use them
+    Return tuple[bool, bool]: (use_ssl, use_starttls)"""
     if use_ssl is True:
         return True, False
     if use_starttls is True:
         return False, True
 
-    # Legacy EMAIL_USE_TLS behavior
+    if use_ssl is False or use_starttls is False:
+        return False, False
+
     if use_tls:
-        return True, False  # SSL
+        return True, False
     else:
-        return False, (port != 25)  # STARTTLS if not port 25
+        return False, (port != 25)
 
 
 def send_email(


gramps_webapi/api/util.py
```diff
@@ -538,17 +538,38 @@ def send_email(
...
-        else:
+            smtp.ehlo()
+        elif use_starttls:
             smtp = smtplib.SMTP(host=host, port=port, timeout=10)
             smtp.ehlo()
-            if port != 25:
-                smtp.starttls()
+            smtp.starttls()
+            smtp.ehlo()
+        elif use_ssl is False:
```
DavidMStraub: This `elif` condition doesn't make sense as `use_ssl` cannot be `False` in this branch.

ericmoret: I was trying to handle the case where EMAIL_USE_SSL=false is explicitly set (for plain SMTP). At this point use_ssl could be None or False.

DavidMStraub: Ok I missed that, but PEP 8 frowns upon `is True` for good reason - if you want to handle the case `None` specially, it would be better to have `elif use_ssl is not None`.

But I think if we separate the treatment of the deprecated config option from the main logic, we don't need this distinction here anymore, do you agree?

gramps_webapi/api/util.py
```diff
@@ -514,6 +514,30 @@ def get_buffer_for_file(filename: str, delete=True, not_found=False) -> BinaryIO
...
+def _resolve_smtp_config(
+    use_ssl: bool | None,
+    use_starttls: bool | None,
+    use_tls: bool | None,
+    port: int
+) -> tuple[bool, bool]:
+    """Helper to resolve SMTP encryption settings.
+
+    Handles backward compatibility with EMAIL_USE_TLS.
+    Can be removed once EMAIL_USE_TLS is fully deprecated.
```
DavidMStraub: Let's add to the doc string what the return values are (in which order...)

gramps_webapi/api/util.py
```diff
@@ -514,6 +514,30 @@ def get_buffer_for_file(filename: str, delete=True, not_found=False) -> BinaryIO
...
+    """
+    # If new params are set, use them
+    if use_ssl is not None:
+        return use_ssl, False
+    if use_starttls is not None:
+        return False, use_starttls
```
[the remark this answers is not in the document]
DavidMStraub: Oops - good point!

DavidMStraub:
Great, unit tests are passing! Is this ready form your side?

gramps_webapi/api/util.py
```diff
@@ -514,6 +514,29 @@ def get_buffer_for_file(filename: str, delete=True, not_found=False) -> BinaryIO
...
 
+def _resolve_smtp_config(
+    use_ssl: bool | None,
+    use_starttls: bool | None,
+    use_tls: bool | None,
+    port: int
+) -> tuple[bool, bool]:
+    """Helper to resolve SMTP encryption settings.
+
+    Return tuple[bool, bool]: (use_ssl, use_starttls)"""
```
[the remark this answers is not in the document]
DavidMStraub: Reproducing the entire function logic in the doc string - nah.

PR #741, opened 2025-12-10
Test with Python 3.14

DavidMStraub:
This pull request makes a small update to the test workflow configuration. The Python version matrix in `.github/workflows/test.yml` has been updated to test against Python 3.14 instead of 3.13.


DavidMStraub:
Doesn't work yet - need to wait for all upstream packages to support 3.14.

DavidMStraub:
Works!

Issue #743, opened 2025-12-12
[Bug] Birth and death dates lost after editing person names on NAS with Btrfs


mahula:
### Bug Description

**Environment:**
- 

- Gramps Web deployed via Docker Compose on a Synology NAS
  - 2-threaded [Intel® Celeron® Prozessor J3355](https://www.intel.de/content/www/de/de/products/sku/95597/intel-celeron-processor-j3355-2m-cache-up-to-2-50-ghz/specifications.html)
  - 10 GB RAM
  - filesystem: Btrfs (Copy-on-Write enabled)
  - docker-compose.yml
    ```yaml
    services:
      grampswebtest: &grampswebtest
        image: ghcr.io/gramps-project/grampsweb:latest
        container_name: Gramps-Web-Test
        restart: on-failure:5
        ports:
          - "5450:5000"
        environment:
          GRAMPSWEB_IGNORE_DB_LOCK: "true"
          GRAMPSWEB_LOG_LEVEL: "DEBUG"
          GRAMPSWEB_REQUEST_CACHE_CONFIG__CACHE_TYPE: "NullCache"
          GUNICORN_CMD_ARGS: "--log-level debug"
          GUNICORN_NUM_WORKERS: "1"

          # === TREE CONFIGURATION ===
          # Single tree mode: specify a tree name (will be created if not exists)
          # Multi-tree mode: use "*"
          GRAMPSWEB_TREE: "test"

          # === MULTI-TREE ONLY (uncomment if GRAMPSWEB_TREE is set to "*") ===
          # GRAMPSWEB_MEDIA_PREFIX_TREE: "true"

          # === CELERY/REDIS (Required for background tasks) ===
          GRAMPSWEB_CELERY_CONFIG__broker_url: "redis://grampswebtest_redis:6379/0"
          GRAMPSWEB_CELERY_CONFIG__result_backend: "redis://grampswebtest_redis:6379/0"
          GRAMPSWEB_RATELIMIT_STORAGE_URI: "redis://grampswebtest_redis:6379/1"

          # === USER REGISTRATION ===
          # Set to "true" to disable self-registration (users must be created by owner)
          GRAMPSWEB_REGISTRATION_DISABLED: "true"

          # === EMAIL CONFIGURATION (Required for user registration & password reset) ===
          # Uncomment and configure these for email functionality:
          # GRAMPSWEB_EMAIL_HOST: "smtp.example.com"
[...]

PR #744, opened 2025-12-12
fix: skip birth/death index recalculation when event references unchanged

mahula:
### Problem

On a NAS with Btrfs filesystem, after editing a person's names (or other non-event data),  
when navigating to the persons' page in the persons list the the birth and death dates of the person just edited are not displayed.
The API client sends the correct `birth_ref_index` but the server returns `-1`.

### Cause

The `update_object()` function unconditionally calls `set_birth_death_index()` for every Person update. This recalculates the indices by reading events from the database. On Btrfs with Copy-on-Write, reads during an active write transaction can return stale data, causing the indices to be incorrectly reset to `-1`.

### Solution

Only call `set_birth_death_index()` when the person's event references have actually changed. A new helper function `_event_refs_changed()` compares the old and new event reference lists. If unchanged, the existing indices are preserved.

### Related

- **Issue:** [#743]
- https://github.com/gramps-project/gramps/pull/2149 — adds resilience to `set_birth_death_index()` when events *are* modified

### DIFF
--- a/gramps_webapi/api/resources/util.py
+++ b/gramps_webapi/api/resources/util.py
@@ -1210,6 +1210,29 @@ def _get_class_name(super_name, key_name) -> str:
     raise ValueError(f"Unknown classes: {super_name}, {key_name}")
 
 
+def _event_refs_changed(obj_old: Person, obj_new: Person) -> bool:
+    """Check if the event_ref_list has changed between old and new Person objects.
+
+    Returns True if the event references have changed (added, removed, or modified),
+    False if they are identical.
+    """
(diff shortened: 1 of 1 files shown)


mahula:
This fix prevents here in the API the recalculation, but the filesystem nearer fix has bee proposed in the GrampsRepo https://github.com/gramps-project/gramps/pull/2149.

Sincethe issue occured, I went through a broader and deeper analysis throughout several possible causes and the the three relevant code bases `gramps`, gramps-web-api`, and ramps-web`:

These two areas were related to the reported issue and pushes as pull requests in the two separate repos.

PR #752, opened 2026-01-01
Remove mediapath from imported Gramps XML file

### LINKED ISSUE #521: Gramps XML import fails due to media path
```
werkzeug.exceptions.HTTPException: 500 Internal Server Error: Could not change media path
The opened file has media path /home/XXX/XXX/, which conflicts with the media path of the Family Tree you import into. The original media path has been retained. Copy the files to a correct directory or change the media path in the Preferences.
```

It works if I removed the `<mediapath>` tag in the XML.

DavidMStraub:
Gramps always sets the media path when importing an XML file. When the media path is already set, this fails and prevents the import from succeeding (#521). In Gramps Web, we actually don't care about the media path in the XML file as Gramps Web has its own media path (for security reasons).

Since currently there is no elegant way to disable setting the media path, this change simply modifies an uploaded file by removing its media path tag.

Fixes #521.

### DIFF
--- a/gramps_webapi/api/resources/util.py
+++ b/gramps_webapi/api/resources/util.py
@@ -1451,24 +1451,16 @@ def remove_mediapath_from_gramps_xml(file_name: FilenameOrPath) -> None:
     Args:
         file_name: Path to the Gramps XML file.
     """
-    # First, detect if file is gzipped by trying to read it
-    is_compressed = False
+    # Try to read as gzipped file first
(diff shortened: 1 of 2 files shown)


dsblank (approved):
It is a lot to have to change it in the XML (could be a very large file) but not a lot of options here.

We can remove after https://github.com/gramps-project/gramps/pull/2162 is available.

PR #817, opened 2026-04-18
Fixed broken link in Readme

eduralph:
The link on the readme page returns 404. Replaced it with a sensible alternative.


DavidMStraub:
Thanks for this! Actually the right one would be https://www.grampsweb.org/development/backend/ though

DavidMStraub:
Fixed in 1fcc0d2f04b94c686591d10ed6d2000b5fac2207.

Issue #501, opened 2024-03-22
Delete records


hartenthaler:
The new function for deleting all records in a tree is quite beneficial. However, it operates slowly, taking several minutes to delete approximately 15,000 individuals and numerous NOTE records. I suspect a timeout occurs after some time, as I had to initiate the "delete" function multiple times before it finally completed.

emyoulation:
Gramps for desktops suffered from a similar slowness because the GUI was trying to refresh after every affected record in mass deletions.

That was one of the fixes for the 5.2 release.

hartenthaler:
That happend in Gramps Web based on 5.2

DavidMStraub:
There are two separate problems here:

1) process aborts
2) process is slow.

Concerning 1): this is the Grampshub celery log for all "delete all" operations in the time frame in question:

```
[2024-03-21 18:38:48,873: INFO/MainProcess] Task gramps_webapi.api.tasks.delete_objects[53277a78-b500-407b-a0cb-e987a30d298e] received
[2024-03-21 18:41:20,829: INFO/MainProcess] Task gramps_webapi.api.tasks.delete_objects[8d1e9722-9c23-414a-b7dc-3682fb758374] received
[2024-03-21 18:41:33,746: ERROR/ForkPoolWorker-7] Task gramps_webapi.api.tasks.delete_objects[53277a78-b500-407b-a0cb-e987a30d298e] raised unexpected: DeadlockDetected('deadlock detected\nDETAIL:  Process 99749 waits for ShareLock on transaction 286824; blocked by process 99782.\nProcess 99782 waits for ShareLock on transaction 286823; blocked by process 99749.\nHINT:  See server log for query details.\nCONTEXT:  while updating tuple (3045,17) in relation "person"\n')
  File "/usr/local/lib/python3.11/dist-packages/gramps_webapi/api/tasks.py", line 297, in delete_objects
    delete_all_objects(
  File "/usr/local/lib/python3.11/dist-packages/gramps_webapi/api/resources/delete.py", line 468, in delete_all_objects
  File "/usr/local/lib/python3.11/dist-packages/gramps_webapi/api/resources/delete.py", line 48, in delete_person
    db_handle.delete_person_from_database(person, trans)
  File "/usr/local/lib/python3.11/dist-packages/gramps/gen/db/base.py", line 1889, in delete_person_from_database
[...]

PR #825, opened 2026-04-27
Speed up delete all

DavidMStraub:


DavidMStraub:
Fixes #501 (the fact that it's too slow).

@jittymolmathew92 FYI.

PR #868, opened 2026-05-29
Agent: get place and event tools

DavidMStraub:


DavidMStraub:
Reviewed offline as Github copilot has reached the monthly limit.

PR #870, opened 2026-06-03
Recompute Date sortval on write (fixes #869)

### LINKED ISSUE #869: Date `sortval` is stored verbatim on write, not recomputed from `dateval` (silently breaks date sorting)
## Summary

When writing an object that contains a `Date` (e.g. `PUT /api/events/{handle}`), the API **stores the client-supplied `date.sortval` verbatim** and does **not** recompute it from `date.dateval`. A client that sends a stale or zero `sortval` (but a correct `dateval`) silently corrupts date ordering, because `sortval` is what Gramps uses to sort dates (timelines, "born/died", chronological views).

This is easy to hit from any custom client/integration: you naturally build a date from its `dateval`/`year`, but unless you also reproduce Gramps' Julian-Day-Number math for `sortval`, the written record sorts wrong even though it *displays* correctly.

## Environment

- `gramps-webapi` **3.15.0**
- `gramps` **6.0.8**

## Steps to reproduce

```python
# event E1608 has a Birth date of 3 May 1922 -> dateval [3,5,1922,false], sortval 2423178
ev = GET /api/events/?gramps_id=E1608           # ev["date"]["sortval"] == 2423178

ev["date"]["sortval"] = 1                        # wrong sortval; dateval left unchanged ([3,5,1922,false])
PUT /api/events/{handle}  body=ev                # 200 OK

GET /api/events/?gramps_id=E1608                 # -> ev["date"]["sortval"] == 1   (NOT recomputed)
```

The same happens with `sortval: 0` and with any partially-built date (e.g. a year-only `dateval` of `[0,0,1925,false]` written with `sortval: 0`).

## Expected

On write, the server should **recompute `sortval` from `dateval`** (the source of truth), so it is internally consistent regardless of what the client sends — matching how the desktop app maintains it. For `[3,5,1922,false]` the stored `sortval` should be `2423178`.

## Actual

The client-supplied `sortval` is persisted unchanged (`1` above), so the record sorts at the wrong position while still displaying the correct date.

## Impact

- Any non-trivial write client (web integration, importer, bulk editor, AI/agent tooling) that constructs dates can silently break chronological sorting.
- The corruption is invisible in the UI's date *display*; it only shows up as mis-ordered events/people, which is hard to trace back to the write.

## Likely cause & suggested fix

`Date.unserialize()` in gramps core assigns the stored `sortval` directly and does not recalculate it (recalculation normally happens in `Date.set(...)` / `recalc_sortval()`, not on unserialize). So when the API rebuilds the object from JSON and commits it, the client `sortval` is trusted.

Suggested fix: when deserializing/validating a `Date` on write, call `date.recalc_sortval()` (or otherwise recompute `sortval` from `dateval`) before committing — ideally for any `Date` field on any object type, not just events. This keeps `sortval` authoritative on the server and makes clients robust by default.

## Workaround (client-side)

Until fixed, clients can recompute `sortval` themselves as the Julian Day Number of `(year, month or 1, day or 1)` for Gregorian dates before writing. (I verified this matches the server's values exactly for full, month-year, and year-only dates: `3 May 1922 -> 2423178`, `1999 -> 2451180`, `Dec 2006 -> 2454071`.) Happy to open a PR for the server-side recompute if that's a welcome direction.

brianmcaudill:
Fixes #869.

## Problem

On write, the API stored a client-supplied `Date.sortval` verbatim. `Date` deserialization does not recompute `sortval` from `dateval` (recalculation normally happens in `Date.set(...)`, not on unserialize), so a stale, zero, or missing `sortval` was persisted as-is. Because date sorting is driven by `sortval`, this silently mis-orders events/people while the date still *displays* correctly — hard to trace back to the write.

Repro (before this change): `PUT` an event whose `date.sortval` is `0` (with a valid `dateval`) → `GET` returns `sortval: 0`.

## Fix

`gramps_object_from_dict()` — the single deserialization chokepoint used by both the create (`/api/objects/`) and update (`PUT /api/.../{handle}`) paths — now runs the completed object dict through a new `recalc_date_sortvals()` helper before instantiation. The helper walks the dict, and for every embedded serialized `Date` lets Gramps recompute `sortval` from `dateval` via `Date.recalc_sort_value()`, so it stays correct across calendars, modifiers, and BCE dates. It covers all object types and all nested dates (event/citation/name/address/LDS-ord/place names, …), and never blocks a write on a single malformed date.

## Tests

Adds `test_recalc_date_sortvals_fixes_stale_sortval` (full date, including end-to-end through `gramps_object_from_dict`) and `test_recalc_date_sortvals_year_only`.

Verified against the running Gramps version: a zeroed `sortval` for `3 May 1922` is recomputed to `2423178`, and a year-only `1827` to `2388358` — both matching `Date.recalc_sort_value()`.

## Notes

`black` + `isort` clean. The change is confined to `gramps_object_from_dict()` plus the new helper, so existing read paths are untouched (the API GET output is unchanged; only writes now normalize `sortval`).

### DIFF
--- a/gramps_webapi/api/util.py
+++ b/gramps_webapi/api/util.py
@@ -877,9 +877,40 @@ def complete_gramps_object_dict(data: dict[str, Any]):
     return data
 
 
+def recalc_date_sortvals(data: Any) -> Any:
+    """Recompute the ``sortval`` of every embedded ``Date`` in an object dict.
+
+    Gramps Web stores whatever a client sends, and ``Date`` deserialization does not
+    recalculate ``sortval`` from ``dateval`` — so a stale, zero, or missing ``sortval``
+    is persisted verbatim and silently breaks date sorting (gramps-project/gramps-web-api#869).
+    This walks the dict and, for every serialized ``Date``, lets Gramps recompute
+    ``sortval`` from ``dateval`` (correct across calendars, modifiers, and BCE dates).
+    Modifies ``data`` in place and returns it.
+    """
+    if isinstance(data, dict):
+        if data.get("_class") == "Date" and "dateval" in data:
+            try:
+                date = data_to_object(data)
+                date.recalc_sort_value()
+                data["sortval"] = date.sortval
+            except Exception:  # pragma: no cover - never block a write on one bad date
+                pass
+        else:
+            for value in data.values():
+                recalc_date_sortvals(value)
+    elif isinstance(data, list):
+        for value in data:
+            recalc_date_sortvals(value)
+    return data
+
+
 def gramps_object_from_dict(data: dict[str, Any]):
     """Instantiate a Gramps object from a dictionary.
 
     The dictionary can be incomplete, i.e. not contain all properties of
     the object class."""
-    return data_to_object(complete_gramps_object_dict(data))
+    completed = complete_gramps_object_dict(data)
+    # Keep Date sortval authoritative on write (gramps-web-api#869): the client's
+    # value is not to be trusted, since sorting depends on it.
+    recalc_date_sortvals(completed)
+    return data_to_object(completed)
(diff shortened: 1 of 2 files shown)


DavidMStraub:
Turns out your PR is better than my local uncommitted version, well done.

Issue #599, opened 2025-01-05
More permissive match data parser


DavidMStraub:
The match data parser should smarter & more permissive, e.g. allowing MyHeritage formatted data:

```csv
Name,Match Name,Chromosome,Start Location,End Location,Start RSID,End RSID,Centimorgans,SNPs
Doe J,Joe D,1,15359967,19022911,rs4520412,rs12563055,7.6,2304
```

DavidMStraub:
Right now, the parser is based on the DNASegmentMap gramplet code by @GaryGriffin in order to ensure mutual compatibility.

@GaryGriffin, would it be ok for you if I create a new pip-installable external pure-Python library that provides a function string &rarr; segments that we can use in both the DNASegmentMap grampet and in Gramps Web API? This way we could ensure that any note that works with one also works with the other, and improvements only have to be done once. Thanks to the new addon manager, it would be enough to provide a `requires_mod` to the new library in the addon.

GaryGriffin:
Of course. The current gramplets that use this code are DNASegmentMap, DNAmatches and referenced in FamilyTreeDNA. 

Mac (and Win, I think) gramps installs  create a bundle so cannot run pip to add more libraries. They have to be bundled in the creation of gramps. I am thinking that this will mean that I cannot update DNASegmentMap until the 6.0 release to use this external lib. And will have to tell the bundlers to be sure to include this library going forward. Please provide instructions on how to test/use this when available.

I am aware of the MyHeritage format (these formats tend to change) and what I do is to import and then delete the leading strings. I cannot tie these names to grampsID, so they are useless. Or I use a spreadsheet and cut sections out and put into the correct Association Notes. Since MyHeritage puts multiple people in the file, the spreadsheet copy/paste was the most viable option I could think of.

You can also look at FamilyTreeDNA gramplet which requires a Note in the FT match to allow the match to be found in Gramps. I dont know if this gramplet is used often by people (created this summer). But it does a blind extract of the segment data and adds to the Association Note. 

Are you planning on addressing how to tie the name in the segment file to a grampsID for MyHeritage?

Nick-Hall:
> Mac (and Win, I think) gramps installs create a bundle so cannot run pip to add more libraries. They have to be bundled in the creation of gramps. I am thinking that this will mean that I cannot update DNASegmentMap until the 6.0 release to use this external lib. And will have to tell the bundlers to be sure to include this library going forward. Please provide instructions on how to test/use this when available.

[...]

PR #871, opened 2026-06-04
Parse DNA match cM values with mixed thousands and decimal separators

brianmcaudill:
## Summary

The DNA match parser's `cast_float` helper (`gramps_webapi/api/dna.py`) handles values that use a single locale separator — e.g. `"15,5"` (German decimal comma) or `"1.234.567"` (dot thousands grouping). But a value that combines **both** a thousands separator and a decimal separator was not matched by any branch and silently fell back to `0.0`:

| Input | Before | After |
|---|---|---|
| `1.234,56` (EU: `.` thousands, `,` decimal) | `0.0` | `1234.56` |
| `1,234.56` (US/UK: `,` thousands, `.` decimal) | `0.0` | `1234.56` |
| `1.234.567,89` | `0.0` | `1234567.89` |
| `1,234,567.89` | `0.0` | `1234567.89` |

For the centimorgans field this means a validly formatted value could be dropped to `0.0`.

## Root cause

`cast_float` had three independent `if` branches: multi-dot → strip dots, multi-comma → strip commas, single comma + no dot → comma is decimal. The mixed *one-dot-and-one-comma* case fell through all of them, so `float("1.234,56")` raised `ValueError` and returned `0.0`.

## Fix

Add a branch for when both `.` and `,` are present: the **rightmost** separator is treated as the decimal separator and the other is removed as a thousands separator. The remaining branches are made mutually exclusive (`elif`) but their behaviour for single-separator inputs is unchanged. This continues the locale-tolerance work from the more permissive parser (#599).

## Verification

Tests run in isolation (the full suite requires PyGObject, but `dna.py` only depends on the `MatchSegment` TypedDict, so it is exercised directly):

- Against **unfixed** code: the two new tests **fail**, all 12 existing tests pass.
- Against the **fix**: all 14 tests pass.
- `black --check` clean on both changed files.
- `mypy` introduces no new errors in `dna.py` (the function signature is unchanged).

Two regression tests are added to `tests/test_dna_match_parser.py`: one exercising `cast_float` directly across both groupings (and asserting the existing single-separator cases still hold), and one end-to-end parse asserting a mixed-separator centimorgans value yields `1234.56` rather than `0.0`.

Issue #599.

> Opened as a draft: this edge case was found while reviewing the parser rather than from a filed report, and per CONTRIBUTING I'd welcome a maintainer's call on whether mixed-separator cM values are worth handling. The fix is small and the existing tests are preserved.

### DIFF
--- a/gramps_webapi/api/dna.py
+++ b/gramps_webapi/api/dna.py
@@ -68,11 +68,19 @@ def cast_int(value: str) -> int:
 def cast_float(value: str) -> float:
     """Cast a string to a float."""
     value = value.replace(" ", "")
-    if value.count(".") > 1:
+    if value.count(".") >= 1 and value.count(",") >= 1:
+        # Both separators are present (e.g. "1.234,56" or "1,234.56"). The
+        # rightmost one is the decimal separator; the other groups thousands
+        # and is removed.
+        if value.rfind(",") > value.rfind("."):
+            value = value.replace(".", "").replace(",", ".")
+        else:
+            value = value.replace(",", "")
+    elif value.count(".") > 1:
         value = value.replace(".", "")
-    if value.count(",") > 1:
+    elif value.count(",") > 1:
         value = value.replace(",", "")
-    if value.count(",") == 1 and value.count(".") == 0:
+    elif value.count(",") == 1 and value.count(".") == 0:
         value = value.replace(",", ".")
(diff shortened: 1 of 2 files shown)


DavidMStraub:
Thank you, I think this is worth fixing!

PR #887, opened 2026-06-27
Handle FileNotFoundError when checking thumbnail file size

dsblank:
## Summary

- When `check_thumbnail_size_limit` calls `get_file_size()` for a file that doesn't exist, it now catches `FileNotFoundError` and returns early
- This allows the thumbnail handler to proceed and return a proper 404, rather than raising an unhandled exception

## Test plan

- [ ] Request a thumbnail for a media object whose file is missing — should return 404, not 500
- [ ] Request a thumbnail for an oversized file — should still return 413
- [ ] Request a thumbnail for a normal file — should return the thumbnail as before

### DIFF
--- a/gramps_webapi/api/file.py
+++ b/gramps_webapi/api/file.py
@@ -117,7 +117,11 @@ class FileHandler:
         """Abort with 413 if the file exceeds the thumbnail size limit."""
         max_bytes = current_app.config.get("MAX_THUMBNAIL_FILE_BYTES")
         assert max_bytes is not None  # for type checker
-        if self.get_file_size() > max_bytes:
+        try:
+            size = self.get_file_size()
+        except FileNotFoundError:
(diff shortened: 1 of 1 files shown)


DavidMStraub:
Good catch, this is a bug, but swallowing the exception with a comment and no mention in the doc string is not ok. "The thumbnail handler will raise" - that's action at a distance, relying on something the call site does without declaring it in the doc string.

I see a couple of cleaner options:

- check for file existence and return 404 before calling this method
- Explicitly mention in the doc string that non-existing files silently return `None`
- `abort_with_message` a 404 directly in this method

Feel free to choose the one you think makes most sense :slightly_smiling_face:

dsblank:
I went with option 3: abort_with_message a 404 directly in this method

DavidMStraub:
Oops, should have merged that before releasing 3.20. Thanks!

Issue #756, opened 2026-01-15
[Bug] Custom OIDC Role Mapping from Entra ID Not Working


switenky:
### Bug Description

When configuring OIDC authentication in Gramps Web with Entra ID via Custom OIDC provider, role mapping does not work as expected. The `roles` claim is not being recognized, and all users are assigned the guest role.

```yaml
GRAMPSWEB_OIDC_ENABLED = "True"
GRAMPSWEB_OIDC_ISSUER = "https://login.microsoftonline.com/<tenant-id>/v2.0"
GRAMPSWEB_OIDC_CLIENT_ID = "<redacted>"
GRAMPSWEB_OIDC_CLIENT_SECRET = "<redacted>"
GRAMPSWEB_OIDC_NAME = "Microsoft"
GRAMPSWEB_OIDC_SCOPES = "openid profile email"
GRAMPSWEB_OIDC_ROLE_CLAIM = "roles"
GRAMPSWEB_OIDC_GROUP_OWNER = "admin"
GRAMPSWEB_OIDC_GROUP_MEMBER = "user"
GRAMPSWEB_OIDC_USERNAME_CLAIM = "email"
```

On successful login with a user with both `admin` and `user` roles, the following warning appears:
`WARNING:gramps_webapi.auth.oidc:No 'roles' claim found in user claims. Assigning guest role.`

I see the same message with a user with just the `user` role.

The roles claim exists in the Entra ID token and works correctly in other OAuth2 apps using the same client ID/secret. So Gramps Web API does not seem to detect or map the roles claim properly for some reason.

How can I inspect the OIDC claims that Gramps receives on login? Are there logs or debug flags that can show exactly what claims are being processed? Any tips for verifying why the roles claim isn’t being mapped correctly?

Thanks.

### Gramps Web Version Information

Gramps 6.0.6
Gramps Web API 3.6.0
Gramps Web Frontend 25.12.0
Gramps QL 0.4.0
Sifts 1.1.1
locale: en
multi-tree: false
task queue: true
OCR: true
chat: false

### Verification

- [x] I have searched for existing issues before creating this one

DavidMStraub:
@switenky this is fully fixed now, right?

switenky:
@DavidMStraub Yeah, I tried linking this issue to [PR #894](https://github.com/gramps-project/gramps-web-api/pull/894) but it wouldn't let me.

I've closed this issue as completed.

switenky:
[...]

PR #894, opened 2026-07-08
Merge ID-token claims into OIDC userinfo so role mapping works with Microsoft Entra

switenky:
## Problem

For standard OIDC providers, `OIDCCallbackResource` reads user claims from the
userinfo endpoint (`oidc_client.userinfo(token=token)`). Some providers —
notably **Microsoft Entra ID** — return authorization claims such as app roles
(`roles`) and group memberships (`groups`) **only in the ID token**, never from
the userinfo endpoint.

Because `get_role_from_claims` (driven by `OIDC_ROLE_CLAIM` / `OIDC_GROUP_*`)
only sees the userinfo response, those claims are invisible, so it returns
`ROLE_DISABLED`. Since role mapping runs on every login, an Entra user with an
assigned app role is repeatedly forced to the disabled role and can never be
auto-assigned — they're stuck at "account created, an administrator will review
it."

## Fix

Merge the ID token's claims (Authlib's already-validated `token["userinfo"]`)
into the userinfo dict, filling only claims the userinfo endpoint did not
return (userinfo values take precedence). Role/group mapping then works
consistently across providers, and providers that already return these claims
from userinfo are unaffected.

## Test

Adds `test_oidc_callback_merges_id_token_claims`: a `roles` claim present only
in the ID token reaches user creation, while userinfo-endpoint values are
preserved.

## Notes

- Backwards compatible: when the ID token carries no extra claims (or the
  provider already returns them from userinfo), behavior is unchanged.
- Verified `py_compile` and `black`; I was unable to run the full test suite
  locally (heavy `gramps` / PyGObject native deps), so relying on CI for the
  complete run.

### DIFF
--- a/gramps_webapi/api/resources/oidc.py
+++ b/gramps_webapi/api/resources/oidc.py
@@ -220,9 +220,13 @@ class OIDCCallbackResource(Resource):
                 # claims missing from the userinfo response so that role
                 # mapping via OIDC_ROLE_CLAIM works consistently across
                 # providers. userinfo endpoint values take precedence.
-                id_token_claims = token.get("userinfo") or {}
-                for claim, value in id_token_claims.items():
-                    userinfo.setdefault(claim, value)
+                # Guard against a non-mapping value under "userinfo": an
+                # unexpected shape must be ignored, not turn a successful
+                # login into a 401.
+                id_token_claims = token.get("userinfo")
+                if isinstance(id_token_claims, dict):
+                    for claim, value in id_token_claims.items():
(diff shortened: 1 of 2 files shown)


switenky:
Related to #756

switenky:
Just validated the new behavior with my test Gramps Web instance:

|**Before**|**After**|
|-|-|
|Stuck hereScreenshot 2026-07-09 at 5 37 48 PM|Successfully logged in as an administrator with "admin" rolescreenshot|

DavidMStraub:
Thanks!

Issue #597, opened 2025-01-02
Limit access for user to main person and relatives


Luensche:
**Is your feature request related to a problem? Please describe.**
I have a relatively large family tree and would like to give other family members access. However, as living persons are maintained in the family tree, I would only like to grant access to the user's own person and their ancestors/descendants + persons directly related by marriage. Other people can appear anonymised with initials, so that it is clear that there are other people, but without dates etc.

**Describe the solution you'd like**
When creating a user, it should be possible to select a main person. If this is done, it should not be possible to change this in the user's own settings.
All direct ancestors+descendants plus relatives should be displayed normally, all others only with initials.

**Describe alternatives you've considered**
Multiple family trees on one instance, which have to be synced manually.

DavidMStraub:
Related discussion: https://gramps.discourse.group/t/restricting-parts-of-the-tree-from-distant-relatives/6661/7

DavidMStraub:
Also related: https://github.com/gramps-project/gramps-web-api/discussions/536

banalg:
Hi everyone,

I'm facing the exact same need: I want to share my tree with my cousins and ease collaborative updates, so I need to restrict the information each family branch can see and edit. While looking into how to design this, I read the open tickets discussing fine-grained access limitations.

I've just shared my thoughts and an architectural proposal in response to David's gist here: 
https://gist.github.com/DavidMStraub/5a7f1f4b1e374db4445baa93654b88f4?permalink_comment_id=6275667#gistcomment-6275667

Best regards,

PR #911, opened 2026-07-31
Fast SQL-pushed-down structured query endpoints for all object types

dsblank:
## Motivation

`GET /api/people/?page=1` and `?page=999` both take ~8 seconds flat, regardless of page, on a 100k-person tree — [reported on the discourse forum](https://gramps.discourse.group/t/gramps-web-api-list-performance/9007) (104s for a filtered query on 100k people). The cause: `GrampsObjectsResource.get()` unconditionally does `objects = list(iter_objects_method())` — full deserialization of *every* object in the table — before any filter, sort, or page is applied. Sorting and pagination then happen afterward, in Python, on the fully materialized list. This means the API's cost is `O(table size)` on every single request, independent of what's actually being asked for.

This PR adds a parallel, opt-in fast path: `POST /api/<type>/query/` for every primary object type, which compiles a small closed query AST (`select`/`where`/`order_by`/`limit`/`after`) directly to parameterized SQL against the secondary columns Gramps already flattens and indexes server-side (`gender`, `surname`, `given_name`, `gramps_id`, `private`, etc.) — no object deserialization, no Python-side filtering/sorting, real keyset pagination instead of materialize-then-slice.

**Verified against a live 100k-person `SharedPostgreSQL` database** (not just SQLite): the old `GET /api/people/` path reproduces the ~8s-flat bug exactly (10.8s at page=1, 7.6s at page=999); the new `POST /api/people/query/` path is flat and fast at depth — ~49ms/request walking forward via keyset to a simulated "page 999" (~19,800 rows deep), vs. 2.2s for a cold first query.

## Tree isolation on `SharedPostgreSQL` (reviewer-caught, now fixed)

`SharedPostgreSQL` stores every tree's rows in the same physical tables, discriminated only by a `treeid` column that's part of every object table's primary key. Nothing applies that filter automatically at the connection level — every one of `SharedDBAPI`'s own query methods (`get_person_handles`, etc.) adds `WHERE treeid = ?` by hand.

An earlier version of this PR changed `gramps_webapi/api/util.py`'s `is_dbapi = isinstance(self.basedb, DBAPI)` to `hasattr(self.basedb, "dbapi")`, on the theory that the `isinstance` check was an oversight — `DBAPI` and `SharedDBAPI` are sibling classes, not parent/child, so it silently excluded `SharedPostgreSQL`. It wasn't an oversight: that exclusion was the mechanism keeping `SharedPostgreSQL` on the slower but correctly tree-scoped `self.db.iter_*_handles()` path instead of `_iter_handles()`'s raw, unscoped `SELECT handle FROM <table> WHERE private=0`. A reviewer caught that this change had made a real leak reachable. **That change has been fully reverted** — `is_dbapi` is back to `isinstance(self.basedb, DBAPI)` exactly as before, `_iter_handles()` is unchanged, and `SharedPostgreSQL` stays on the existing, correct, tree-scoped path for that fast-iteration helper.

This PR's own new query compiler turned out to have the identical gap independently (new code, not a regression from the above): `compile_query()`/`compile_count_query()` built `SELECT ... FROM <table> WHERE ...` with no `treeid` filter at all — every `POST .../query/` request against `SharedPostgreSQL` would have returned rows from every tree sharing the instance, not just the caller's. Rather than excluding `SharedPostgreSQL` from these new endpoints, `treeid` scoping is now threaded through the compiler and baked into `_where_clauses()` the same non-optional way privacy is: `AND treeid = ?`, appended automatically whenever the current backend has one (`resources/object_query.py`'s `_resolve_treeid()` reads `basedb.dbapi.treeid`; `None` — no clause — on single-tree-per-database backends, which have no such column). The same fix was applied to `_resolve_after()`'s `after`-cursor lookup, which had a lower-severity version of the same gap (can't leak main-query rows, but could confirm another tree's handle exists and leak its sort-column values).

## What's new

- **`gramps_webapi/api/query.py`** — a small, closed query AST and SQL compiler. Not a general query language, not GraphQL, not raw SQL passthrough:
  ```python
  Eq(column, value); Ne(...); Lt(...); Lte(...); Gt(...); Gte(...); Like(...); In(column, values)
  And(*exprs); Or(*exprs); Not(expr)
  OrderBy(column, direction)  # "asc" | "desc"
  Query(select=[...], where=..., order_by=[...], limit=50, after=cursor)
  ```
  Every column name is checked against a fixed per-type whitelist (`ObjectTypeSpec`, derived from `get_secondary_fields()`) before the compiler touches it; values are always bound as `?` parameters. There is no path from client input to a raw SQL string. Pure, unit-testable without a running server (see `tests/test_query.py`).
- **`gramps_webapi/api/resources/object_query.py`** — `POST /api/<type>/query/` for `people`, `families`, `events`, `places`, `repositories`, `sources`, `citations`, `media`, `notes`, and `tags`. One shared `ObjectQueryResource` base class, ten thin per-type subclasses differing only in which `ObjectTypeSpec` they bind to — the same pattern `GrampsObjectResourceHelper` subclasses already use with `gramps_class_name`.
- **Keyset pagination** (`after=<cursor>`, not `OFFSET`) — `OFFSET` still degrades at depth even with SQL pushdown, since the DB has to skip N rows either way.
- **Privacy, non-optional**: every compiled query includes `AND private = 0` unless the caller has `PERM_VIEW_PRIVATE` — baked into the compiler, not a query option, so it can't be omitted by a malformed or malicious request. (`Tag` is the one type with no `private` column; handled via a derived `has_privacy` flag rather than special-casing call sites.)
- **Locale-aware sorting** — reuses gramps core's existing `Connection.check_collation()` (already used internally by `get_person_handles(sort_handles=True, locale=...)`) rather than adding new migration machinery. Opt-in via an explicit `locale` param; deliberately *not* applied by default, since falling back to the system locale would silently change sort order for every caller who never asked for it.
- **Refuses to bypass any proxy other than `PrivateProxyDb`.** The endpoint reaches raw SQL by unwrapping `db.basedb`, which skips whatever filtering a wrapping proxy applies. That's only actually safe for `PrivateProxyDb`/`ModifiedPrivateProxyDb`, since its `private`-flag filtering is independently reimplemented in SQL above. Any other proxy (`LivingProxyDb`, `FilterProxyDb`, `ReferencedBySelectionProxyDb`) has no SQL equivalent here, so silently unwrapping past one would return exactly the data it exists to hide. `get_db_handle()` never constructs anything other than the raw db or `PrivateProxyDb` today — this isn't a fix for a currently-reachable bug — but the endpoint now checks explicitly and refuses (`501`) rather than relying on that invariant holding implicitly.
- **Optional total count via `X-Total-Count`** (`"count": true` in the request, default `false`) — runs a second `SELECT COUNT(*)` with the same `where`/privacy predicate (ignoring `select`/`order_by`/`limit`/`after`, which don't affect how many rows match), and returns it in the `X-Total-Count` response header — the same convention `objects.py`/`emit.py`/`history.py` already use elsewhere in this API, rather than a new body field (reviewer feedback: an earlier version of this PR added a redundant `total_count` body field instead of reusing the existing convention). Still opt-in and off by default, unlike those endpoints, since it's a genuinely separate query here rather than a free byproduct of already-materialized data. `compile_query()`/`compile_count_query()` share a `_where_clauses()` helper so the two can't drift on privacy/tree-scoping enforcement.
- **`JsonPath`** — a path into a JSON blob column (`json_data`), for querying fields not flattened to a real secondary column (e.g. `primary_name.surname_list[0].surname`). Usable in `select` (`{"json_path": [...], "as": "<key>"}` — `as` is optional, defaulting to a derived dotted/bracket key like `primary_name.surname_list[0].surname`) and `where` (`{"column": {"json_path": [...]}, "op": ..., "value": ...}`). Not (yet) usable in `order_by`/keyset pagination. Backend-dialect-aware: `json_extract(...)` on SQLite; `jsonb_extract_path_text(...)` on PostgreSQL, or the non-`_text` `jsonb_extract_path(...)` + `CAST(... AS NUMERIC/BOOLEAN)` when the comparison's own Python value is numeric/boolean (`jsonb_extract_path_text` always returns `TEXT`, which compares lexicographically, not numerically). Dialect is resolved per-request from the live backend (`resources/object_query.py`'s `_resolve_dialect()`), not hardcoded. See `query.py`'s `Dialect`/`_render_json_path`/`JsonPath` and `tests/test_query.py`.
- **Tree-scoped on `SharedPostgreSQL`** — every compiled query includes `AND treeid = ?` on that backend, non-optional, the same way privacy is. See the dedicated section above.
- **`gramps_webapi/api/query_lang.py`** — an "almost Python" expression parser, so callers don't have to hand-build the `where` array. `"gender == 1 and primary_name.surname_list[0].surname == 'Smith'"` parses into exactly the same JSON leaf-condition list `where` already accepts. Uses `ast.parse(expr, mode="eval")` as pure syntax — the tree is inspected and translated, **never** `eval()`/`compile()`/`exec()`'d. Safety is node-shape whitelisting, not name blacklisting: any AST node this module doesn't explicitly handle (arbitrary function calls, `lambda`, comprehensions, f-strings, walrus, ...) falls straight through to a parse error rather than being interpreted. Wired into every `POST .../query/` endpoint as an optional `where_expr` string field, mutually exclusive with `where` (a request setting both is rejected, `422`) — see examples below. Python's own grammar already treats `x in [1, 2, 3]` as the same `Compare` node shape as `==`/`<`, so the `In` operator comes for free; `Like` has no Python operator, so it gets exactly one whitelisted call form, `like(path, 'pattern%')` — the only function call the parser accepts at all. Scope matches what the wire format supports today: a conjunction of comparisons only (`a == b and c > d and ...`) — `or`/`not` are rejected with a clear error rather than silently mishandled, since the current `where` shape is a flat, always-AND'd list with no JSON representation for boolean nesting yet. On the value side of a comparison, `ClassName.CONST` (e.g. `gender == Person.MALE`, `format == Note.FLOWED`, `confidence >= Citation.CONF_HIGH`) resolves to the real value read off the actual Gramps class, not a hardcoded int, so it can't drift out of sync with core — scoped to the fields already reachable as a flat column today (`Person.gender`, `Citation.confidence`, `Note.format`).

## Example usage

```bash
# First 50 people, all columns
curl -X POST https://your-server/api/people/query/ \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{"limit": 50}'

# Filtered + sorted: female people, sorted by surname, only handle/name columns
curl -X POST https://your-server/api/people/query/ \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{
    "select": ["handle", "given_name", "surname"],
    "where": [{"column": "gender", "op": "eq", "value": 1}],
    "order_by": [{"column": "surname", "direction": "asc"}],
    "limit": 50
  }'

# Next page: pass back the previous response'"'"'s next_after as the cursor
curl -X POST https://your-server/api/people/query/ \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{
    "select": ["handle", "surname"],
    "order_by": [{"column": "surname", "direction": "asc"}],
    "limit": 50,
    "after": "<next_after from previous response>"
  }'

# Locale-aware sorting
curl -X POST https://your-server/api/people/query/ \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{
    "select": ["surname"],
    "order_by": [{"column": "surname", "direction": "asc"}],
    "locale": "de",
    "limit": 50
  }'

# Same pattern for every other type, e.g. events with a LIKE filter
curl -X POST https://your-server/api/events/query/ \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{
    "select": ["handle", "description"],
    "where": [{"column": "description", "op": "like", "value": "Birth%"}],
    "limit": 50
  }'

# JsonPath: reach into json_data for fields with no flat secondary column
curl -X POST https://your-server/api/people/query/ \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{
    "select": [
      "handle",
      {"json_path": ["primary_name", "surname_list", 0, "surname"], "as": "surname"}
    ],
    "where": [
      {"column": {"json_path": ["gender"]}, "op": "eq", "value": 1}
    ],
    "limit": 50
  }'

# where_expr: same query as above, written as an expression instead of
# hand-building the "where" array -- mutually exclusive with "where"
curl -X POST https://your-server/api/people/query/ \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{
    "select": ["handle", "gramps_id", "given_name", "surname"],
    "where_expr": "gender == 1 and primary_name.surname_list[0].surname == '"'"'Smith'"'"'",
    "order_by": [{"column": "given_name", "direction": "asc"}],
    "limit": 50
  }'

# where_expr also supports "in" and a whitelisted like(...) call
curl -X POST https://your-server/api/events/query/ \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{
    "where_expr": "like(description, '"'"'Birth%'"'"')"
  }'

# where_expr with Person.MALE-style constants, instead of magic numbers
curl -X POST https://your-server/api/people/query/ \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{
    "where_expr": "gender in [Person.MALE, Person.OTHER]"
  }'

# birth.date.sortval: select the full Date struct, filter by sortval range
curl -X POST https://your-server/api/people/query/ \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{
    "select": ["handle", "given_name", "surname", {"json_path": ["birth", "date"]}],
    "where_expr": "gender == Person.MALE and birth.date.sortval >= Date('"'"'Jan 1, 1968'"'"') and birth.date.sortval < Date('"'"'Jan 1, 1969'"'"')",
    "order_by": [{"column": "surname", "direction": "asc"}],
    "limit": 50
  }'

# Relationship-crossing paths: Family -> Person (father/mother), and a
# real 2-hop chain, Person -> Event -> Place
curl -X POST https://your-server/api/families/query/ \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{
    "where_expr": "father.surname == '"'"'Smith'"'"' and mother.surname == '"'"'Jones'"'"'"
  }'

curl -X POST https://your-server/api/people/query/ \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{
    "where_expr": "birth.place.title == '"'"'Chicago'"'"'"
  }'

# Field-vs-field: compare two relationship paths to each other, not just to
# a literal -- "families where the mother died before the father"
curl -X POST https://your-server/api/families/query/ \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{
    "where_expr": "mother.death.date.sortval < father.death.date.sortval"
  }'

# Same query, as raw "where" instead of "where_expr" -- "value_column" is
# the wire-format equivalent of a path-shaped comparison RHS
curl -X POST https://your-server/api/families/query/ \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{
    "where": [{
      "column": {"json_path": ["mother", "death", "date", "sortval"]},
      "op": "lt",
      "value_column": {"json_path": ["father", "death", "date", "sortval"]}
    }]
  }'
```

Response shape (all endpoints):
```json
{
  "items": [{"handle": "...", "surname": "...", ...}],
  "next_after": "<cursor for the next page, or null if this was the last page>"
}
```
Plus an `X-Total-Count` response header with the total matching row count, present only when the request set `"count": true`.

## Explicitly out of scope for this PR

- **Relationship-graph filter rules** (`IsDescendantOf`, `HasCommonAncestorWith`, etc.) — need either the existing slow path or a per-rule override mechanism at a different layer (tracked separately, `dsblank/gramps#2177`).
- **`Like` case-sensitivity is not portable across backends** — SQLite's `LIKE` is case-insensitive for ASCII by default, PostgreSQL's is case-sensitive. A separate addon-side fix (translating `LIKE`→`ILIKE`) is the right place for this, not gramps-web-api.
- **`SharedPostgreSQL`'s un-migrated `_hack_query()`** — confirmed still present in the deployed addon (not yet ported from the fix already merged for the single-user `PostgreSQL` addon in `gramps-project/addons-source#943`). Its behavior was accounted for (verified live) rather than worked around, since it's self-correcting for every column this PR's whitelists actually use.
- **`JsonPath` in `order_by`/keyset pagination** — not supported yet; `order_by` stays plain-column-only for now.
- **Relationship paths in `order_by`/keyset pagination** — same reasoning as `JsonPath` above; see the dedicated section below for what *is* supported.
- **Relationship-path range-query performance** — filtering by these fields is not indexed (see below); a range (two comparisons against the same field) currently re-executes the underlying subquery once per comparison rather than sharing one evaluation, doubling the cost for that case specifically. Not fixed here.

## Relationship-crossing paths: `select`/`where`/`where_expr` can follow Family→Person, Person→Event, Event→Place

`Person` has no birth/death date of its own — only `birth_ref_index`/`death_ref_index`, an index into `event_ref_list` saying which entry is the birth/death event; the actual date lives on that separate `Event` row. `Family.father_handle`/`mother_handle` and `Event.place` are the same shape of problem, one step simpler (a direct foreign key, not a dynamic index). All three are handled by one general mechanism: `query.py`'s `RelatedObject` + a small relationship registry (`birth`/`death`: Person→Event; `father`/`mother`: Family→Person; `place`: Event→Place) + one recursive resolver (`resolve_column_path()`) that both the raw wire format and `where_expr` funnel through — adding a new relationship is one registry entry, not new rendering code. Resolved via a **correlated scalar subquery** at each hop, not a `JOIN` — deliberately, so it needs no alias/qualification changes anywhere else in the compiler (each subquery is a fully independent scope, correlated back by the table's own unaliased name; confirmed live this holds even when the same table is hit twice in one query — e.g. father's and mother's `person` lookups — and when subqueries nest for a multi-hop chain).

- **A path is dotted**, e.g. `birth.date.sortval`, `father.surname`, or a genuine 2-hop chain, `birth.place.title` (Person→Event→Place) — all verified live end-to-end, including the case where the referenced event's place is unset, correctly resolving to `null`.
- **`select: [{"json_path": ["birth", "date"]}]`** returns the full `Date` struct (`format`/`calendar`/`modifier`/`quality`/`dateval`/`text`/`sortval`) of the referenced event, or `null` if none is recorded — a UI needs the whole struct to render partial/approximate/non-Gregorian dates correctly, not just a sort key. `where`/`where_expr` on a date compares against `sortval` only — the one property Gramps itself treats as chronologically orderable. (`dateval`'s raw `(day, month, year, slash)` tuple isn't safely comparable as a whole — SQL would sort by day first, not year — and `modifier`/`quality`, which distinguish before/about/exact, collapse to the same `sortval` regardless: a known, documented limitation, not fixed here.) A `.sortval`/`.dateval`/etc. suffix past a bare relationship name in `where_expr` (`birth.sortval` instead of `birth.date.sortval`) is rejected explicitly rather than silently falling through to a meaningless `JsonPath`.
- **The direct-FK relationships (`father`/`mother`/`place`) need no `-1`-sentinel guard at all** — unlike `birth`/`death`'s dynamic index into `event_ref_list`, a `NULL` handle already fails the handle-equality naturally.
- **Two real correctness bugs found and fixed while building this, both verified live against SQLite and PostgreSQL 16, not just asserted**: (1) `birth_ref_index`/`death_ref_index` of `-1` ("no such event recorded") must be guarded explicitly — PostgreSQL's `->` operator treats a negative array index as "count from the end," so without the guard a person with no birth event but *some* other event would silently get that unrelated event's date reported as their birth date. (2) The same numeric-vs-text comparison bug `JsonPath` already had reopens here on PostgreSQL (`->>'sortval'` gives `TEXT`, compared lexicographically) — fixed the same way, `CAST(...AS NUMERIC)` via the non-text `->` operator.
- **Privacy and `treeid` scoping apply to each subquery's own row**, not the outer query — a private/other-tree related row with no view permission makes that field come back `null`, not exclude the outer row from the results entirely.
- **Benchmarked live at 100k rows** (SQLite and PostgreSQL): cheap when the filter is unselective and `LIMIT`-bound (sub-millisecond to a few ms), but scales toward the full row count once the filter is selective or `count: true` is requested, since there's no index on the computed value — unlike every other filter in this compiler. Still meaningfully faster than the ~8s-flat bug this PR fixes even in the worst measured case (~570ms/100k rows on PostgreSQL), but a different performance class than the rest of this compiler.
- A real materialized column (indexed, keyset-paginatable) for any of these would need a **Gramps core** change — the same trick already used for `given_name`/`surname` — with a genuine open invalidation problem for the `Event`/`Place`-backed ones (the cached value is owned by a *referenced* row, so editing that row needs to recompute whatever references it). Out of scope for this PR.

## Field-vs-field comparisons: comparing two relationship paths to each other

`where`/`where_expr` conditions could previously only compare a column/path against a literal (`path OP value`, right-hand side always bound as a parameter). This adds `path OP path`: a `column` can be compared directly against another column or relationship path via a new `value_column` key (mutually exclusive with `value`), so "families where the mother died before the father" — `mother.death.date.sortval < father.death.date.sortval` in `where_expr` — is now a single query, not something the client has to fetch both sides for and compare in Python.

- **Rendering**: `Comparison.compile()` renders `value_column` by calling the same `_render_column` machinery used for the LHS on the RHS too, and joins the two directly with the operator (`<lhs subquery> < <rhs subquery>`) — no `?` placeholder between them. The two sides are independent, sibling correlated subqueries (confirmed live: they don't collide even when both hit the same table unaliased, e.g. mother's and father's `person` lookups in the same query).
- **`where_expr` syntax**: the parser (`query_lang.py`) detects a path-shaped comparison right-hand side (`_is_path_node`) and emits `value_column` instead of `value` automatically — no new syntax to learn, `a.b.c < d.e.f` just works. The one genuinely ambiguous shape, a single-level `Attribute(Name, attr)` (`Person.MALE` the constant vs. `father.surname` the path), is disambiguated the same way the existing constant-resolution code already does: whether the base name is a registered constant class.
- **The numeric-cast problem**: `JsonPath`/`RelatedObject` rendering picks TEXT vs. numeric/boolean `CAST` by inspecting the *type* of the comparison's Python-side `value` — but a field-vs-field comparison has no literal value to inspect on either side. Ordering operators (`<`/`<=`/`>`/`>=`) between two paths now pass a dummy `int` (`0`) as that hint on both sides, which reuses the existing cast-selection logic unmodified (it only ever looks at the hint's type, never its content) and picks numeric casting, matching what's overwhelmingly the actual case (comparing two `sortval`s or similar). Equality (`==`/`!=`) doesn't need the hint — TEXT extraction is correct there regardless of underlying type, since both sides serialize identically.
- **Verified live that the cast is load-bearing, not just theoretical**: crafted a `mother.sortval=9` vs. `father.sortval=10` pair specifically because `'9' > '10'` is true lexicographically but `9 < 10` is true numerically — confirmed PostgreSQL gives the wrong answer without the `CAST(...AS NUMERIC)` and the right one with it.
- **Benchmarked live against the same real 100k-person `SharedPostgreSQL` database** referenced elsewhere in this PR (46,315 families, `treeid`-scoped): `mother.death.date.sortval < father.death.date.sortval` — 9,607 matches in 2.86s full-table (each side is two nested correlated subqueries, no index on the derived value, consistent with the relationship-path performance caveat above); a flat-column field-vs-field comparison, `mother.surname == father.surname`, is far cheaper at 396ms since both sides hit the indexed `surname` column directly.
- Not supported for `in`/`like` (rejected with `422`) — comparing a path against a *list* of other paths, or LIKE-matching one path's pattern against another's value, aren't meaningful operations here and were never asked for.

## NULL-safe equality: `==`/`!=` now use `IS [NOT] DISTINCT FROM`, not `=`/`!=`

Found while live-verifying field-vs-field comparisons on the 100k-person dataset: `birth.place.title == death.place.title` (18,160) and `birth.place.title != death.place.title` (81,840) didn't sum to 100,000 — they summed to 47,095, less than half. Plain SQL `=`/`!=` use three-valued logic: if either side is `NULL`, the comparison is `UNKNOWN`, not `TRUE`/`FALSE`, so a person missing a birth or death place (52,905 of them here) satisfied *neither* `==` nor `!=` and was silently dropped from both counts — exactly backwards for a comparison whose entire point is "are these different."

- **Fix**: `Eq`/`Ne` now render as `IS NOT DISTINCT FROM`/`IS DISTINCT FROM` instead of `=`/`!=` — NULL-safe equality, where `NULL` is a normal, comparable value (`NULL IS DISTINCT FROM 'x'` is true; `NULL IS DISTINCT FROM NULL` is false). After the fix, the same two counts are 18,160 + 81,840 = **100,000** exactly.
- **Scope: applies everywhere `==`/`!=` are used**, not just field-vs-field — literal comparisons (`gender == 1`) get the same rewrite, for the same reason (a `!=` search should surface missing data as "different," not hide it). `<`/`<=`/`>`/`>=`/`LIKE`/`IN` are unaffected — no natural NULL-safe equivalent exists for ordering in standard SQL, and changing them wasn't asked for.
- **Known consequence, worth knowing about**: `>=` is no longer equivalent to `> OR ==` when either side can be `NULL`, because the two operator families now use genuinely different NULL semantics. Verified live: for `mother.death.date.sortval` vs. `father.death.date.sortval` on the 100k-family dataset, `>=` (9,629) minus `>` (9,628) is 1 — the true count of families where both dates are recorded *and* exactly equal — but `==` returns 5,910, because it also counts every family where *neither* parent's death date is recorded at all (`NULL IS NOT DISTINCT FROM NULL`). A caller who assumes `>=` decomposes into `> OR ==` will get a very different (much larger) set from `==` alone than they'd expect.
- **Requires SQLite 3.39+** (2022-06-25) for `IS DISTINCT FROM` support — a slightly newer floor than Gramps core's own existing minimum (3.35.0, for `DROP COLUMN`). Standard, unconditionally supported on PostgreSQL. The dev/CI environment here runs 3.45.2.
- **Verified live on both backends**, real 100k-row data, not just SQL-string assertions: counts match exactly between SQLite and PostgreSQL before and after the fix, and the fixed counts now correctly sum to the full row count (100,000 people; 46,315 families) for every `==`/`!=` pair checked.

## Privacy guard: NULL-safe equality can't be allowed to compare a privacy-masked field as if it were real data

A direct follow-on from the NULL-safe equality fix above. A privacy-masked related field resolves to `NULL` (see the relationship-crossing-paths section: "a private related row with no view permission makes this field come back `null`"). Once `==`/`!=` became NULL-safe, that masked `NULL` started participating in comparisons as if it were genuinely missing data — which is wrong specifically for privacy, in a way it isn't for "no birth event recorded": the caller isn't shown *why* the value is missing, so a comparison result that varies depending on the real (hidden) value is a correctness problem, not just a UX quirk.

Found live, both directions: comparing two privacy-masked fields to each other read as **equal** (`NULL IS NOT DISTINCT FROM NULL`); comparing one masked field to a literal it genuinely matched read as **not equal** — e.g. a father marked private whose real surname is "Smith" made `father.surname != 'Smith'` match, confidently asserting something false about data the caller was never shown.

- **Fix**: `Eq`/`Ne` now render as `CASE WHEN <visibility guard> THEN (<comparison>) ELSE NULL END` for any `RelatedObject` side that crosses a privacy-bearing relationship, whenever the caller can't view private data. The guard (`_related_object_visible_sql()`) is a per-hop, correctly-nested correlated subquery — mirrors `_render_related_object`'s own nesting exactly, so a multi-hop chain (`birth.place.title`) gets a guard at *each* hop, correlated to its own parent scope. `select` is untouched — a masked field there already correctly read as `null`; this only changes `where`.
- **"Blocked" renders as SQL `NULL`, not `FALSE`** — this is the subtle part. `(guard) AND (comparison)` collapses to a definite `FALSE` when blocked, and `NOT FALSE` is `TRUE` — wrapping a guarded `Eq` in `Not(...)` would have silently re-opened the exact same leak through a different path. Representing "blocked" as `NULL` instead means it propagates correctly through `NOT`/`AND`/`OR` via SQL's own three-valued logic with zero special-casing needed elsewhere in the compiler — verified live for `Not`, `And`, and `Or`, including confirming a blocked branch inside an `Or` doesn't suppress a sibling branch that's independently visible and genuinely true (`NULL OR TRUE` is `TRUE`).
- **Proved this can't leak the underlying value** (not just "rows get excluded," which alone doesn't rule out a subtler channel): swept a masked field against 13 different literal probes — the true value, near-misses, case variants, a SQL-injection string — under both `==` and `!=`; every probe produced the identical result. The response is structurally incapable of correlating with what's being guessed, because `_related_object_visible_sql()`'s signature never takes the compared value as an argument in the first place — it's not filtered out after the fact, it was never in scope.
- **`treeid` scoping uses the identical subquery mechanism (`AND treeid = ?`) and was checked for the same class of issue, but doesn't need an equivalent guard**: the same "compared value never enters the subquery's own `WHERE`" argument applies unconditionally, and was verified live even under a deliberate cross-tree handle collision (two different rows sharing the same handle in two different trees) — the caller's own tree's row (if any) is used normally; if only the other tree's row exists, the result is indistinguishable from that row not existing at all.
- **Cost**: adds a second correlated subquery per privacy-bearing hop, only when `can_view_private` is `False`. Benchmarked live on the 100k-person dataset: roughly +15-20% on SQLite, +40-50% on PostgreSQL for the `birth.place.title == death.place.title` two-hop case specifically.

## Test plan

- [x] `tests/test_query.py` — pure compiler/AST unit tests (whitelist enforcement, privacy baked-in, keyset seek-expansion, collation, reserved-word quoting, `compile_count_query()` ignoring select/order_by/limit/after, `JsonPath` rendering for both dialects including the numeric/boolean `CAST` cases, `treeid` scoping baked into `WHERE` the same way privacy is, `resolve_column_path()`/`RelatedObject` for `birth`/`death`/`father`/`mother`/`place` in both `select` and `where` for both dialects incl. the `-1`-guard and numeric-cast regressions, a 2-hop chain (`birth.place.title`), sibling-subquery-same-table (`father`+`mother` in one query) and nested-subquery non-collision, real SQLite/end-to-end execution tests proving the *dynamic* per-row `event_ref_list` index picks the right entry and a direct-FK chain (`father.surname`) filters correctly, and field-vs-field comparisons (`value_column`): SQLite/PostgreSQL wire shape, the numeric-cast hint firing for ordering ops and not equality, a 2-hop chain compared against a literal alongside a field-vs-field comparison in the same query, privacy/`treeid` scoping applying independently to each side, and a real SQLite end-to-end execution test with a deliberately crafted 9-vs-10 sortval pair proving the numeric cast is load-bearing, not just theoretical; and NULL-safe equality: `Eq`/`Ne` render as `IS [NOT] DISTINCT FROM` for both literal and field-vs-field comparisons, ordering/`LIKE`/`IN` are unaffected, and a real SQLite end-to-end execution test with a "one side missing" family and a "both sides missing" family proving `Ne` includes the former and excludes the latter, and `Eq` the reverse; and the privacy guard on NULL-safe equality: SQL-shape checks (guard present without view-private permission, absent with it, absent for ordering ops), a live SQLite execution test reproducing the field-vs-field and field-vs-literal leaks found live, a 2-hop nested-chain regression test for the guard's own correlation bug (an earlier draft's flat top-level guard referenced an out-of-scope table two hops down -- caught only by executing the SQL, not by string assertions), a genuinely-missing-data test proving the guard doesn't over-block, and `Not`/`Or` composition tests proving "blocked" survives negation and doesn't suppress an independently-visible sibling in an `Or`)
- [x] `tests/test_object_query_parsing.py` — pure unit tests for the request-parsing helpers (`json_path`/`as`-alias parsing, duplicate-select-key and reserved-`"handle"`-alias rejection, dialect resolution incl. real-`SQLite`-instance detection, `treeid` resolution, `where`/`where_expr` mutual-exclusivity, relationship-crossing paths resolving correctly for both `select` and `where`, `_terminal_is_json_path`/`_normalize_json_value` correctly scoping response normalization to JSON-sourced values only — not a plain flat column like `father.surname`, which never touches a JSON function at all, and `_build_where`'s `value`/`value_column` handling: field-vs-field builds the expected `Comparison`, `value`+`value_column` both present or both absent is rejected, `value_column` rejected for `in`/`like`) — deliberately independent of the Flask app/test client, since the shared endpoint test fixture pulls in a `sentence-transformers` model load that's broken in some dev environments
- [x] `tests/test_query_lang.py` — pure unit tests for the expression parser: every comparison operator, `in`, the whitelisted `like(...)` call, `and` conjunctions, plain-column-vs-`JsonPath` resolution, namespace resolution (`"person"`/`"Person"`), `ClassName.CONST` constant resolution (`Person.MALE`, `Citation.CONF_HIGH`, `Note.FLOWED`) incl. unknown-constant-name/deep-chain rejection, `Date('...')` parsing to `sortval` incl. unparseable-input rejection, relationship-crossing dotted paths (`birth.date.sortval`, `father.surname`, `birth.place.title`) in `where_expr`; explicit rejection tests for `or`/`not`/`not in`/`is`/chained comparisons (no wire-format equivalent yet) and a dedicated safety block confirming arbitrary code (`__import__(...)`, arbitrary function calls, `lambda`, comprehensions, f-strings) is rejected, not silently ignored or executed; and field-vs-field (`_is_path_node`): every operator between two paths, flat-column-vs-flat-column, a `Subscript` RHS, `in` rejecting a bare path RHS, `Person.MALE` still resolving as a constant (not misfiring as a path) alongside `and`-combined path/literal conditions, plus two now-permissive cases documented as intentional rather than silently left broken — an unrecognized single-level attribute base (`Foo.BAR`) or a multi-level chain (`a.Person.MALE`) is treated as a path in a comparison (same deferred-validation philosophy as any other unrecognized relationship name), while still strictly rejected inside `in [...]` lists, which never consult path detection
- [x] `tests/test_endpoints/test_people_query.py` — integration tests against `example_gramps` (SQLite): filtering, sorting, keyset pagination correctness (no gaps/overlap), privacy exclusion, locale collation, non-`PrivateProxyDb` proxy refusal, `X-Total-Count` (omitted by default, matches full result set not just the page, respects `where`, stable across pages), regression test for a real bug caught mid-session (default locale silently changing sort order)
- [x] `tests/test_endpoints/test_object_query.py` — cross-type smoke tests for the other 9 endpoints, including dedicated regression coverage for `Media.desc` (a reserved SQL word on PostgreSQL)
- [x] Live-verified against a real 100k-person `SharedPostgreSQL` database — performance, correctness, and privacy filtering all confirmed with real data, not just `example_gramps`. **Caveat**: this was a single-tree test instance, run before the `treeid` fix above existed, so it did not — and could not have — exercised cross-tree isolation.
- [x] **Live-verified with 2 real trees on the same `SharedPostgreSQL` instance** (disposable Docker Postgres 16, not the unit-test mocks): two trees created via `SharedPostgreSQL.load()`, each with real `Person` records added through gramps' own `db.add_person()` (not hand-crafted SQL), confirming:
  - `compile_query()`: each tree's `treeid`-scoped query returns only its own rows; the *same* query with `treeid=None` (the pre-fix code path) provably returns rows from every tree sharing the instance -- reproducing the exact leak the reviewer found, not just asserting the fix in the abstract.
  - `compile_count_query()`: scoped count matches the caller's own tree; unscoped count is higher, confirming `X-Total-Count` would have leaked a cross-tree total too.
  - `_resolve_after()`: resolving an `after` cursor using another tree's handle is correctly rejected (`422`) when tree-scoped.
  - `isinstance(SharedPostgreSQL_instance, DBAPI)` is `False`, confirming `util.py`'s reverted `is_dbapi` check still correctly excludes `SharedPostgreSQL` from `_iter_handles()`'s raw-SQL path.
- [ ] Not yet run against `SharedPostgreSQL` in CI (CI currently doesn't run a Postgres service at all)


DavidMStraub:
:scream: `isinstance(basedb, DBAPI)` :exploding_head: 

I'll fix that immediately in master before looking in detail at your feature, which looks very promising!

I didn't attack this so far as it's not the main bottleneck for "few thousand" people trees, but indeed it's the main scalability limit so it would be awesome to finally unblock Web API from working with huge trees, which would make it very attractive for genealogical societies.

DavidMStraub:
One issue to be wary of: the most wanted feature is more granular permissions (https://github.com/gramps-project/gramps-web-api/issues/597), and it's still not clear what the database representation will be (tags, filters, something else...). We should make sure the speedup changes don't make this feature even harder to implement.

DavidMStraub:
> A lighter-weight primitive (a pooled/reused raw .dbapi connection, skipping DbState/undo-manager/name-format setup entirely) would be a meaningful follow-up, but is a separate, larger architectural change affecting the whole API, not something to bundle into this PR.

Claude is overreaching here. We also want to continue sharing as much code as possible with Gramps core and not rewrite the whole database access layer from scratch. My https://github.com/gramps-project/gramps/pull/2229 will reduce this overhead significantly.

DavidMStraub:
The `total_count` is redundant, the endpoints return an `X-Total-Count` header.

Please separate performance optimization and API changes into separate PRs. Thanks!

DavidMStraub:
Your fix to the `isinstance` issue would leak data from all trees in a SharedPosgreSQL instance: `"SELECT handle FROM %s WHERE private=0" % table` with no `tree` filter.

That's exactly the security/privacy issues that I'm scared of when writing database primitives to improve performance, bypassing the Gramps core database layer.

dsblank:
@DavidMStraub thanks for looking at this already... still in development. This is a draft. 

> One issue to be wary of: the most wanted feature is more granular permissions (https://github.com/gramps-project/gramps-web-api/issues/597), and it's still not clear what the database representation will be (tags, filters, something else...). We should make sure the speedup changes don't make this feature even harder to implement.

The endpoints here should not affect decisions in the overall design, but the other way. If more granular permissions are added, then these endpoints will have to change (or they just won't support that use--checks are already there).

dsblank:
@DavidMStraub It looks like the isinstance() dbapi "bug" isn't a bug. It was designed to not work on SharedPostgresql as iter_handles() doesn't respect treeid. Anyway, I've removed that issue from this PR.

dsblank:
> The total_count is redundant, the endpoints return an X-Total-Count header.

These are new endpoints, and I changed them to use this name (when count is requested)

dsblank:
> That's exactly the security/privacy issues that I'm scared of when writing database primitives to improve performance, bypassing the Gramps core database layer.

You were right that getting the privacy secure without leaking subtle evidence was indeed hard. But I believe I have a provably-correct system.

I didn't intend to make this PR a full-coverage of WHERE filtering, but I am thinking that these endpoints can form the basis of a very powerful set of functions for a webapp.

gramps_webapi/api/util.py
```diff
@@ -181,6 +191,12 @@ def get_media_from_gramps_id(self, val):
     def _iter_handles(self, obj_key):
         """
         Return an iterator over handles in the database
+
+        Only reachable when `self.is_dbapi` is `True` (plain `DBAPI`
+        backends -- `SQLite`, the single-user `PostgreSQL` addon), which
+        deliberately excludes `SharedPostgreSQL`: this raw query has no
+        `treeid` filter, and `SharedPostgreSQL` stores every tree's rows in
+        the same physical tables, discriminated only by `treeid`.
```
dsblank:
```suggestion
```

DavidMStraub:
Thanks! Here is Claude Opus's review, after I've fed it with all my German angst :wink:

----

# Review of #911 — fast SQL-pushed-down query endpoints

**Scope reviewed:** 24 commits, 11 files, +5,407 / −16 lines.

**Summary:** The capability is worth having in Web API, and worth having in a form that
serves the application being built on top of it. Merging it as a single unit is not
advisable. What follows is an assessment of what the PR does well, one blocking issue
that has not yet come up in the thread, and a set of recommendations — several of which
concern *where* this code should live rather than what it does.

---

## What the PR does well

Worth stating first, because most of what follows is critical.

- The `treeid` scoping is **thorough**: not just the main query but the `COUNT`, the
  `RelatedObject` correlated subqueries, and the `after` cursor lookup in
  `_resolve_after`. That last one is subtle, would have been a real leak on its own, and
  would have been easy to miss.
- The proxy-bypass guard in `object_query.py` — refusing with 501 under any proxy other
  than `PrivateProxyDb`, rather than assuming `get_db_handle()` will never return
  anything else — is exactly the right instinct.
- Values are consistently bound as parameters, columns are whitelisted against
  `spec.columns`, and privacy/tree predicates are built in one shared place
  (`_where_clauses`) so the query and count paths cannot drift apart.
- The comments explaining *why* each backend quirk is handled the way it is are better
  than most of what is already in this repository.
- The expression language in `query_lang.py` is a substantial piece of design work, and
  clearly more capable than gramps-ql. That is why recommendation R4 is about giving it a
  home rather than replacing it.

## 1. What it adds

Ten new endpoints, `POST /api/<type>/query/`, for person, family, event, place,
citation, source, repository, media, note, tag. Each compiles a structured JSON request
into a single parameterised SQL statement executed directly against the backend,
bypassing Gramps object deserialisation entirely.

Three logically separate layers, delivered as one PR:

| Layer | Content | Commits | Lines |
|---|---|---|---|
| **(a)** Base compiler + endpoints over flat secondary columns | `query.py`, `object_query.py` | `e4a05f3` | ~1,800 |
| **(b)** `JsonPath` pushdown into `json_data`, then `RelatedObject` relationship traversal | `f7ff2df`…`dd5abfe` | 8 commits | ~600 |
| **(c)** `query_lang.py` — an "almost Python" expression parser exposed as `where_expr` | `0de440e`…`a5e0d7d` | 5 commits | 395 + 583 test |

Each layer is independently useful and independently shippable; (b) and (c) can be
removed without touching (a). This is the concrete form of the split requested earlier in
the thread.

## 2. Effect on existing behaviour

Essentially none, which is good: `api/__init__.py` is import reordering plus ten
`register_endpt` calls, and the net diff to `api/util.py` is a comment on `is_dbapi` plus
an `Optional[str]` type hint. No existing endpoint changes behaviour in the final state
of the branch.

## 3. Why a small first slice

This is not about the quality of the patch. It is that this *category* of code — SQL
issued directly against backend tables, below the layer where Gramps normally enforces
its invariants — has failure modes that none of the project's automated checks can see.

The `treeid` question earlier in this thread illustrates the shape. Whether `is_dbapi`
should cover SharedPostgreSQL is genuinely non-obvious: `SharedDBAPI` and `DBAPI` are
sibling classes rather than parent and child, so the natural reading of the type hierarchy
points the wrong way, and the only thing separating tenants is a column that nothing in
this repository knows exists. The resulting fix (`58541b7`, `765f419`) handles it properly
and in more places than would be expected.

The notable part is what did *not* flag it: not mypy, not the ~3,000 lines of tests in
this PR, not CI — there is no PostgreSQL job at all. Section 4 below is a second instance
of the same shape, and it was not caught during this review either. That is not a
statement about anyone's care; it is that this code currently has no backstop, so its
correctness rests on whoever reads it happening to hold the relevant backend internals in
mind at the time.

The recommendations below therefore have two goals: build the backstop, and keep the
surface that Web API itself must guarantee as small as possible.

## 4. Blocking: private sub-fields leak through `json_path`

For a user **without** `PERM_VIEW_PRIVATE`, the new endpoints return private sub-object
values that the existing proxy path strips.

`private = 0` replicates only the *row-level* half of `PrivateProxyDb`. The other half —
`sanitize_person` / `sanitize_event` / etc., which strip private **sub-objects**
(alternate names, attributes, event refs, note refs, addresses, URLs, media refs,
associations, LDS ordinances) — has no SQL equivalent and is not reimplemented.
`_render_json_path` in `gramps_webapi/api/query.py` applies no sub-field privacy logic,
and `resolve_column_path` accepts arbitrary paths with no allowlist.

### Reproduction

A **public** person carrying a **private** alternate name and a **private** attribute:

```python
proxy = ModifiedPrivateProxyDb(db)                 # what the API returns today
pp = list(proxy.iter_people())[0]
# proxy alternate_names : []
# proxy attributes      : []

q = Query(select=[JsonPath(("alternate_names", 0, "first_name"))])
sql, params = compile_query(PERSON, q, can_view_private=False, dialect=Dialect.SQLITE)
# query  alt name      : ['SECRET-ALT-NAME']
# query  attribute     : ['SECRET-ATTR-VALUE']
# SQL: SELECT json_extract(json_data, ?) FROM person WHERE private = 0 ...
```

The row-level `private = 0` is present and correct — it is simply the wrong control for
this class of data. And `where` is an oracle for the same values even without `select`:
a restricted caller can binary-search private data through filter predicates.

## 5. Multi-tenancy in the current state

After the fix, `treeid` is threaded through every raw query: `_resolve_treeid` in
`object_query.py`, `_where_clauses` for the main query and `COUNT`, `_resolve_after` for
the cursor lookup, and the `RelatedObject` subqueries in `query.py`. The only remaining
un-scoped `SELECT ... WHERE private = 0` is the pre-existing `_iter_handles` in
`util.py`, which is unreachable on SharedPostgreSQL for the sibling-class reason above.

Two things still worth changing:

1. `_resolve_treeid` returns `None` when the attribute is absent, and `None` means "omit
   the clause" — **fail-open**. A shared backend that names it differently silently gets
   unscoped queries. `_resolve_dialect` has the same shape, falling back to PostgreSQL
   for anything unrecognised.
2. All of this is asserted at the **SQL-string level only** (`assert "treeid" not in
   sql`). There are 44 `treeid` references in the tests and zero execution against a real
   shared backend. The highest-blast-radius path in the PR is never actually run.

## 6. Maintenance surface

The compiler necessarily hard-codes knowledge that Gramps core and the addons own: table
names, per-type secondary columns, the `private` column, `treeid`, collation setup,
dialect detection, reserved-word quoting. The PR's own comments document three live drift
hazards:

- gramps-project/gramps#2178 is unmerged, so `.dialect` has to be sniffed;
- `SharedPostgreSQL`'s `Connection.execute()` runs a blind `desc` → `desc_` string
  replace on *every* query, so column naming is correct only by a documented coincidence;
- `check_collation()` has different return contracts per backend.

Each is a place where an upstream change breaks or corrupts queries silently rather than
loudly. That is ~1,550 lines of compiler and parser whose correctness depends on
assumptions about two codebases this project does not control, and which track their
release cycles rather than Web API's. R4 and R5 are aimed squarely at this.

## 7. Interaction with granular permissions (#597)

Picking up the earlier exchange, where the answer was that if granular permissions are
added *"these endpoints will have to change"* — that is the crux.

Privacy is currently enforced in exactly one place: the proxy DB, in Python object space.
This adds a second enforcement path in SQL that has to stay in agreement. Any future
permission model — per-record, per-tag, role-based, living-person rules — then has to be
expressed twice, and anything not expressible as a SQL predicate cannot be expressed at
all. The 501 guard for non-`PrivateProxyDb` proxies is the right behaviour and also the
prognosis: when a fine-grained layer lands, these endpoints go dark for exactly the users
with restricted access, unless the rules are reimplemented in SQL.

Since #597's data representation is still undecided, merging a second consumer of it
first would be premature. That is an argument for keeping the initial surface minimal,
not for rejecting the feature.

---

## Recommendations

R1–R3 are blocking. R4 and R5 are the structural ones, since they change how much of this
Web API has to carry.

### R1. Behind a config flag, default off, marked experimental

Default-off endpoints with no API-stability promise keep exposure to deployments that opt
in — including the one driving this work, from day one. Given §3 and §4, the safe
assumption is that further issues of that class exist and have not been found yet; R1 is
what makes that assumption survivable.

### R2. Drop layer (b) from the first slice

This is where the §4 leak lives, and it is cleanly separable. Flat secondary columns are
per-row scalars, so row-level `private = 0` genuinely covers them. Arbitrary paths into
`json_data` mean reimplementing `sanitize_*` in SQL, which is not maintainable against
core drift.

**First slice = layer (a):** filter, sort and paginate over flat columns, which covers
most list-view needs. `JsonPath`/`RelatedObject` return as their own PR with an explicit
privacy design — most plausibly a curated allowlist of known-safe projections per type
rather than open traversal.

### R3. A proxy-conformance test

For a fixture corpus, assert that for every selectable column the query endpoint returns
**exactly** what the proxy path returns, for both a private-viewing and a restricted
user. Anything that cannot satisfy that invariant is not selectable.

This is the backstop from §3, and it would have caught §4 immediately. Whatever else
moves out of this repository (R4, R5), this test belongs here — it is how Web API
verifies its own guarantees regardless of who implements the machinery.

### R4. Give `query_lang` its own library

Folding this into gramps-ql is *not* the recommendation. The language here is more
expressive than gramps-ql, and rewriting it against a weaker grammar would be a real
loss.

The better outcome is for it to live where it can develop at its own pace: published as
its own package, maintained by its author, with Web API depending on it the way it
already depends on `gramps-ql`, `object-ql`, `sifts`, `yclade` and `gramps-gedcom7`. That
is an established pattern in this project, and this is a better fit for it than most — a
self-contained parser with no Web API-specific concepts in it, useful to anything that
queries Gramps data, including the application driving this PR.

Concretely: the author owns the grammar, the release cadence and the roadmap; Web API
takes a versioned, pinnable dependency and reviews an integration rather than a language.
That is plausibly a better arrangement for the author too, since changes to the language
would no longer be gated on maintainer review bandwidth here.

### R5. Consider the same for the SQL compiler — with one boundary

The same logic applies, more strongly, to `query.py`: it is the part that tracks Gramps
core and addon internals (§6), and it would benefit most from a release cycle that can
respond to a Gramps release without waiting for a Web API release. As its own library it
could also be tested against a matrix of Gramps versions and backends far more thoroughly
than is practical here.

One boundary is worth holding: **externalising the mechanism must not externalise the
policy.** A privacy or tenancy bug in a dependency compromises Web API users exactly as
much as one in this repository. So the library can own SQL generation, dialect handling
and path rendering, while Web API keeps and tests the parts that decide *what is allowed*
— which columns are selectable, the privacy predicate, the tree-scoping requirement —
with R3 and the R6 isolation test living here permanently. That interface is worth
working out jointly; it is the part most important to get right.

### R6. Fail closed, and PostgreSQL in CI

- Invert `_resolve_treeid` and `_resolve_dialect`: refuse unless the backend is *proven*
  single-tree or correctly scoped, instead of defaulting to "omit the clause" / "assume
  PostgreSQL".
- Validate at first use that the expected columns exist on the table (`PRAGMA table_info`
  / `information_schema`), so schema drift raises loudly rather than returning wrong data.
- Add a PostgreSQL/SharedPostgreSQL CI job with a **two-tree fixture** and a test
  asserting that a query in tree A returns nothing from tree B — or have the endpoints
  refuse to run on shared backends until that job exists. Per §3 and §5 this is the
  highest-value test in this list: it catches the next instance of this class without
  depending on anyone holding the addon's schema in mind.

### R7. Ownership of whatever remains here

For the parts that stay in this repository after R4/R5, `CODEOWNERS` entries would make
tracking the relevant upstream changes (#2178, the `desc_` hack, `check_collation`
contracts) assigned rather than assumed.

dsblank:
Great review, and it made some things more clear. I'm going to close this PR, and come back with one that does NOT attempt to implement any proxy. Then we can adjust roles + proxies in the futures.

PR #912, opened 2026-07-31
Make ModifiedPrivateProxyDb work with shared Postgres, too

DavidMStraub:
See https://github.com/gramps-project/gramps-web-api/pull/911 for the issue this would solve.

@dsblank here is a version that would not require scary raw SQL queries to make `ModifiedPrivateProxyDb` work with SharedPostgreSQL. It's around 8 times slower than the existing version on normal `DBAPI` backends, but still ~5 times faster than the "slow" version.

Honestly, I think a really fast and simple iterator over non-private database handles would be the addition of something like a 3-line method in Gramps core, this would be the much safer long term option.

What do you think?

### DIFF
--- a/gramps_webapi/api/util.py
+++ b/gramps_webapi/api/util.py
@@ -55,7 +55,6 @@ from gramps.gen.db.dbconst import (
     DBBACKEND,
     EVENT_KEY,
     FAMILY_KEY,
-    KEY_TO_NAME_MAP,
     MEDIA_KEY,
     NOTE_KEY,
     PERSON_KEY,
@@ -81,7 +80,6 @@ from gramps.gen.proxy.private import (
 from gramps.gen.proxy.proxybase import ProxyDbBase
 from gramps.gen.user import UserBase
 from gramps.gen.utils.grampslocale import GrampsLocale
-from gramps.plugins.db.dbapi.dbapi import DBAPI
 from marshmallow import RAISE
 from webargs.flaskparser import FlaskParser
 from werkzeug.exceptions import HTTPException
@@ -133,7 +131,6 @@ class ModifiedPrivateProxyDb(PrivateProxyDb):
         """Initialize self."""
         super().__init__(*args, **kwargs)
         self.name_formats = self.db.name_formats
-        self.is_dbapi = isinstance(self.basedb, DBAPI)
 
     def get_dbname(self):
         """Get the name of the database."""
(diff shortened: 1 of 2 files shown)


dsblank:
This doesn't respect treeid, right? I think the original analysis was wrong.

DavidMStraub:
This one here does respect treeid because `_iter_raw_data` is implemented in `SharedDBAPI`:

https://github.com/gramps-project/addons-source/blob/maintenance/gramps61/SharedPostgreSQL/shareddbapi.py#L1042-L1057

dsblank (approved):
## Review: Make `ModifiedPrivateProxyDb` work with shared Postgres, too

### Overview
This PR replaces the raw-SQL, backend-specific `_iter_handles()` helper in `gramps_webapi/api/util.py` with a generic implementation built on Gramps' `_iter_raw_data(obj_key)`, and drops the `is_dbapi`/`DBAPI`-instance branching that every `iter_*_handles()` method used to need. It's a follow-up to #911, where a narrowing of `isinstance(self.basedb, DBAPI)` was needed specifically to keep the SharedPostgreSQL addon (`SharedDBAPI`, a sibling class of `DBAPI`, not a subclass) off the old raw-SQL fast path — because that raw SQL had no `treeid` filter and SharedPostgreSQL stores every tree's rows in the same physical tables.

### Privacy/leak analysis (the main ask)
I traced this carefully since `iter_person_handles()` and friends feed Gramps' native `Rule`/`Filter.apply()` machinery — the fallback path used by `object_query` when a db is proxied. A handle-list leak here could surface through search/filter results even though per-object getters re-check privacy.

- **Correct approach, and actually fixes the root cause**: `_iter_raw_data()` is polymorphic — `DBAPI._iter_raw_data` has no treeid concept, but `SharedDBAPI` overrides it with `WHERE treeid = ?` scoping (`shareddbapi.py:1042-1052`). By dispatching through this instead of hand-rolled SQL, the new `_iter_handles()` gets correct tree-scoping "for free" on the shared-Postgres backend, which is exactly the bug #911 could only work around, not fix.
- **Privacy semantics preserved**: `data.private` off the raw deserialized struct corresponds directly to `obj.get_privacy()` — confirmed against `gramps/gen/proxy/private.py`'s `include_*` predicates, which likewise just check the object's own `.private`/`get_privacy()`. No behavior change vs. upstream `PrivateProxyDb.include_*`.
- **No new leak surface**: the `is_dbapi` attribute and its narrow-isinstance guard are fully removed and had no other callers in the codebase (verified via grep) — nothing outside this method depended on it.
- **Gap**: there's no test exercising `SharedDBAPI`/actual SharedPostgreSQL — understandably hard in CI without a Postgres addon install — but that's precisely the scenario motivating this PR. The included tests only run against a SQLite-backed `DBAPI`, so the fix's core claim ("works with shared Postgres too") is unverified by CI, only by code-reading/inspection. Worth at least a comment noting this is manually verified, or a follow-up issue for a SharedPostgreSQL-backed test.

I did not find any private-data leak introduced by this change — if anything, it closes a latent one (SharedPostgreSQL previously depended on the slower `filter(include_*, ...)` fallback continuing to be selected correctly by the sibling-class check, which was fragile/easy to accidentally widen, as the removed code's own comments warned).

### Code quality
- Clean deletion of the now-dead `KEY_TO_NAME_MAP`/`DBAPI` imports and the `self.is_dbapi` field — confirmed no dangling references remain.
- The new docstring/comment on `_iter_handles` explaining *why* `_iter_raw_data` is used (backend-agnostic, no schema/tree-scoping assumptions) is good.
- Every `iter_*_handles` method collapses to a single-line call, removing ~40 lines of repetitive branching — a nice simplification.

### Performance
- Author-disclosed: ~8x slower than the old raw SQL on plain `DBAPI` (since every row is now deserialized to check `.private`, rather than filtering in SQL), but ~5x faster than the previous fallback (`filter(include_*, db.iter_*_handles())`) that non-`DBAPI`/`SharedDBAPI` backends were using. Reasonable tradeoff for correctness across backends; worth keeping an eye on for very large trees, but not a blocker.

### Test coverage
- New parametrized test (`test_iter_handles_agrees_with_include_predicate`) is a good generic equivalence check across all 9 object types against Gramps' own `include_*` predicates.
- Two additional targeted tests for private note/event exclusion add clarity beyond the generic check.
- As noted above, all new tests run against SQLite (`DBAPI`) only — no coverage of the `SharedDBAPI` path this PR specifically targets.

### Recommendation
Sound, well-scoped fix that generalizes the privacy-safe handle iteration correctly rather than patching around backend quirks. Approving — would be great to confirm manual verification against an actual SharedPostgreSQL instance, given the gap in automated coverage for that path.

DavidMStraub:
Thanks! I'll merge this then.

PR #923, opened 2026-08-04
Pass Postgres credentials when creating new trees

dsblank:
## Summary
- `TreesResource.post()` creates a new tree's database via `WebDbManager` but never passed `username`/`password`, so tree *creation* always relied on passwordless/default Postgres auth — even though every other `WebDbManager` call site (see `api/util.py`) already passes `POSTGRES_USER`/`POSTGRES_PASSWORD` from app config.
- This opens up using Postgres via a non-standard auth path (i.e. a Postgres instance that actually requires a username/password) for multi-tree deployments.
- `POSTGRES_USER`/`POSTGRES_PASSWORD` already have `None` defaults in `config.py`, and `WebDbManager` already treats them as optional, so this is a no-op for existing sqlite/default-Postgres setups.

## Test plan
- [x] Existing test suite (sqlite backend) still passes
- [x] Manually verified: create a tree against a Postgres backend configured with non-default `POSTGRES_USER`/`POSTGRES_PASSWORD`

### DIFF
--- a/gramps_webapi/api/resources/trees.py
+++ b/gramps_webapi/api/resources/trees.py
@@ -166,6 +166,8 @@ class TreesResource(ProtectedResource):
         WebDbManager(
             dirname=tree_id,
             name=args["name"],
+            username=current_app.config["POSTGRES_USER"],
+            password=current_app.config["POSTGRES_PASSWORD"],
             create_if_missing=True,
             create_backend=backend,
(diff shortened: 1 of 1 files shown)


DavidMStraub:
Oh wow, huge regression in my #909 :grimacing:  Thanks for the fix...

PR #927, opened 2026-08-09
Reduce server load when polling the transaction history

DavidMStraub:
@dsblank since you mentioned you rely on polling, I invested some work to make polling of the history endpoint less expensive for the server.

---

**Claude:**

`GET /api/transactions/history/` had no ETag, no `Cache-Control` and no caching,
so every poll did full work. It also loaded every change of a connection to
serve the few transactions on one page.

## Changes

**Change loading** (`undodb.py`). `Transaction._to_dict()` no longer walks
`connection.changes`. `_get_changes()` fetches the changes of a page in one
query, built from each transaction's `(connection_id, first..last)` range, and
partitions the rows in Python. The legacy binary columns are always deferred,
`old_json`/`new_json` only when not requested. The connection is loaded with
`contains_eager` instead of lazily per row. Since `page` defaults to `0`
(unpaginated), transactions are chunked at 500 per query.

**ETag and 304** (`history.py`). `get_transactions_state()` returns
`(max_id, count)` from one aggregate query. The validator is
`sha256(tree_id, max_id, count, args)` and is checked against `If-None-Match`
before any change-log work, so a steady-state poll costs one aggregate query.
Responses carry `ETag`, `Cache-Control: no-cache` and `X-Total-Count`;
304s carry the headers without a body.

**User names** (`cache.py`). `get_user_dict()` moved out of `history.py` into
`request_cache` with a 60 s timeout, and is now called only after the 304
check. Users live in the auth database, so there is no change timestamp to key
on; a cached mapping missing one of the IDs being rendered is refetched.

`return_304_if_unchanged()` in `resources/util.py` was refactored to share the
new `etag_unchanged()` predicate.

## Notes

- The validator does not cover user names: renaming a user becomes visible once
  the next transaction is written.
- `get_transaction()` now returns `None` instead of raising `AttributeError`
  for a missing ID. Two `except AttributeError` blocks became dead and were
  removed; `TransactionHistoryResource` gained the 404 it was missing (it
  previously returned 500).
- Unchanged and still worth a look: a transaction with `first is None` (an
  empty `DbTxn`) reports every change of its connection.

## Tests

43 passing in `test_history.py`, `test_undodb.py` and `test_transactions.py`,
7 of them new: ETag round trip with invalidation on a new transaction, the
validator depending on query args, change scoping across transactions sharing a
connection, the chunk boundary, `get_transactions_state`, `get_transaction`
including the missing-ID case, and `old`/`new` gating.

### DIFF
--- a/gramps_webapi/api/cache.py
+++ b/gramps_webapi/api/cache.py
@@ -147,7 +147,7 @@ def get_user_dict(user_ids: Collection[str] = ()) -> dict[str, dict]:
     include_treeless = current_app.config["TREE"] != TREE_MULTI
     cache_key = f"user_dict:{tree}:{int(include_treeless)}"
     user_dict = request_cache.get(cache_key)
-    if user_dict is None or not set(user_ids) <= set(user_dict):
+    if user_dict is None or any(user_id not in user_dict for user_id in user_ids):
         user_dict = _fetch_user_dict(tree, include_treeless)
         request_cache.set(cache_key, user_dict, timeout=USER_DICT_CACHE_TIMEOUT)
     return user_dict
--- a/gramps_webapi/api/cache.py
+++ b/gramps_webapi/api/cache.py
@@ -6,7 +6,9 @@ import hashlib
 import json
 import os
 
-from flask import g, request
+from typing import Collection
+
+from flask import current_app, g, request
 from flask_caching import Cache
 from gramps.gen.errors import HandleError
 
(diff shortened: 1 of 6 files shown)


DavidMStraub:
## Polling the transaction history: instruction for frontends :robot: 

```
GET /api/transactions/history/?page=1&pagesize=20&sort=-id&after_id=<last seen>
If-None-Match: <etag from last 200>
```

- Always send `If-None-Match` — unchanged history returns `304`, no body.
- Always set `page=1` — it defaults to `0`, meaning the entire history.
- Use `after_id`, not `after` — the timestamp cursor redelivers rows.
- Never send `old=1`/`new=1` — full object JSON per change.
- Keep the URL byte-identical — a changed query string cannot `304`.
- `X-Total-Count` is on the `304` too.
- Back off when idle, reset on a `200`.

PR #936, opened 2026-08-15
Negotiate thumbnail format instead of hardcoding AVIF

dsblank:
## Summary
- `send_thumbnail`/`send_thumbnail_cropped`/`send_cropped` always encoded thumbnails as AVIF with no way for a client to ask for anything else
- AVIF decoding isn't universal: confirmed live against Ubuntu's stock WebKitGTK (2.52.3, no libavif/dav1d linked) that a thumbnail request 200s with valid AVIF bytes and then silently fails to decode in an `` — no console error, no broken-image fallback. Every thumbnail in a client built on that engine (gramps-connect's standalone Linux build's native window) was therefore just blank
- `negotiate_thumbnail_format()` (`api/image.py`) now returns JPEG unless the client's `Accept` header explicitly lists `image/avif` as an exact entry — deliberately not `accept_mimetypes.best_match()`, since that treats a bare `*/*` (curl's default, and the trailing catch-all on most real `Accept` headers) as a match for every candidate and can't distinguish "this client said it decodes AVIF" from "this client accepts anything" (confirmed it returned AVIF even with no avif/webp/image/* entry at all)
- The thumbnail cache key (`api/cache.py`) now includes the negotiated MIME type, since it previously hardcoded `:avif` — otherwise an AVIF-capable and AVIF-incapable client would collide on the same cached response

## Test plan
- [x] Confirmed fixed by loading the real thumbnail URL in a headless WebKitGTK view and checking `img.naturalWidth`: 0 before, 100 after
- [ ] CI

### DIFF
--- a/gramps_webapi/api/cache.py
+++ b/gramps_webapi/api/cache.py
@@ -13,6 +13,7 @@ from flask_caching import Cache
 from gramps.gen.errors import HandleError
 
 from gramps_webapi.api.auth import has_permissions
+from gramps_webapi.api.image import negotiate_thumbnail_format
 from gramps_webapi.api.util import (
     get_db_handle,
     get_db_manager,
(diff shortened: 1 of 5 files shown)


DavidMStraub:
Please open a feature request for this first. Thanks!

dsblank:
> Please open a feature request for this first. Thanks!

It is really a bug, and I already had the code so that I could get a system to work, and didn't want to lose it.

DavidMStraub:
In my opinion it's not a bug, but let's please discuss it without me having to review a PR first.

I don't want to be rude, but CONTRIBUTING.md states that non-trivial changes require an issue first, because reviewing a PR takes more time, and when I'm short on time I think I have the right to insist on it.

PR #939, opened 2026-08-17
Add concise Roadmap

DavidMStraub:
This would make it easier for API consumers to plan ahead in case of breaking changes.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,0 +1,32 @@
+# Roadmap
+
+Rough timelines and planned breaking changes, so projects building on the API can
+plan ahead. Not a commitment, not an issue tracker.
+
+## Versioning
+
+Semver applies to the HTTP API, config options and env vars, not to Python internals.
+Only majors break: minors add, patches fix. Deprecations are announced in a minor and
+removed no earlier than the next major.
+
+## Releases
+
+| Version | Target | Gramps | Focus |
+|---|---|---|---|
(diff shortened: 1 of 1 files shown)


dsblank:
Some questions:

1. Does 3.21 include the new new fast query endpoints? I hope so
2. When switching to multi-tree only, does that mean gramps-web-api only supports SharedPostgresql? Can we add a SharedSQLite as well for easy testing and use?

dsblank:
Personally, I'd like to work on making importXML faster, but I think a lot can be done in gramps core. I'm also interested in the multi-tree per user feature as well.

DavidMStraub:
This was not meant to replace the issue tracker and not meant as a feature roadmap - just to give a rough indication of

- upcoming breaking changes
- release timing
- Gramps version ties

I tried to clarify the wording now, is it better?

dsblank:
Yes, I understand your motivations for the document better.

Although, it still leaves the question:

> When switching to multi-tree only, does that mean gramps-web-api only supports SharedPostgresql? Can we add a SharedSQLite as well for easy testing and use?

If the breaking changes prevent SQLite use, then I do not like the proposed plan.

DavidMStraub:
No, multi-tree is completely independent from database backends. You can start Web API with your 20 existing SQLite Gramps dbs and it will work just fine. A tree is identified by its directory name under GRAMPSHOME. That's no different from Gramps desktop.

dsblank (approved):

PR #949, opened 2026-08-23
Add formatted address dates to the person profile

### LINKED ISSUE #1353: [Bug] Addresses show Date 0-0-0
### Bug Description

An Address on a Person without a date shows a date row with the value 0-0-0.

### Gramps Web Version Information

Gramps 6.0.8
Gramps Web API 3.20.1
Gramps Web Frontend 26.8.0
Gramps QL 0.4.0
Sifts 1.3.1
locale: en
multi-tree: false
task queue: true
OCR: true
chat: false

### Verification

- [x] I have searched for existing issues before creating this one

karlvr:
Used by gramps-project/gramps-web#1352 and fixes gramps-project/gramps-web#1353.

`Address` has a date (from `DateBase`), but no profile exposes it as a formatted string. Clients that want to display it have to format the raw `dateval` themselves, which loses modifiers (`about`, `between`), non-Gregorian calendars, and the user's locale and date-format preference.

This adds `addresses` to the person profile using `date_displayer`, paralleling `address_list`:

```json
"profile": {
  "addresses": [{"date_str": "2000-01-02"}]
}
```

This follows the existing `alternate_place_names` / `direct_parent_places` precedent in the place profile.

I'm working on a PR for the frontend to take advantage of this change.

### Repositories

`address_list` is on **Person** and **Repository** but repositories don't have a profile. If you're open to it, I'm happy to add a profile for repository so this same change can apply there. That would produce symmetry in the frontend too.

### Testing

Added a test covering the one person in the example database with an address (I0044). Updated the existing exact-match profile fixtures in `test_people.py`, `test_families.py` and `test_events.py` to include the new key.

### DIFF
--- a/gramps_webapi/api/resources/repositories.py
+++ b/gramps_webapi/api/resources/repositories.py
@@ -20,11 +20,18 @@
 
 """Repository API resource."""
 
+from typing import Dict
+
+from gramps.gen.const import GRAMPS_LOCALE as glocale
+from gramps.gen.lib import Repository
+from gramps.gen.utils.grampslocale import GrampsLocale
+
 from .base import (
     GrampsObjectProtectedResource,
     GrampsObjectResourceHelper,
     GrampsObjectsProtectedResource,
 )
+from .util import get_repository_profile_for_object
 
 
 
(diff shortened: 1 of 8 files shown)


DavidMStraub:
Thanks!

> address_list is on Person and Repository but repositories don't have a profile. If you're open to it, I'm happy to add a profile for repository so this same change can apply there. That would produce symmetry in the frontend too.

Yes, I think it makes sense!

karlvr:
@DavidMStraub I've added profile support to Repository following the pattern in Person pretty closely. I called `super().object_extend` at the end of `object_extend` to avoid duplicating the logic from `base.py`. That's different to the other resources so let me know if you'd prefer it to match.

I've also added the schema improvements suggested by copilot above; they were indeed missing.

DavidMStraub:
Thanks! Indeed the old resource classes are not very DRY...

PR #961, opened 2026-08-28
Drop unique constraint on user email

DavidMStraub:
... because it's annoying on multi-tree deployments and doesn't buy anything.


DavidMStraub:
Ready

PR #966, opened 2026-09-02
Replace RelationshipCalculator with a fast SQL-derived ancestry graph

dsblank:
## Summary

Gramps' own `RelationshipCalculator` finds the relationship between two
people by recursively walking `Person`/`Family` objects and enumerating
every distinct path to a common ancestor, not just the shortest one.
Under pedigree collapse (shared distant ancestors), a normal
genealogical pattern, that becomes exponential: a pair connected through
a handful of real ancestors can take minutes and pin a CPU core,
regardless of tree size. On a ~100k-person tree this reproduced a
multi-minute hang and a real kernel OOM-kill.

This replaces the *search* (not the wording) with a SQL-derived
parent/child edge set plus a plain breadth-first search, each node
visited once, so cost tracks distinct people rather than distinct paths.
Locale-aware wording, privacy filtering (mirroring `PrivateProxyDb`'s
rules), and the pedigree-collapse tie-break are all preserved.

The implementation itself now lives in a standalone package,
[gramps-sql-extensions](https://github.com/dsblank/gramps-sql-extensions)
(published on PyPI), decoupled from gramps-web-api via a single
`execute(sql, params) -> rows` callable. `gramps_webapi/api/resources/relationship_graph.py`
is the thin adapter that wires a `db_handle` into it.

- `RelationResource` and `RelationsResource` both now use the fast graph
- Privacy filtering done as live SQL predicates, matching
  `PrivateProxyDb`'s three rules (private person / private family /
  private `ChildRef`)
- `gramps-sql-extensions>=0.1.0` added to `pyproject.toml` dependencies

## Test plan

- [x] `tests/test_endpoints/test_relations.py` — 17/17 pass, both against
      the editable local checkout and the published PyPI release
- [x] Verified against live SQLite and live Postgres/SharedPostgreSQL
- [x] `gramps-sql-extensions` has its own independent test suite (19
      tests: relationship lookup, privacy filtering, edge cases like
      pedigree-collapse tie-break and remarriage ordering) with CI on
      every push/PR

## Known gaps

- No trigger-maintained permanent ancestry table — this builds an
  ephemeral per-request temp table instead (deliberate scope decision,
  a permanent table would need a gramps-core schema migration)


DavidMStraub:
Thanks for this! I think it would be great to close this bottleneck.

As discussed, I now did a first review pass of this PR with the main goal to assess whether it's realistic to merge it this week (in time for 3.22). The answer is no - I have architectural concerns and am not comfortable with merging it as is.

In the light of [this discussion](https://gramps.discourse.group/t/proposal-process-for-upstreaming-web-api-improvements-into-gramps-core/9940), I think this is a textbook example of an improvement that should live in Gramps core: it's about the algorithmic performance of a core genealogical function (relationship calculation). Adding another library depdency is not the problem per se; the problem is that the library code reaches deep into the internals of Gramps core, and even depends on internals of SharedDBAPI which lives (as of today) in addons-source. So, in effect, we get a coupling between 3 different, independently versioned projects with no clear API contract.

And the library does scary things: it takes a `readonly=True` which is doc-string documented as "write operations will fail … this is not enforced by Gramps but must be taken care of in Web API methods". The library issues `DROP TABLE` and `CREATE INDEX` on those!

I propose the following way forward:

1. Agree on an architecture for this that lives in Gramps core
2. Assess how this can be patched into `gramps_webapi.future_gramps`
3. Merge that version into Gramps Web API and release as soon as it's ready

I will have time to contribute to step 1 next week (either proposing something or reviewing a proposal).

In the meantime, below I'm also posting some correctness issues that Claude Opus identified in its review. FYI, Claude also recommends to split this into two steps, and start by only implementing the following as step 1:

> Inside RelationshipCalculator.__apply_filter, replace the recursive walk that enumerates every path with one that goes generation by generation and visits each person once. Keep the outputs exactly as they are: the same path strings, the same family-index lists, the same stop-at-the-nearest-common-ancestor rule.

I haven't had time to check whether that proposal makes sense.

---

Correctness findings (architecture-independent) :robot: 

Verified against published `gramps-sql-extensions==0.2.0` on the standard `example.gramps`, diffed against `RelationshipCalculator`.

**1. Wrong relationship terms in non-English locales — 19/400 random pairs (~5%) in German**

Core feeds `collapse_relations`' **family-level** paths (`'ma'`, `'fa'` = `REL_FAM_BIRTH`) to `get_single_relationship_string`; the PR feeds raw **person-level** paths (`'mf'`, `'ff'`). `rel_de.py:236-254` derives the common-ancestor count from the last path char, so full relations become half relations:

```
core: ('Tante vierten Grades', 5, 4)     PR: ('Halbtante vierten Grades', 5, 4)
core: ('Obergroßneffe', 1, 9)            PR: ('Halbobergroßneffe', 1, 9)
core: ('Oberneffe zweiten Grades', 2, 9) PR: ('Halboberneffe zweiten Grades', 2, 9)
```

Same in `rel_de_AT`. 23 locale calculators read `reltocommon_a` (`rel_ru` 27 uses, `rel_pl` 20, `rel_sv` 12, `rel_uk`/`rel_tr` 9). English is 400/400 identical — which is why the existing tests pass. The tie-break comment ("English wording depends only on `(Ga, Gb, gender, only_birth)`") is true *for English* and is precisely the bug: it's stated as a general justification.

**2. `/relations/.../all` over-reports — 20/300 pairs**

The docstring claims the gap is *under*-reporting; it is the opposite. Core stops recursion at the nearest common ancestor (`relationship.py:1592`, `if commonancestor and not self.__crosslinks: return`). The PR walks all 15 generations and emits every common ancestor found:

```
core: 1 entry  ->  PR: 12 entries
core: [{'granduncle', [2 ancestors]}]
PR:   granduncle + 1st..7th cousin twice removed, ~35 ancestor handles
```

Also emits a spouse entry (`common_ancestors: []`) that core's `get_all_relationships` never returned.

**3. Non-deterministic response ordering**

`sorted(common, key=lambda h: dist1[h] + dist2[h])` sorts a `set` on a non-unique key, so `common_ancestors` order follows `PYTHONHASHSEED`:

```
seed 0: ['LV4KQCMDZ6MWABTXHN', 'LW4KQCI6NR543UDCSL']
seed 1: ['LW4KQCI6NR543UDCSL', 'LV4KQCMDZ6MWABTXHN']
```

Breaks response stability across workers and anything ETag- or cache-keyed. Needs a total sort key.

**4. Wasted work on the spouse path**

`relationship()` calls `ensure_child_of()` — the full whole-tree edge-table build (~1.5s on 100k people, per the commit message) — *before* `check_spouse`, so a spouse lookup pays the entire cost and then discards it.

### What holds up

- English single-relationship: **400/400** identical at depth 15.
- Depth semantics: **250/250** identical at depths 2, 3, 5, 8 — the `depth < max_depth - 1` boundary reasoning is right.
- Privacy filtering: **150/150** identical against `PrivateProxyDb` with `restricted=True`.

dsblank:
Thank you! The plans sound good. I will explore the correctness issues.

dsblank:
> Inside RelationshipCalculator.__apply_filter, replace the recursive walk that enumerates every path with one that goes generation by generation and visits each person once. Keep the outputs exactly as they are: the same path strings, the same family-index lists, the same stop-at-the-nearest-common-ancestor rule.

Claude says:

**The idea**

Right now, relationship() (and relationship_path()) computes ancestor_map() for both people independently, each walking out to max_depth (15 generations by default) regardless of where — or whether — a common ancestor actually exists, then intersects the two results afterward. The suggestion (as understood by my Claude), applied here, is: interleave the two searches so the moment one side's frontier crosses into the other side's already-discovered ancestors, stop expanding further — avoiding walking generations that turn out to be irrelevant.

**Pros**

- Real, if narrow, savings. For a pair that's either unrelated or only distantly related, both sides currently get walked to the full 15-generation limit for nothing. Stopping the moment a common ancestor appears would cut that wasted exploration.
- Matches the already-adopted stop-at-nearest-common-ancestor principle. _nearest_common_ancestors() already implements this rule for correctness (which results get reported); this would extend the same idea to the search itself, not just the filtering step.

**Cons**

- The dominant costs aren't where this helps. Per the module's own measured numbers, the expensive part of a call is building/querying the child_of edge table (~100ms–1.5s depending on backend/tree size) — a fixed cost paid once regardless of search depth. The Python-side BFS over already-fetched edges is cheap in-memory dict work. So the "wasted" generations this would trim are the cheap part of the operation; the payoff is likely small relative to total call cost.
- It breaks a real architectural strength. Each person's ancestor search is currently one clean, independent, portable SQL query (works the same on SQLite and Postgres). Making the two searches aware of each other's progress means either: (a) restructuring into multiple SQL round-trips, one per generation, checking for overlap after each — which trades "search too far" for "many small round trips," likely worse in Postgres where network latency dominates; or (b) fetching everything anyway and just stopping the Python loop early, which barely saves anything since the SQL query already returned the full edge set in one shot.
- Only applies to the single-answer paths. all_relationships()/all_relationship_paths() genuinely need every common ancestor within max_depth, so they can't stop early without missing results — this would only benefit relationship()/relationship_path()/relationships_to()'s per-target lookups.

**Bottom line**: the principle is sound but the expected win looks small against the already-dominant fixed cost of the edge-table query, and the way to get it (interleaved/round-tripped search) trades against the current design's simplicity and portability. I'd want to see it justified with a profile on a real large/deep tree showing the BFS-to-max-depth step actually costing something non-trivial before taking on that complexity — not something I'd build speculatively.

dsblank:
Refactoring to remove temp table and DROP table writing on a read-only table, via a DDL-free version. And its faster.

dsblank:
Removed the table creation/deletion code. Less scary!

pyproject.toml
```diff
@@ -16,6 +16,7 @@ classifiers = [
 requires-python = ">=3.10"
 dependencies = [
     "gramps[all]>=6.0.4,<6.1.0",
+    "gramps-sql-extensions>=0.1.0",
```
dsblank:
```suggestion
    "gramps-sql-extensions>=0.2.2",
```

DavidMStraub:
Some more Claude discussion lead to this assessment. What do you think?

---

## Reassessment, taking David's review comments and requirements into account

David asked me to re-read this thread and propose a way forward that fits the requirements in his review. Short version: the algorithm in this PR is the right one, it should land, and I think the fastest route to landing it is as a change to `RelationshipCalculator` itself, staged through `gramps_webapi.future_gramps` — not as a separate SQL layer alongside it.

### What 0.2.x already resolved

Three of the four review points are closed, in the library rather than in this PR:

- **DDL against a `readonly=True` handle** — gone in 0.2.2. `ensure_child_of()` is a plain `SELECT` now. The investigation also turned up a real bug on the way (the temp table was never `ANALYZE`d, so Postgres seq-scanned it: 291 ms vs 0.26 ms).
- **Correctness findings 1–4** — all four fixed in one commit: family-level path collapse for the locale calculators, `_nearest_common_ancestors()` for the `/all` over-reporting, a total sort key for determinism, spouse check before the edge build.
- **`[{}]` on no relationship** is preserved, so no response-shape change.

Two deltas versus core are now documented rather than fixed, and both look like deliberate calls worth confirming explicitly: `all_relationships()` reports only the shortest path per ancestor (a longer, differently-worded second path to the *same* ancestor under pedigree collapse is under-reported), and it emits a spouse entry with `common_ancestors: []` that `get_all_relationships()` never returned.

### What the architectural objection was, and still is

The remaining concern isn't the DDL — it's that the fast path re-derives things gramps-core owns, in a repo that versions independently of both core and the shared Postgres addon:

- **Storage layout**: `person`/`family` tables, the `json_data` column, the `father_handle` / `mother_handle` / `private` / `gender` secondary columns, the JSON paths `$.child_ref_list`, `$.frel.value`, `$.type.value`, per-dialect JSON functions, and `treeid` reached via `db_handle.db.dbapi`.
- **Core logic**: `collapse_relations` / `_famrel_from_persrel`, sibling typing, spouse typing — and `PrivateProxyDb`'s three privacy rules. Wording drift there is cosmetic; privacy drift is a disclosure bug. The German half-relation bug was exactly this failure mode, and it was caught by chance rather than by a test.

### The observation that I think changes the picture

The expensive thing in core is that `__apply_filter` re-expands a person's entire ancestry once per distinct path that reaches them (`relationship.py:1561-1595` — the `person.handle in pmap` branch appends and then recurses again). Under pedigree collapse that's exponential. The visited-once BFS in this PR fixes precisely that.

But that BFS doesn't need SQL. Run the same search over `db.get_person_from_handle()` / `get_family_from_handle()` and you get:

- every backend, including BSDDB and in-memory, with no dialect table;
- privacy from the existing proxy — nothing to reimplement, nothing to drift;
- and, I'd expect, *less* work than 0.2.2 rather than more. Since removing the temp table, `ensure_child_of()` loads the whole tree's parent/child edges plus every private person/family handle on every call, deliberately uncached (the library only sees an opaque `execute`, so it can't know the connection is exclusive). Per the README's own numbers that's ~600 ms on SQLite and ~1.5 s on the 100k Postgres tree, per request, per worker, for a two-person lookup. An object-API BFS touches only the two people's ancestors — thousands of objects at depth 15, not the whole tree. It would also let the endpoint drop today's `cache_people()` / `cache_families()` full preload.

I want to be careful not to assert that as a measured result: it's a prediction, and the harness in this PR is the right thing to test it with. The one case where the whole-tree edge load clearly wins is bulk (`relationships_to()` over everyone), which no current endpoint uses.

One clarification on the earlier exchange: the suggestion Doug evaluated was *interleaving the two searches for an early stop*, and the response to that is correct — the win is small and it costs portability. The win that matters is deduplicating the expansion, which this PR already implements.

### Proposed way forward

1. **A gramps-core PR**: replace `__apply_filter`'s path-enumerating recursion with the visited-once BFS and best-path tie-break from this work, over the `DbReadBase` object API. `collapse_relations` and all of `rel_*.py` stay untouched. The shortest-path-per-ancestor change is a real (and I think correct) behaviour change to `get_all_relationships()`, so it should be decided upstream rather than shipped quietly. The 400-pair English + German diff against `example.gramps` that found the locale bug is the natural parity test.
2. **Debut `gramps_webapi.future_gramps` here**: a module carrying that core patch *verbatim*, applied as a monkeypatch at app init, with the upstream PR link, a `gramps.version` guard that disables it once the fix ships in core, and a test asserting parity with the unpatched calculator. Staying byte-identical to the submitted core patch, and expiring, is what keeps this from becoming a permanent fork. The `depth=5` retry hack in `util.py:1962-1975` exists only because of this bug and can go at the same time.
3. **Reshape this PR** around that: endpoints keep their current shape, and `RelationsResource`'s over-reporting gets fixed by the core change rather than by a second implementation running beside it.
4. **Re-measure afterwards.** If a 100k tree is still too slow once the exponential blowup is gone, the edge-index acceleration is worth revisiting — with the index exposed as a DBAPI-level method in core, where the schema knowledge (including `treeid`) legitimately lives; privacy taken from the proxy; and the index cached per tree against `get_db_last_change_timestamp()` instead of rebuilt per request.

That sequence keeps the performance fix on a path to being available to every Gramps user rather than only to Web API, and it gets the endpoints unblocked in the meantime without a parallel implementation to keep in sync.

dsblank:
Thanks! I'll look at the proposed algorithm in detail over the next few days, and do some performance testing.

But I do have ideas about where the code should live: we should put the code in gramps/addons. That way it can be used in other places (gramps, gramps/addons) and gramps-web-api. If gramps maintainers decide to move the code into core, then we can methodically change the imports, and some day mark the addon as deprecated.

dsblank:
I think I already wrote some of what Claude suggests: https://github.com/gramps-project/gramps/pull/2111

I'll explore the options.

DavidMStraub:
I suspect at least part of it should be an algorithm bug fix in core...

Nick-Hall:
Yes.  I also suspect that there may be a bug in Gramps core.

dsblank:
Yes, there were 3 issues found in core:

https://github.com/gramps-project/gramps/pull/2526

dsblank:
There is also a hack that we should remove in `gramps-web-api` when we use a fixed algorithm (either from an addon or from gramps-core):

https://github.com/gramps-project/gramps-web-api/blob/master/gramps_webapi/api/resources/util.py#L1895-L1917

PR #970, opened 2026-09-05
Add transaction_id to object change history

DavidMStraub:
The object history endpoint returned a connection id but no transaction id, so a change row could not be linked to its revision diff at /transactions/history/<id>. Connection.id and Transaction.id are not interchangeable: one connection holds many transactions, each covering a Change.id range via first/last, so the frontend had no way to deep-link a row and rendered the rows as non-clickable.

Resolve the covering transaction for each returned change. The id ranges are matched in SQL and the per-change terms chunked at the existing CHANGES_QUERY_CHUNK_SIZE, so the rows loaded scale with the page rather than with the number of transactions in the connections it touches. A transaction without a range covers its whole connection, as in _get_changes_chunk. The field is nullable when no covering transaction is found.

---

@dsblank this is needed so Gramps Web can link back from the individual change to the transaction view. It costs a little, but I tried to optimize the SQL query so it's acceptable. Adding a query parameter to switching this on/off would be overkill IMO. Agreed?

### DIFF
--- a/gramps_webapi/api/resources/schemas.py
+++ b/gramps_webapi/api/resources/schemas.py
@@ -1902,7 +1902,7 @@ class ObjectChangeSchema(_Base):
     transaction_id = fields.Int(
         allow_none=True,
         metadata={
-            "description": "ID of the transaction (see /transactions/history/<id>) covering this change, if any."
+            "description": "ID of the transaction this change was part of, if known."
         },
     )
     obj_class = fields.Str(
--- a/gramps_webapi/api/resources/schemas.py
+++ b/gramps_webapi/api/resources/schemas.py
@@ -1899,6 +1899,12 @@ class ObjectChangeSchema(_Base):
     connection = fields.Raw(
         metadata={"description": "Internal connection object."},
     )
+    transaction_id = fields.Int(
+        allow_none=True,
+        metadata={
+            "description": "ID of the transaction (see /transactions/history/<id>) covering this change, if any."
+        },
+    )
     obj_class = fields.Str(
         metadata={"description": "Object class name (e.g. 'Person', 'Event')."},
     )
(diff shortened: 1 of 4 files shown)


dsblank (approved):
Agreed on skipping the toggle.

The correlated subquery approach makes sense given the schema constraints here: the Transaction row doesn't exist yet when its Change rows are written (_after_commit creates it after the fact, from the already-known id range), so a direct FK set at write time isn't naturally available without restructuring the write path. Resolving the covering transaction at read time is consistent with how _get_changes_chunk already does the reverse lookup, and the tie-break (ranged transaction wins over whole-connection, lowest id wins) looks like it correctly keeps an undo/redo transaction from stealing credit from the transaction that originally created a change.

Confirmed on pr-970: the four new/updated tests pass locally.

On the toggle: I don't think it's needed for the common case — the subquery only runs once per row already bounded by pagesize (default 20) against an indexed connection_id lookup, so the cost tracks page size, not history size. The one case where it's not free is the endpoint's existing unpaginated mode (page omitted → all matching changes returned in one response, per `ObjectHistoryQueryArgs`'s own doc string) — an object with a long edit history pulled that way pays the subquery cost once per row in the full result, not just per page. That's a pre-existing API contract this PR doesn't introduce, so I wouldn't block on it, but worth keeping in mind if it ever shows up in practice (e.g. a bulk audit/export caller) — a cheap follow-up would be to skip transaction-id resolution when page is falsy, since that caller path presumably isn't rendering deep-link rows anyway.

**Minor, non-blocking nit**: `test_transaction_id_stable_when_ranges_absent` computes MIN(id) globally across transactions rather than per-connection — works because the fixture's changes all live in one connection, just a bit coupled to that assumption.

PR #975, opened 2026-09-09
Expose some config options in metadata

DavidMStraub:
# Expose client-relevant server capabilities under `/api/metadata`

Adds fields to the `server` block of `GET /api/metadata/`, visible to all
authenticated roles, so frontends can adapt up front instead of discovering
capabilities and limits from error responses.

| Key | Value |
|---|---|
| `face_detection` | bool — OpenCV + NumPy available |
| `thumbnails` | `{pdf, video}` — package *and* binary available |
| `email` | bool — `DEFAULT_FROM_EMAIL` set |
| `max_thumbnail_file_bytes` | int |
| `max_media_archive_upload_bytes` | int, omitted if uncapped |
| `rate_limit_media_archive` | `[{amount, window_seconds}]` |

The rate limit is parsed server-side with `limits.parse_many`, since the
configured string is Flask-Limiter's mini-language (`1 per day`,
`100/hour;1000/day`) that clients cannot be expected to parse. `MetadataSchema.server`
becomes a nested `ServerSchema`, so the block is now documented in the OpenAPI
spec. `limits` becomes a direct dependency.

All additions are additive; clients must tolerate missing keys on older servers.

### DIFF
--- a/gramps_webapi/api/resources/metadata.py
+++ b/gramps_webapi/api/resources/metadata.py
@@ -21,9 +21,12 @@
 """Metadata API resource."""
 
 import functools
+import shutil
 from importlib import metadata
+from importlib.util import find_spec
 
 import gramps_ql as gql
+import limits
 import object_ql as oql
 import sifts
 from flask import Response, current_app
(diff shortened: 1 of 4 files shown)


DavidMStraub:
@dsblank this should also address the issue you raised that frontends don't know in advance which file types can be thumbnailed.

dsblank:
Yes, this is great!

Issue #885, opened 2026-06-19
Make multi-tree the default


DavidMStraub:
**Is your feature request related to a problem? Please describe.**
The distinction between single-tree and multi-tree deployments is a headache for both backend and frontend, without any real benefit.

**Describe the solution you'd like**
Make multi-tree the default.

**Additional context**
We must ensure existing deployments continue to work without any - or with minimal - changes to configuration needed.

PR #981, opened 2026-09-13
Deprecate non-shared PostgreSQL

DavidMStraub:
I realized when we make multi-tree the default #885, we should remove support for the non-shared PostgreSQL addon, because it does not allow creating multiple trees.

Of course SharedPostgreSQL will still be supported.

The downside of this is that there might be users with deployments running PostgreSQL - since we don't collect telemetry on that, we have no way of knowing how many.

An alternative would be to continue supporting it, only not the `NEW_DB_BACKEND`, so databases created manually - or already existing - could still be used.

Let's use this PR for discussing the options before making a final decision.


DavidMStraub:
@dsblank FYI - this is for 4.0, which doesn't have a planned date yet.

dsblank:
@DavidMStraub I have no concerns except for the ones that you mentioned, except I have no desire to maintain it. It is fine with me to remove it. I only created it as an example.

PR #982, opened 2026-09-13
Add query endpoints to profiling script

DavidMStraub:
Background about the script: it's meant as a standardized benchmark to compare different deployments (SQLite vs Postgres, local vs cloud, etc.).


dsblank (approved):
This is a useful set of queries to keep an eye on. Note that this is the type of test that revealed the VACUUM issue in https://github.com/gramps-project/addons-source/pull/1026 (which is still a blocker for using more than one tree).

BTW, I believe that with the latest gramps-object-query-language, you can now use dotted paths instead of having to construct the _json_path().

DavidMStraub:
Changed to that.

Btw, piggybacked to this PR I'm also pinning the minor version of the library, we can bump it anytime as needed.

PR #986, opened 2026-09-14
Fix private proxy being applied twice for users without access to private records

DavidMStraub:
### Problem

For users without the `ViewPrivate` permission (Guests), `get_db_outside_request` already wraps the database in `ModifiedPrivateProxyDb`, and the result is cached in `g.db`. `get_db_handle` then wrapped `g.db` in `ModifiedPrivateProxyDb` a second time on every call.

Every request by such a user therefore went through two private proxies: every object was checked and sanitized twice, and methods like `find_backlink_handles` of the outer proxy loaded referring objects through the inner proxy, which sanitized them (including extra lookups) just to check their privacy flag.

### Change

`get_db_handle` returns the cached `g.db` directly for users without access to private records, since it is already a private proxy database. The result is unchanged, only the redundant second proxy layer is gone.

### Tests

- New `tests/test_endpoints/test_db_handle.py`: checks that an endpoint gets a single `ModifiedPrivateProxyDb` (whose underlying database is not a proxy) for a Guest, and the unproxied database for an Owner. The Guest test fails without the fix.
- Existing private proxy tests and endpoint test modules using the Guest role pass.

### Benchmarks

Profiler default endpoints on a synthetic tree with 100,000 people, 25,000 families, 159,908 events and 500 places (SQLite, single CPU core, request cache disabled), median of 3 runs, as Guest:

| Endpoint | Before | After |
|---|---|---|
| People List | 59.8 s | 18.8 s |
| Events by Date | 12.6 s | 7.0 s |
| Places List (`backlinks=1`) | 12.7 s | 4.1 s |
| Metadata | 1.56 s | 1.47 s |

Owner requests are unaffected (no proxy).

Remaining Guest-specific cost comes from the (single) private proxy itself, e.g. object counts in `/api/metadata/` and backlink privacy checks, which is not addressed here.


DavidMStraub:
@dsblank look at the bug I hid for years :face_with_peeking_eye:

dsblank:
Oh, my. That <s>might</s> does make a difference!

PR #990, opened 2026-09-15
Don't block foreground transactions on search index updates

DavidMStraub:
POST /transactions/ without background=1 applied the transaction and then updated the full-text and semantic search indices inside the request, holding a web worker for minutes on large payloads or with semantic search enabled. Split process_transactions into apply_transactions (database write) and the index update, and dispatch the index update via run_task in the foreground path, as the object endpoints already do. The background task still does both.


DavidMStraub:
This is in line what is done anyway on most of the endpoints.

It didn't surface so far because neither Gramps Web nor Gramps Web Sync use non-background transactions.

But recently a rogue agent swamped Grampshub with requests like that.

Note: this should also make GrampsWebApiDb edits faster as the sync requests don't have to wait for index updates. On deployments with semantic search (like Grampshub), it would probably be barely usable without this fix (I haven't tried yet). @dsblank FYI

DavidMStraub:
Ready

PR #994, opened 2026-09-16
Correctly handle thumbnailing of grayscale images

### LINKED ISSUE #983: [Bug] Greyscale JPGs result in broken AVIF thumbnails
### Bug Description

`save_image_buffer()` converts grayscale pixels to RGB using Pillow's `convert("RGB")`, but retains the source grayscale ICC profile. Saving as AVIF embeds that incompatible profile in the RGB output. Chrome refuses to decode the resulting thumbnail.

This produces broken portrait placeholders for an otherwise valid uploaded JPEG.

### Code

`gramps_webapi/api/image.py`, `save_image_buffer()`:

```python
elif image.mode not in ("RGB", "RGBA"):
    image = image.convert("RGB")
image.save(buffer, format=fmt)
```

### To Reproduce
Save the attached `reproduce_avif_profile.py`, then run:

```sh
uv run reproduce_avif_profile.py
open avif-profile-repro/index.html
```

[reproduce_avif_profile.py](https://github.com/user-attachments/files/32193690/reproduce_avif_profile.py)

### Recommended Fix

Use a colour-managed conversion when changing a profiled source into another colour space. `Image.convert("RGB")` changes pixel mode but does not transform or replace the ICC profile.

For this grayscale case, before the existing generic mode-conversion logic:

```python
from PIL import ImageCms

if image.mode == "L" and image.info.get("icc_profile"):
    source_profile = ImageCms.ImageCmsProfile(
        io.BytesIO(image.info["icc_profile"])
    )
    image = ImageCms.profileToProfile(
        image,
        source_profile,
        ImageCms.createProfile("sRGB"),
        outputMode="RGB",
    )
```

`profileToProfile` attaches the destination sRGB profile. Preserve that profile when encoding. This snippet addresses the reproduced grayscale case; review profiled CMYK and alpha-bearing inputs separately when implementing a general policy, preserving alpha where applicable.

Simply dropping the stale ICC after conversion is a smaller compatibility workaround, confirmed by the control, but can change colour appearance for non-default source profiles. Prefer proper source-to-sRGB conversion rather than stripping profiles indiscriminately from every image.

### Gramps Web Version Information

Gramps Web API v3.22.0

### Verification

- [x] I have searched for existing issues before creating this one

DavidMStraub:
Fixes #983


DavidMStraub:
Ready

PR #998, opened 2026-09-21
Fix OIDC login 500 for disabled/unconfirmed accounts (negative roles)

### LINKED ISSUE #997: OIDC login of a disabled/unconfirmed account 500s: get_permissions() KeyErrors on PERMISSIONS[role] for negative roles
### Summary

An OIDC login for an account whose role is `ROLE_DISABLED` (-1) or `ROLE_UNCONFIRMED` (-2) returns **500** instead of the intended "Account Under Review" page. `get_permissions()` does an unguarded `PERMISSIONS[user.role]` lookup, and `PERMISSIONS` has no entry for the two negative roles, so it raises `KeyError`.

Because new OIDC users are created with `ROLE_DISABLED` by default (pending admin approval), **every first-time OIDC sign-up hits this** when no group->role claim mapping is configured.

### Where

`gramps_webapi/auth/const.py` defines the roles `ROLE_DISABLED = -1` and `ROLE_UNCONFIRMED = -2`, but the `PERMISSIONS` dict is only populated for `ROLE_GUEST` (0) ... `ROLE_ADMIN` (5). There is no `PERMISSIONS[ROLE_DISABLED]` / `PERMISSIONS[ROLE_UNCONFIRMED]`.

`gramps_webapi/auth/__init__.py`, `get_permissions()`:

```python
user = query.filter_by(name=username).one()
permissions = PERMISSIONS[user.role].copy()   # KeyError when user.role < 0
```

### Why it's OIDC-only

Local login is safe -- `authorized()` returns `False` for `user.role < 0` before any permission lookup. But the OIDC callback (`api/resources/oidc.py`) calls `get_tree_id_and_permissions()` -> `get_permissions()` **before** its `user_details["role"] < 0` "Account Under Review" branch, so the `KeyError` fires first and that branch is never reached.

### Reproduce

1. Enable OIDC (e.g. Google) with no `OIDC_GROUP_*` role mapping.
2. Sign in with a new account -> `create_or_update_oidc_user()` creates it as `ROLE_DISABLED`.
3. The callback 500s (`KeyError`) instead of showing "Account Under Review".

Confirmed on `master`; also reproduced on 26.9.x / `gramps_webapi` 3.22.1.

### Suggested fix

Give the two non-positive roles an explicit empty permission set in `const.py`:

```python
PERMISSIONS[ROLE_UNCONFIRMED] = set()
PERMISSIONS[ROLE_DISABLED] = set()
```

Semantically correct (disabled/unconfirmed accounts have no permissions) and lets the OIDC callback reach the existing `role < 0` branch. Alternatively, guard the lookup: `PERMISSIONS.get(user.role, set()).copy()`. Happy to open a PR for whichever you prefer.

jon99nelson-svg:
Fixes #997.

## Problem

An OIDC login that resolves to an account with a negative role — `ROLE_DISABLED` (-1) or `ROLE_UNCONFIRMED` (-2) — returns **500** instead of the intended "Account Under Review" page.

Because new OIDC users are created as `ROLE_DISABLED` by default (pending admin approval) when no `OIDC_GROUP_*` role mapping is configured, this affects **every first-time OIDC sign-up** on such deployments, not just accounts an admin has explicitly disabled.

## Root cause

`PERMISSIONS` (in `gramps_webapi/auth/const.py`) is only populated for `ROLE_GUEST` (0) … `ROLE_ADMIN` (5); there is no entry for the two negative roles. `get_permissions()` does an unguarded `PERMISSIONS[user.role].copy()`, so it raises `KeyError` for a negative role.

The OIDC callback (`api/resources/oidc.py`) calls `get_tree_id_and_permissions()` → `get_permissions()` **before** its `user_details["role"] < 0` "Account Under Review" branch, so the `KeyError` fires first and that branch is never reached. Local login is unaffected: `authorized()` rejects `role < 0` before any permission lookup.

## Fix

Give the two non-positive roles an explicit empty permission set in `const.py`:

```python
PERMISSIONS[ROLE_DISABLED] = set()
PERMISSIONS[ROLE_UNCONFIRMED] = set()
```

Semantically correct (disabled/unconfirmed accounts have no permissions) and it lets the OIDC callback reach the existing `role < 0` branch and render the review page.

## Test

Adds `TestOIDCSingleTree.test_callback_disabled_account_renders_review_page`, which drives a real OIDC callback for a `ROLE_DISABLED` account and asserts it renders the "Account Under Review" page (HTTP 200) rather than 500. It intentionally does **not** mock `get_tree_id_and_permissions`/`get_permissions`, so it exercises the exact lookup that previously raised. The test fails before this change and passes after.


DavidMStraub:
Thanks!

PR #1001, opened 2026-09-22
feat(webpush): add user push subscription infrastructure

### LINKED ISSUE #974: Add Web Push infrastructure
Add generic Web Push support to gramps-web-api.

**Scope**

* Store push subscriptions per user.
* Add endpoints to create/update and delete subscriptions.
* Add a reusable server-side helper to send push notifications.
* Remove invalid/expired subscriptions when delivery fails.
* Add tests.

elig-45:
## Summary

Add generic Web Push infrastructure to Gramps Web API.

## Changes

- store multiple Web Push subscriptions per user
- add authenticated endpoints to list, create/update, and delete subscriptions
- validate subscription endpoints and browser-generated keys
- add configurable VAPID settings and subscription limits
- add a reusable server-side delivery helper
- automatically remove expired subscriptions and subscriptions rejected with HTTP 404/410
- add the user-database migration and endpoint/delivery tests

Subscription secrets and endpoints are never returned by the listing endpoint.

## Validation

- 13 targeted endpoint and delivery tests pass
- mypy passes for the new backend modules
- Black and Python compilation checks pass
- Alembic upgrade and downgrade smoke tests pass
- the migration chain has a single head

## Frontend

Companion client implementation: [gramps-web#1456](https://github.com/gramps-project/gramps-web/pull/1456).

Closes #974


DavidMStraub:
Thanks! Below are some suggestions by Claude to make it more minimal. What do you think? Is some of the stuff dispensable?

---


**Table / migration**

- Drop `created_at` and `updated_at`: nothing reads them. This also removes the deprecated `datetime.utcnow()`.
- Drop `expiration_time`: browsers send `null` in practice. Removing it also removes the expiry check on POST, the expiry sweep in the sender, and the corresponding test.
- `endpoint_hash`: the SHA-256 collision check is dead code. Either put the unique index directly on `endpoint` with a reasonable length cap, or keep the hash and drop the collision branch.
- Remaining columns: `id`, `user_id`, `endpoint`, `p256dh`, `auth`.

**Config**

- Keep only `WEB_PUSH_VAPID_PRIVATE_KEY`, `WEB_PUSH_VAPID_PUBLIC_KEY`, and `WEB_PUSH_VAPID_SUBJECT`.
- Drop `WEB_PUSH_MAX_PAYLOAD_BYTES`, `WEB_PUSH_TTL`, and `WEB_PUSH_TIMEOUT`: library defaults or constants suffice, and push services enforce the payload limit themselves.
- `WEB_PUSH_MAX_SUBSCRIPTIONS_PER_USER` can be a module-level constant.
- Optional: derive the public key from the private key and default the subject to `BASE_URL`, leaving a single setting.

**Endpoint**

- GET only needs to return the VAPID public key (or `null` if push is not configured). The subscription count is not needed.
- POST should return 201 and DELETE 204, without a status body.
- Drop the public key format validation in `_config()`: it is a configuration concern and does not need checking on every request.
- Keep the endpoint URL validation and the key validation.
- Move the schemas to `schemas.py` and add `metadata={"description": ...}` to every field for the OpenAPI docs.

**Delivery helper**

- Drop `send_web_push_to_user` and the `WebPushDeliveryResult` dataclass: neither is used. A single `send_web_push(user_id, payload)` that deletes subscriptions rejected with 404/410 is sufficient.
- Drop `WebPushNotConfiguredError` together with the payload size check.

elig-45:
Thanks, I agree that most of this can be simplified.

I’ll keep `endpoint_hash` to retain a short, portable unique index, but remove the theoretical collision branch. I’ll apply the remaining reductions: a smaller table and configuration surface, a public-key-only GET response, empty 201/204 mutation responses, documented shared schemas, and a single delivery helper that removes 404/410 subscriptions.

I’ll keep the VAPID public key and subject explicit for now rather than deriving them, and I’m updating the companion frontend to the leaner API contract as well.

PR #1002, opened 2026-09-22
Fix preload_event_backlinks() qmark placeholders on single-user PostgreSQL addon

### LINKED ISSUE #999: preload_event_backlinks() sends qmark placeholders straight to psycopg2 on the single-user PostgreSQL addon backend
Introduced by #958 (merged 2026-09-08). `preload_event_backlinks()` in `api/resources/util.py` builds `SELECT ref_handle, obj_class, obj_handle FROM reference WHERE ref_class = ? AND obj_class IN ('Person', 'Family')` and calls `dbapi.cursor().execute(sql, params)` directly. On the single-user Gramps PostgreSQL addon (not SharedPostgreSQL), `dbapi.cursor().execute()` (`gramps/plugins/PostgreSQL/postgresql.py:315`) passes the query straight to `psycopg2`, which expects `%s`-style placeholders, not `?`. `reindex_full()` (and therefore the manual full-reindex path, `POST /api/search/index/?full=1`) crashes immediately:

```
psycopg2.errors.SyntaxError: syntax error at or near "AND"
LINE 1: ...ss, obj_handle FROM reference WHERE ref_class = ? AND obj_cl...
```

`preload_event_backlinks()`'s own docstring says it targets "sqlite, the single-user PostgreSQL addon; shared with object_query.py's `_resolve_treeid()`" — so this backend is meant to be supported, but the qmark params aren't translated for it. Reproduced on gramps-web-api 3.22.1, gramps 6.0, single-user `PostgreSQL` addon backend (not `SharedPostgreSQL`).

Setup: `gramps-web-api` in Docker, `GRAMPSWEB_NEW_DB_BACKEND=postgresql` (single-user addon, not SharedPostgreSQL), tree already populated (not a fresh import).

dsblank:
## Summary
- `preload_event_backlinks()` in `api/resources/util.py` (added in #958) builds raw SQL with sqlite-style `?` placeholders and executes it via `dbapi.cursor()`.
- On the single-user `PostgreSQL` addon, `dbapi.cursor().execute()` passes SQL straight to psycopg2 untranslated (unlike the addon's connection-level `dbapi.execute()`, which does translate `?` → `%s`), so the query fails with a `psycopg2.errors.SyntaxError`.
- This crashes the full search reindex path (`POST /api/search/index/?full=1`) on that backend, reported in #999.
- Adds `needs_qmark_translation()` to `db_backend.py` (detected by class name, consistent with the existing `is_sqlite()`/`SINGLE_TREE_DBAPI_CLASS_NAMES` checks in that module) and applies it in `preload_event_backlinks()` before executing, so the fix doesn't depend on the addon itself being patched or re-released — see companion addon fix at gramps-project/addons-source (targeting `maintenance/gramps60`).

## Test plan
- [x] `python -m pytest tests/test_util.py -k preload_event_backlinks -q` — 6 passed (2 new: single-user `PostgreSQL` and `SharedPostgreSQL` both now get `%s` placeholders)
- [x] `python -m pytest tests/test_util.py -q` — 53 passed

Fixes #999


DavidMStraub:
Thanks!

Would this be easier if we fixed it in the addon?

dsblank:
> Would this be easier if we fixed it in the addon?

I asked Claude that too, but it still leaves the older single postgresql addon in a bad state. So I made the fix in both places.

